PR #250, opened 2016-10-05
Next/sergio/ashell rebase master

sergioamr:
[ashell] Added bluetooth and improved serial handling

Added extra machine readable tokens so the editor is able to read and use the outputs from the command line.

Changes to support uploading and check of code from the browser

makefile - Added support for grove LCD

jerry-code Split parse and run so we can check if the syntax is correct.

Fixed bug on carriage return that was introducing a \0 value after printing text

Working on the uart to save some memory during the transactions

Uart is still a bit unstable in slow devices like a nuc

It depends on https://github.com/01org/zephyr.js/pull/241 to work.

Signed-off-by: Sergio Martinez <sergio.martinez.rodriguez@intel.com>


jimmy-huang:
@sergioamr please rebase and resubmit to devel branch, and then close this one.

PR #253, opened 2016-10-05
[ARC] Fix a segfault in ARC when IPM malloc fails

jimmy-huang:
When the ipm message fails to allocate memory on the x86
for the IPM message, the message still gets sent as NULL,
the ARC doesn't check NULL messages and then segfaults.


jimmy-huang:
Closing this, and resubmitted #264 to devel branch

PR #254, opened 2016-10-06
[IPM] No need to malloc memory on ipm

jimmy-huang:
We don't have to malloc heap memory for sending/receiving IPM
messages in sync API, since the ARC is guaranteed to receive the message
after the send() function returns, we can immediately free up
the IPM message, and so to make it more efficient, we just use
the stack memory to store the struct.


jimmy-huang:
Closing this, resubmitted #265 to devel branch

PR #263, opened 2016-10-07
[BLE] Change BLE to use zjs callback API

jimmy-huang:

### DIFF
--- a/src/zjs_ble.c
+++ b/src/zjs_ble.c
@@ -31,7 +31,8 @@
 struct nano_sem zjs_ble_nano_sem;
 
 typedef struct ble_handle {
-    struct zjs_callback zjs_cb;
+    int32_t id;
+    jerry_value_t js_callback;
     const void *buffer;
     uint16_t buffer_size;
     uint16_t offset;
@@ -39,8 +40,8 @@ typedef struct ble_handle {
 } ble_handle_t;
 
 typedef struct ble_notify_handle {
-    struct zjs_callback zjs_cb;
-    uint16_t max_value_size;
+    int32_t id;
+    jerry_value_t js_callback;
 } ble_notify_handle_t;
 
 typedef struct ble_event_handle {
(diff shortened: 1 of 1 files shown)


jimmy-huang:
@jprestwo thanks, updated PR

jprestwo:
+1 Looks good

src/zjs_ble.c
```diff
@@ -285,52 +283,45 @@ static jerry_value_t zjs_ble_write_attr_call_function_return(const jerry_value_t
...
-           } else {
-               args[0] = jerry_create_null();
+               buf->bufsize == cb->buffer_size) {
+               memcpy(buf->buffer, cb->buffer, cb->buffer_size);
            }
-        } else {
-            args[0] = jerry_create_null();
+
+           jerry_release_value(args[0]);
+           args[0]= buf_obj;
```
grgustaf: space before equals

src/zjs_ble.c
```diff
@@ -285,52 +283,45 @@ static jerry_value_t zjs_ble_write_attr_call_function_return(const jerry_value_t
...
-    rval = jerry_call_function(mycb->zjs_cb.js_callback, chrc->chrc_obj, args, 4);
+    rval = jerry_call_function(cb->js_callback, chrc->chrc_obj, args, 4);
     if (jerry_value_has_error_flag(rval)) {
-        PRINT("zjs_ble_write_attr_call_function: failed to call onWriteRequest function\n");
+        PRINT("zjs_ble_write_c_callback: failed to call onWriteRequest function\n");
     }
 
     jerry_release_value(args[0]);
-    jerry_release_value(args[1]);
-    jerry_release_value(args[2]);
```
grgustaf: Why did you remove these two?

src/zjs_ble.c
```diff
@@ -405,28 +395,55 @@ static jerry_value_t zjs_ble_update_value_call_function(const jerry_value_t func
...
-    args[1] = jerry_create_external_function(zjs_ble_update_value_call_function);
-    rval = jerry_call_function(mycb->zjs_cb.js_callback, chrc->chrc_obj, args, 2);
+    args[1] = jerry_create_external_function(zjs_ble_update_value_callback_function);
+    rval = jerry_call_function(cb->js_callback, chrc->chrc_obj, args, 2);
     if (jerry_value_has_error_flag(rval)) {
-        PRINT("zjs_ble_subscribe_call_function: failed to call onSubscribe function\n");
+        PRINT("zjs_ble_subscribe_c_callback: failed to call onSubscribe function\n");
     }
 
-    jerry_release_value(args[0]);
```
grgustaf: Why?

src/zjs_ble.c
```diff
@@ -405,28 +395,55 @@ static jerry_value_t zjs_ble_update_value_call_function(const jerry_value_t func
...
+
+    jerry_value_t rval;
+
+    rval = jerry_call_function(cb->js_callback, chrc->chrc_obj, NULL, 0);
+    if (jerry_value_has_error_flag(rval)) {
+        PRINT("zjs_ble_notify_c_callback: failed to call onNotify function\n");
+    }
+
+    jerry_release_value(rval);
+}
```
grgustaf: These two functions are identical, I think, so make a more generic name and use it twice?

src/zjs_ble.c
```diff
@@ -855,6 +887,11 @@ static bool zjs_ble_parse_service(ble_service_t *service)
         }
 
         memset(chrc, 0, sizeof(ble_characteristic_t));
+        chrc->read_cb.id = chrc->write_cb.id
+                         = chrc->subscribe_cb.id
+                         = chrc->unsubscribe_cb.id
+                         = chrc->notify_cb.id
+                         = -1;
```
grgustaf: Hmm, seems wacky. But if that's what the kids are doing these days. :)

grgustaf:
Also, two unrelated patches from @haoxli are currently in the PR?

src/zjs_ble.c
jimmy-huang: argp[1] and arg[2] are integer and boolean, so they do not need to be released, but I'll put them back in for the sake in the case the jerryscript api changes in the future

src/zjs_ble.c
jimmy-huang: if there's a better/cleaner way to do this, let me know, i can adapt.

src/zjs_ble.c
jimmy-huang: they are not exactly the same, since the first *cb points to &chrc->unsubscribe_cb, but the second one points to &chrc->notify_cb.

PR #266, opened 2016-10-07
[Sample] Adding a sample program for BMP280

brianjjones:
This is a sample javascript program that uses both the BMP280
and the Grove LCD.  Both sensors work off the same I2C bus.  The
BMP280 reads the temperature while the Grove LCD prints it.  The
Grove LCD also changes to red as it gets warmer and blue as it
gets cooler.  There is also a (+) or (-) indicating if the temp
has gone up or down since the last read.


jimmy-huang:
Please submit this to devel branch instead of master

brianjjones:
Whoops forgot we had switched, I'll move it over.

PR #271, opened 2016-10-10
[ashell] Improved serial handling & machine to machine transactions

sergioamr:
Added extra machine readable tokens so the editor is able to read and use the outputs from the command line.

Changes to support uploading and check of code from the browser

Added support for parameters so commands can add a single character parameters like `-x` or `-xyz`

makefile - Added support for grove LCD
jerry-code Split parse and run so we can check if the syntax is correct.

Fixed bug on carriage return that was introducing a \0 value after printing with carriage return.

Working on the uart to save some memory during the transactions on machine to machine transactions.

Uart is still a bit unstable in slow devices like old atom nucs

Signed-off-by: Sergio Martinez <sergio.martinez.rodriguez@intel.com>


src/ashell/acm-uart.c
```diff
@@ -65,18 +69,10 @@
...
-static const char banner[] = "ZephyrJerry\r\n" \
-    __DATE__ " " __TIME__ "\r\n" \
-    "    ____\r\n"\
-    "  ,'   Y`.\r\n"\
-    " /        \\\r\n"\
-    " \\ ()  () /\r\n"\
-    "  `. /\\ ,'\r\n"\
-    "8===|\"\"|===8\r\n"\
-    "    `LL'\r\n";
+static const char banner[] = "ZephyrJerry " __DATE__ " " __TIME__ "\n";
```
poussa: Change to zephyr.js

poussa: ...and add DEV MODE

sergioamr: Makes sense!

poussa:
I was using this branch successfully during my Open IoT Summit presentation demo. So it works.

Ready to merge so we can base future work on top.

PR #272, opened 2016-10-10
[Sample] Adding a sample program for BMP280

brianjjones:
This is a sample javascript program that uses both the BMP280 and the Grove LCD.  Both sensors work off the same I2C bus.  The BMP280 reads the temperature while the Grove LCD prints it.  The Grove LCD also changes to red as it gets warmer and blue as it gets cooler.  There is also a (+) or (-) indicating if the temp has gone up or down since the last read.

### DIFF
--- a/samples/I2CBMP280.js
+++ b/samples/I2CBMP280.js
@@ -72,17 +72,6 @@ var i2cDevice = i2c.open({ bus: 0, speed: 100 });
 var dig_T1;
 var dig_T2;
 var dig_T3;
-
-// Global Buffer objects
-var setupData = new Buffer(2);
-var clearData = new Buffer([0, glcd.GLCD_CMD_SCREEN_CLEAR]);
-var msgData = new Buffer("Temperature is ");
-var tempData = new Buffer(" ##.#C");
-var changeData = new Buffer(" +");
-var redData = new Buffer([glcd.REGISTER_R, 0]);
-var greenData = new Buffer([glcd.REGISTER_G, 0]);
-var blueData = new Buffer([glcd.REGISTER_B, 0]);
-
 var baseTemp = prevTemp = 25;
 var firstRead = true;
 var red = green = blue = 200;
@@ -107,9 +96,9 @@ function resetCursor(col, row) {
         col |= 0xC0;
     }
 
-    setupData.writeUInt8(glcd.GLCD_CMD_SET_DDRAM_ADDR, 0);
-    setupData.writeUInt8(col, 1);
-    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, setupData);
+    // Write new cursor position via I2C
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR,
+                    new Buffer([glcd.GLCD_CMD_SET_DDRAM_ADDR, col]));
 }
 
 function setColorFromTemperature(newTemp) {
@@ -130,24 +119,27 @@ function setColorFromTemperature(newTemp) {
     prevTemp = newTemp;
 
     // Valid range for color is 0 - 255
-    redData.writeUInt8(red, 1);
-    greenData.writeUInt8(green, 1);
-    blueData.writeUInt8(blue, 1);
-    // Send messages as close together as possible,
-    // so that the color change is smoother
-    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, blueData);
-    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, redData);
-    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, greenData);
+    // Write new color values
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([glcd.REGISTER_B, blue]));
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([glcd.REGISTER_R, red]));
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([glcd.REGISTER_G, green]));
 }
 
-function writeWord(word, buffer) {
-    // Add the word to the buffer
-    buffer.write(word, 1);
+function writeText(word) {
+    var wordBuffer = new Buffer(word.length + 1);
+
     // The first byte in the buffer is for the register address,
     // which is where the word is going to be written to.
-    buffer.writeUInt8(glcd.GLCD_CMD_SET_CGRAM_ADDR, 0);
+    wordBuffer.writeUInt8(glcd.GLCD_CMD_SET_CGRAM_ADDR, 0);
+
+    // Append the text we want to print after that
+    wordBuffer.write(word, 1);
+
     // Write the buffer to I2C
-    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, buffer);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, wordBuffer);
 }
 
 function init() {
@@ -160,38 +152,39 @@ function init() {
 
     readCoefficients();
 
-    //var powerData = new Buffer([bmp280.BMP280_REGISTER_CONTROL, 0]);
-    setupData.writeUInt8(bmp280.BMP280_REGISTER_CONTROL, 0);
-    setupData.writeUInt8(bmp280.BMP280_PRESS_OVER | bmp280.BMP280_TEMP_OVER |
-                         bmp280.BMP280_NORMAL_MODE, 1)
+    /* Setup BMP280 sensor */
+    i2cDevice.write(bmp280.BMP280_ADDRESS,
+                    new Buffer([bmp280.BMP280_REGISTER_CONTROL,
+                                bmp280.BMP280_PRESS_OVER |
+                                bmp280.BMP280_TEMP_OVER |
+                                bmp280.BMP280_NORMAL_MODE]));
 
-    i2cDevice.write(bmp280.BMP280_ADDRESS, setupData);
-
-    setupData.writeUInt8(bmp280.BMP280_REGISTER_CONFIG, 0);
-    setupData.writeUInt8(bmp280.BMP280_STANDBY | bmp280.BMP280_FILTER |
-                         bmp280.BMP280_SPI_3W_DISABLE, 1);
-
-    i2cDevice.write(bmp280.BMP280_ADDRESS, setupData);
+    i2cDevice.write(bmp280.BMP280_ADDRESS,
+                    new Buffer([bmp280.BMP280_REGISTER_CONFIG,
+                                bmp280.BMP280_STANDBY |
+                                bmp280.BMP280_FILTER |
+                                bmp280.BMP280_SPI_3W_DISABLE]));
 
     /* Setup Grove LCD */
     // Clear the LCD
-    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, clearData);
-    // Set our preferences for the Grove LCD
-    var setup = glcd.GLCD_CMD_FUNCTION_SET | (glcd.GLCD_FS_ROWS_2 |
-                                              glcd.GLCD_FS_DOT_SIZE_LITTLE |
-                                              glcd.GLCD_FS_8BIT_MODE);
-    setupData.writeUInt8(0, 0);
-    setupData.writeUInt8(setup, 1);
-    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, setupData);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR,
+                    new Buffer([0, glcd.GLCD_CMD_SCREEN_CLEAR]));
 
-    setup = glcd.GLCD_CMD_DISPLAY_SWITCH | (glcd.GLCD_DS_DISPLAY_ON |
-                                            glcd.GLCD_DS_CURSOR_ON |
-                                            glcd.GLCD_DS_BLINK_ON);
-    setupData.writeUInt8(setup, 1);
-    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, setupData);
+    // Set our preferences for the Grove LCD
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR,
+                    new Buffer([0, glcd.GLCD_CMD_FUNCTION_SET |
+                                   glcd.GLCD_FS_ROWS_2 |
+                                   glcd.GLCD_FS_DOT_SIZE_LITTLE |
+                                   glcd.GLCD_FS_8BIT_MODE]));
+
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR,
+                    new Buffer([0, glcd.GLCD_CMD_DISPLAY_SWITCH |
+                                   glcd.GLCD_DS_DISPLAY_ON |
+                                   glcd.GLCD_DS_CURSOR_OFF |
+                                   glcd.GLCD_DS_BLINK_OFF]));
 
     resetCursor(0, 0);
-    writeWord("Temperature is ", msgData);
+    writeText("Temperature is ");
 }
 
 function readCoefficients() {
@@ -209,7 +202,6 @@ function readCoefficients() {
 }
 
 function readTemperature() {
-
     // Read the raw temperature value.
     var tempBuffer = i2cDevice.burstRead(bmp280.BMP280_ADDRESS, 4,
                                          bmp280.BMP280_REGISTER_TEMPDATA);
@@ -235,14 +227,14 @@ function readTemperature() {
     // Draw icon for direction temp has changed
     resetCursor(0, 1);
     if (prevTemp < T) {
-        writeWord("+", changeData);
+        writeText("+");
     } else {
-        writeWord("-", changeData);
+        writeText("-");
     }
 
     // Write out the temp to the LCD
     resetCursor(9, 1);
-    writeWord(tempStr, tempData);
+    writeText(tempStr);
     setColorFromTemperature(T);
     prevTemp = T;
 }
--- a/samples/I2CBMP280.js
+++ b/samples/I2CBMP280.js
@@ -7,11 +7,11 @@
 //   - pull-up resistors for SDA and SCL, use two 10k resistors
 
 // Wiring:
-//   For LCD:
-//     - Wire SDA on the BMP280 to the pull-up resistor and connect that resistor to power (VCC)
-//     - Wire SCL on the BMP280 to the pull-up resistor and connect that resistor to power (VCC)
-//     - Wire SDA on the BMP280 to SDA on the Arduino 101
-//     - Wire SCL on the BMP280 to SCL on the Arduino 101
+//   For LCD and BMP280:
+//     - Wire SDA on the Arduino 101 to the pull-up resistor and connect that resistor to power (VCC)
+//     - Wire SCL on the Arduino 101 to the pull-up resistor and connect that resistor to power (VCC)
+//     - Wire SDA on the LCD and BMP280 to SDA on the Arduino 101
+//     - Wire SCL on the LCD and BMP280 to SCL on the Arduino 101
 //     - Wire power(5V) and ground accordingly
 
 var i2c = require("i2c");
@@ -25,7 +25,7 @@ var bmp280 = {
     BMP280_REGISTER_CHIPID: 0xD0,
     BMP280_REGISTER_CONTROL: 0xF4,
     BMP280_REGISTER_CONFIG: 0xF5,
-    BMP280_REGISTER_TEMPDATA: 0xFA, //Signed 20bit. Always positive
+    BMP280_REGISTER_TEMPDATA: 0xFA, // Signed 20bit. Always positive
     BMP280_STANDBY: 2 << 5,
     BMP280_FILTER: 0x00,
     BMP280_SPI_3W_DISABLE: 0x00,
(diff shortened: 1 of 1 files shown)


samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    // Send messages as close together as possible,
+    // so that the color change is smoother
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, blueData);
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, redData);
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, greenData);
+}
+
+function updateColor(newTemp) {
+    var diff = newTemp - baseTemp;
+    var change = diff > 0 ? diff * 40 : diff * -40;
```
jimmy-huang: i think these 1 liners would be easier to read with parenthesis, (diff > 0) ? (diff * 40) : (diff * -40);

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    change = change > 255 ? 255 : change | 0;
```
jimmy-huang: same here.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+
+    if (baseTemp < newTemp) {
+        blue = 200 - change < 0 ? 0 : 200 - change;
```
jimmy-huang: maybe blue = (200 - change < 0) ? 0 : (200 - change);

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+        green = blue;
+        red = 200 + change > 255 ? 255 : 200 + change;
```
jimmy-huang: ditto

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    } else {
+        red = 200 - change < 0 ? 0 : 200 - change;
```
jimmy-huang: ditto

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+        green = red;
+        blue = 200 + change > 255 ? 255 : 200 + change;
```
jimmy-huang: ditto

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    prevTemp = newTemp;
+    changeRGB(red | 0, green | 0, blue | 0);
```
jimmy-huang: i don't think you'll need to do the | 0 here to convert it to int, since change is already an integer.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    // Read the raw temperature value.
+    var tempBuffer = i2cDevice.burstRead(bmp280.BMP280_ADDRESS, 4,
+                                         bmp280.BMP280_REGISTER_TEMPDATA);
+    // Get the number from the buffer object
+    var tempDec = tempBuffer.readUInt32BE();
+    //Select the first 20 bits of 32 bit integer and maintains the sign
+    tempDec >>>= 12;
+
+    //Convert raw temperature to Celsius
+    var1 = ((((tempDec >> 3) - (dig_T1 << 1))) * (dig_T2)) >> 11;
```
jimmy-huang: looks like there's 1 extra parenthesis,
(((tempDec >> 3) - (dig_T1 << 1)) * (dig_T2)) >> 11

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    var2 = (((((tempDec >> 4) - (dig_T1)) * ((tempDec >> 4) - (dig_T1))) >> 12) *
+            (dig_T3)) >> 14;
+
+    t_fine = var1 + var2;
```
jimmy-huang: t_fine and T is not declared anywhere

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    T = ((t_fine * 5 + 128) >> 8) / 100 ;
+
+    print ("Temp = " + T);
+    var dec = (T * 10) | 0;
+    dec = (dec - (T | 0) * 10) | 0;
```
jimmy-huang: are you trying to make sure the results are integer?  i think there's no need to do so many | 0, if you just make T = T | 0, then the rest of the algorithm should be just integer.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    var tempStr = "@" + (T | 0)  + '.' + dec + "C";
+
+    //Draw icon for direction temp has changed
+    resetCursor(0, 1);
+    if (prevTemp < T) {
+        writeWord("@+", changeData);
```
jimmy-huang: btw, is this "@" the special character it needs for the grove lcd?  since it seems every string needs to start with the "@", so you might have writeWord() appends that character automatically, then you just call it with a regular string.

samples/I2CBMP280.js
brianjjones: You knocked my extra parenths last time ;)

Yeah no problem I'll add it

samples/I2CBMP280.js
brianjjones: Yeah it's the register address for writing to the display.  I'll move it to the write word function.

samples/I2CBMP280.js
brianjjones: This is to get just the first digit of the decimal since I don't have Math available.  But yeah I suppose I don't need the last | 0, I'm just being overly cautious as usual.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+}
+
+function clear() {
+    var clearData = new Buffer([0, glcd.GLCD_CMD_SCREEN_CLEAR]);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, clearData);
+}
+
+function resetCursor(col, row) {
+    // Reset the cursor position
+    row = row === undefined ? 0 : row;
```
kenchris: row = row || 0;

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+//   - A BMP280 Temp/pressure sensor
+//   - pull-up resistors for SDA and SCL, use two 10k resistors
+
+// Wiring:
+//   For LCD:
+//     - Wire SDA on the BMP280 to the pull-up resistor and connect that resistor to power (VCC)
+//     - Wire SCL on the BMP280 to the pull-up resistor and connect that resistor to power (VCC)
+//     - Wire SDA on the BMP280 to SDA on the Arduino 101
+//     - Wire SCL on the BMP280 to SCL on the Arduino 101
+//     - Wire power(5V) and ground accordingly
```
haoxli: For LCD? or BMP280?

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+                                            glcd.GLCD_DS_BLINK_ON);
+    setupData.writeUInt8(setup, 1);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, setupData);
+
+    resetCursor(0, 0);
+    writeWord("Temperature is ", msgData);
+}
+
+function readCoefficients() {
+    //These need to be read as a burst read in order to get accurate numbers
```
kenchris: Space after //

kenchris: Everywhere :)

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    dig_T3 = dig_TAll.readUInt16LE(4);
+
+    //Convert these to signed
+    dig_T1 = uint16toInt16(dig_T1);
+    dig_T2 = uint16toInt16(dig_T2);
+    dig_T3 = uint16toInt16(dig_T3);
+}
+
+function readTemperature() {
+    var var1, var2;
```
kenchris: That is very un-JS like

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+
+    //Convert raw temperature to Celsius
+    var1 = (((tempDec >> 3) - (dig_T1 << 1))  * (dig_T2)) >> 11;
+    var2 = (((((tempDec >> 4) - (dig_T1)) * ((tempDec >> 4) - (dig_T1))) >> 12) *
+            (dig_T3)) >> 14;
+
+    var t_fine = var1 + var2;
+    var T = ((t_fine * 5 + 128) >> 8) / 100 ;
+
+    print ("Temp = " + T);
```
kenchris: no space before (

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    var dec = (T * 10) | 0;
+    dec = dec - (T | 0) * 10;
+    var tempStr = (T | 0)  + '.' + dec + "C";
```
kenchris: ' vs " - very inconsistent :)

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+function readCoefficients() {
+    //These need to be read as a burst read in order to get accurate numbers
+    dig_TAll = i2cDevice.burstRead(bmp280.BMP280_ADDRESS, 24,
+                                   bmp280.BMP280_REGISTER_DIG_T1);
+    dig_T1 = dig_TAll.readUInt16LE();
+    dig_T2 = dig_TAll.readUInt16LE(2);
+    dig_T3 = dig_TAll.readUInt16LE(4);
+
+    //Convert these to signed
+    dig_T1 = uint16toInt16(dig_T1);
```
kenchris: You cannot read the value directly as signed?

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+
+    // Read the raw temperature value.
+    var tempBuffer = i2cDevice.burstRead(bmp280.BMP280_ADDRESS, 4,
+                                         bmp280.BMP280_REGISTER_TEMPDATA);
+    // Get the number from the buffer object
+    var tempDec = tempBuffer.readUInt32BE();
+    //Select the first 20 bits of 32 bit integer and maintains the sign
+    tempDec >>>= 12;
+
+    //Convert raw temperature to Celsius
```
kenchris: Link to math?

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    REGISTER_B : 0x02
+}
+
+/* Globals */
+var i2cDevice = i2c.open({ bus: 0, speed: 100 });
+var reader;
+var dig_T1;
+var dig_T2;
+var dig_T3;
+var setupData = new Buffer(2);
```
kenchris: if you only use this for the set up why not just do it there... ie avoid all these ugly globals

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+var red = green = blue = 200;
+
+function uint16toInt16(value) {
+    if ((value & 0x8000) > 0) {
+        value = value - 0x10000;
+    }
+    return value;
+}
+
+function clear() {
```
kenchris: clearDisplay()

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, clearData);
+}
+
+function resetCursor(col, row) {
+    // Reset the cursor position
+    row = row === undefined ? 0 : row;
+    col = col === undefined ? 0 : col;
+
+    if (row == 0) {
+        col |= 0x80;
```
kenchris: might need a comment

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    } else {
+        col |= 0xC0;
+    }
+
+    setupData.writeUInt8(glcd.GLCD_CMD_SET_DDRAM_ADDR, 0);
+    setupData.writeUInt8(col, 1);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, setupData);
```
kenchris: It doesnt really feel like a setupData if it is being reused later. Data makes it sound like it already carries data, but you are using it just like a reusable buffer. I think that buffer would be a better name, maybe 2bBuffer or similar

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+
+function changeRGB(red, green, blue) {
+    // Valid range for color is 0 - 255
+    var redData = new Buffer([glcd.REGISTER_R, red]);
+    var greenData = new Buffer([glcd.REGISTER_G, green]);
+    var blueData = new Buffer([glcd.REGISTER_B, blue]);
+
+    // Send messages as close together as possible,
+    // so that the color change is smoother
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, blueData);
```
kenchris: Maybe nicer and more concise as

```
var address = glcd.GROVE_RGB_BACKLIGHT_ADDR;
i2cDevice.write(address, new Buffer([glcd.REGISTER_R, red]));
i2cDevice.write(address, new Buffer([glcd.REGISTER_G, green]));
i2cDevice.write(address, new Buffer([glcd.REGISTER_B, blue]));
```

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, redData);
+    i2cDevice.write(glcd.GROVE_RGB_BACKLIGHT_ADDR, greenData);
+}
+
+function updateColor(newTemp) {
```
kenchris: I would do

updateColor -> setColorFromTemperature()
changeRBG -> setColor()

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+    buffer.write(word);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, buffer);
+}
+
+function init() {
+    reader = i2cDevice.read(bmp280.BMP280_ADDRESS, 1,
+                            bmp280.BMP280_REGISTER_CHIPID);
+
+    if (reader.toString('hex') !== "58")
+        print("BMP280 not found! ChipId " + reader.toString('hex') !== "58");
```
kenchris: Sorry I don't know what ChipId something would mean :-) Would others?

brianjjones: Its in the sensor spec sheet for the BMP280, but I certainly can add some more comments to help explain things.  Basically I'm checking that we have a BMP280 on the I2C bus.

samples/I2CBMP280.js
brianjjones: It should be both, I'll fix it to be clearer.

samples/I2CBMP280.js
brianjjones: Not yet, Buffer currently doesn't have that API implemented.

samples/I2CBMP280.js
brianjjones: Its used in resetCursor each time I write something, I didn't want to allocate the Buffer every time that occurred.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+function uint16toInt16(value) {
+    if ((value & 0x8000) > 0) {
+        value = value - 0x10000;
+    }
+    return value;
+}
+
+function clear() {
+    var clearData = new Buffer([0, glcd.GLCD_CMD_SCREEN_CLEAR]);
+    i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, clearData);
```
brianjjones: I realized its only being called in one place so I just replaced it with i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, clearData);

samples/I2CBMP280.js
brianjjones: It would be, but I realized I don't want to be creating a new Buffer each time so I've changed them to be global.  That way the memory can be reused and cut down on the malloc calls.

samples/I2CBMP280.js
brianjjones: Sounds good, also condensing these into one function.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,244 @@
...
+        blue = (200 + change > 255) ? 255 : (200 + change);
+    }
+
+    prevTemp = newTemp;
+    changeRGB(red, green, blue);
+}
+
+function writeWord(word, buffer) {
+    word = "@" + word;  //This is needed to write to the correct register address
+    buffer.write(word);
```
brianjjones: Changing this to use GLCD_CMD_SET_CGRAM_ADDR (which is == to 64, the dec value of @) instead so that its clear I'm setting the register address.  Using @ allowed me to just write the string in one line of code, but its not clear to the user why on earth I'm putting the @ in there ;)

samples/I2CBMP280.js
```diff
@@ -138,38 +128,54 @@ function updateColor(newTemp) {
...
+    // which is where the word is going to be written to.
+    buffer.writeUInt8(glcd.GLCD_CMD_SET_CGRAM_ADDR, 0);
+    // Write the buffer to I2C
     i2cDevice.write(glcd.GROVE_LCD_DISPLAY_ADDR, buffer);
 }
 
 function init() {
-    reader = i2cDevice.read(bmp280.BMP280_ADDRESS, 1,
+    var reader = i2cDevice.read(bmp280.BMP280_ADDRESS, 1,
                             bmp280.BMP280_REGISTER_CHIPID);
```
jimmy-huang: align bmp280

samples/I2CBMP280.js
```diff
@@ -138,38 +128,54 @@ function updateColor(newTemp) {
...
-    powerData.writeUInt8(bmp280.BMP280_REGISTER_CONFIG, 0);
-    powerData.writeUInt8(bmp280.BMP280_STANDBY | bmp280.BMP280_FILTER |
+    setupData.writeUInt8(bmp280.BMP280_REGISTER_CONFIG, 0);
+    setupData.writeUInt8(bmp280.BMP280_STANDBY | bmp280.BMP280_FILTER |
                          bmp280.BMP280_SPI_3W_DISABLE, 1);
 
-    i2cDevice.write(bmp280.BMP280_ADDRESS, powerData);
+    i2cDevice.write(bmp280.BMP280_ADDRESS, setupData);
 
     /* Setup Grove LCD */
```
jimmy-huang: shouldn't this comment go below the clear?

samples/I2CBMP280.js
```diff
@@ -213,32 +218,34 @@ function readTemperature() {
     //Select the first 20 bits of 32 bit integer and maintains the sign
     tempDec >>>= 12;
 
-    //Convert raw temperature to Celsius
-    var1 = (((tempDec >> 3) - (dig_T1 << 1))  * (dig_T2)) >> 11;
-    var2 = (((((tempDec >> 4) - (dig_T1)) * ((tempDec >> 4) - (dig_T1))) >> 12) *
+    // Convert raw temperature to Celsius using the algorithm found here
+    // https://www.bosch-sensortec.com/bst/products/all_products/bmp280
+    var var1 = (((tempDec >> 3) - (dig_T1 << 1))  * (dig_T2)) >> 11;
```
jimmy-huang: extra space before *

samples/I2CBMP280.js
```diff
@@ -213,32 +218,34 @@ function readTemperature() {
...
+    var var2 = (((((tempDec >> 4) - (dig_T1)) * ((tempDec >> 4) - (dig_T1))) >> 12) *
             (dig_T3)) >> 14;
```
jimmy-huang: align spaces here

samples/I2CBMP280.js
brianjjones: No, everything below that is part of setting up the LCD.  Clear is just the first line of  the code after that.

jimmy-huang:
LGTM, +1

samples/I2CBMP280.js
```diff
@@ -0,0 +1,251 @@
...
+
+// Global Buffer objects
+var setupData = new Buffer(2);
+var clearData = new Buffer([0, glcd.GLCD_CMD_SCREEN_CLEAR]);
+var msgData = new Buffer("Temperature is ");
+var tempData = new Buffer(" ##.#C");
+var changeData = new Buffer(" +");
+var redData = new Buffer([glcd.REGISTER_R, 0]);
+var greenData = new Buffer([glcd.REGISTER_G, 0]);
+var blueData = new Buffer([glcd.REGISTER_B, 0]);
```
grgustaf: Let's try allocating these where they're used and letting them run out of scope and get cleaned up. We'll need to make sure they really do get cleaned up and we reclaim the memory.

samples/I2CBMP280.js
```diff
@@ -0,0 +1,251 @@
...
+    var reader = i2cDevice.read(bmp280.BMP280_ADDRESS, 1,
+                                bmp280.BMP280_REGISTER_CHIPID);
+
+    // Check that we have a BMP280 on the I2C bus
+    if (reader.toString('hex') !== "58")
+        print("BMP280 not found! ChipId " + reader.toString('hex') !== "58");
+
+    readCoefficients();
+
+    //var powerData = new Buffer([bmp280.BMP280_REGISTER_CONTROL, 0]);
```
grgustaf: remove this commented code?

brianjjones:
@grgustaf FYI I ran the latest version of this overnight and it was still going this morning.  So the Buffer objects appear to be freeing their data properly.

samples/I2CBMP280.js
brianjjones: Done

samples/I2CBMP280.js
brianjjones: Done.

PR #274, opened 2016-10-10
[UART] UART implementation

jprestwo:
- Fixed 'then' in callbacks by acquiring the value. Without this
   promises did not work correctly if an object other than
   undefined was used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ endif
 			echo "HEAP_SIZE 5120" >> prj.mdef; \
 		fi; \
 	fi
-	@echo "ccflags-y += $(shell ./scripts/analyze.sh $(JS))" >> src/Makefile
+	@echo "ccflags-y += $(shell ./scripts/analyze.sh $(BOARD) $(JS))" >> src/Makefile
 
 .PHONY: all
 all: zephyr arc
--- a/docs/API.md
+++ b/docs/API.md
@@ -11,6 +11,8 @@ I/O
 
 [PWM](./pwm.md)
 
+[UART](./uart.md)
+
 Communications
 --------------
 [BLE](./ble.md)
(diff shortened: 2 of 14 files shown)


jprestwo:
@sergioamr, could you take a look at this compared to the "ashell" implementation. Long ago opening device "UART_0" and reading/writing to it worked, but these days it does not. I have looked at "ashell" for guidance but was unsure if it only works with ACM/USB or if it also works with regular UART (using a serial-to-USB adapter). It would be nice to have both for this module, but for starters I would just like to get UART working. The Zephyr sample does not appear to function properly either. We should discuss how "ashell" works and do a similar implementation here.

Note: this does work on QEMU, just not on the A101

jprestwo:
@jimmy-huang, @sergioamr:

I updated with buggy, but semi-working, code to run on the A101. Since standard UART does not appear to work on the A101 I had to use ACM like ashell does. This means that you have to open up a second terminal (e.g. ttyACM0) where all the `uart.write()` and `uart.on("read", ...)` data will go. This is inconvenient, and if someone has any suggestions I am all ears.

I say this is buggy because there are occasionally random characters that appear when reading data as well as "AT" characters that come in at the very beginning. @sergioamr, maybe you ran into this problem when developing ashell?

Also, I am seeing a problem with buffers regarding freeing up the resources. If I type too many characters before pressing "enter", it creates too many buffers resulting in an error. I am not sure how to solve this on account that `uart.write()` takes a buffer as input, so echo'ing characters back ends up creating a ton before any of them get freed. If there was a way to append to a buffer, it would solve my problem, but I don't think there is a way to do that currently.

brianjjones:
@jprestwo UART Sergio has some new code I don't believe he's merged yet.  And using that code I see UART work with USB0. 

Try  https://github.com/sergioamr/zephyr.js.git using branch next/sergio/ashell_rebase_master

jprestwo:
Do not merge this. I had to create a new PR: https://github.com/01org/zephyr.js/pull/274

PR #275, opened 2016-10-10
[debugging] Print elapsed time since start with DBG_PRINT/ERR_PRINT

jprestwo:

### DIFF
--- a/src/zjs_common.h
+++ b/src/zjs_common.h
@@ -10,11 +10,21 @@
 #define PRINT printf
 
 #ifdef DEBUG_BUILD
+
+int zjs_get_sec(void);
+int zjs_get_ms(void);
+
 #define DBG_PRINT \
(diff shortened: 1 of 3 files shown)


jprestwo:
@grgustaf, @jimmy-huang, @brianjjones:
This is a copy/rework of the previous PR that had to be backed out to fix ARC. This PR will work on ARC.

Another thing I noticed is that the debug flag does not get handed down to the ARC build, so all the DBG_PRINT's in the ARC code have no effect FYI. We can change this, but it would take some additional work if we want the elapsed time prints, since the Zephyr variant requires an updater loop. I don't foresee any problems with doing this on ARC, it would just need to be done. If we want this functionality with this PR let me know and I can investigate and implement.

jimmy-huang:
@jprestwo that's good to know, yeah, I think this will be needed, but not with this PR, feel free to create another PR to fix this.

jimmy-huang:
+1

PR #279, opened 2016-10-11
[callbacks] Fixed callback size increment

jprestwo:


poussa:
What is the issue this is fixing?

jprestwo:
@poussa, the callback list size was getting incremented every time a callback was added. This should only happen when the callback list is full. When a callback is removed it opens up a slot for a future callback, so new callbacks that take up an existing slot should not increment the size. This change checks to see if a callback being added is taking up an existing slot or a new one.

poussa:
OK, LGTM

PR #287, opened 2016-10-12
[callbacks] Callback queue implementation

jprestwo:
Signed-off-by: James Prestwood james.prestwood@intel.com


jprestwo:
@grgustaf, this is an updated PR with the callback queue/ring buffer stuff. I could not quite remember how to reproduce the issue we were having. I know that the root of the problem was that when you signal a callback that has already been signaled, it will not be called twice as it should. To fix this I added a queue for callbacks so signaling one actually pushes a new element into the queue. With this change I also added a ring buffer for the callback arguments, so if you signal a callback twice, but the arguments have changed, the args get pushed onto the ring buffer. This change runs and everything but you'll have to refresh my memory on what sample actually shows this problem in action so I can see if this fixes it.

jprestwo:
Please do not merge this yet. We need to talk about how its implemented and decide which route to go. Next Tuesday we can hash it out.

samples/arduino/starterkit/Spaceship2.js
```diff
@@ -41,31 +43,34 @@ var btnstate = false;
...
+var last;
 
 button.onchange = function (event) {
     if (event.value) {
-        // button has been pressed
-        toggle = false;
-        led1.write(false);
-        led2.write(false);
-        led3.write(true);
+        if (timer == null) {
```
jprestwo: I had to put this in because, even using JUST Zephyr samples, a GPIO input can trigger the same value multiple times. If this happens, this top if block will get called causing two `setInterval`'s to trigger. For this reason I set `timer` to null once the interval is cleared. This provides synchronization between `setInterval` and `clearInterval`

src/zjs_callbacks.c
```diff
@@ -387,57 +418,112 @@ void print_callbacks(void)
 
 void zjs_call_callback(int32_t i)
```
jprestwo: This function got ugly. We need to decide if we want to have both pre/post callbacks AND a ring buffer, or just the ring buffer.

samples/arduino/starterkit/Spaceship2.js
```diff
@@ -17,18 +17,20 @@
...
     edge: 'any'
 });
 var led1 = gpio.open({
-    pin: pins.IO4,
+    pin: pins.LED0,
     direction: 'out'
 });
 var led2 = gpio.open({
-    pin: pins.IO7,
+    pin: pins.LED1,
```
grgustaf: You probably didn't mean to include these changes, since it was just the way you were testing locally? But these are following the Arduino sample code which uses external LEDs so the pins used should not be changed.

src/zjs_callbacks.c
```diff
@@ -315,6 +334,20 @@ void zjs_remove_callback(int32_t id)
...
 
+void zjs_signal_callback_args(int32_t id, void* args, uint32_t size)
+{
+    if (args && size) {
+        int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id, 0, (uint32_t*)args, (uint8_t)size / 4);
+        if (ret != 0) {
+            DBG_PRINT("error putting into ring buffer, ret=%u\n", ret);
+        }
+    }
+    zjs_port_queue_put(&service_queue, cb_map[id]);
```
grgustaf: This is a problem because it will overwrite the void *reserved (linked list pointer) in cb_map[id] the second time it is called for the same callback.

src/zjs_callbacks.c
```diff
@@ -315,6 +334,20 @@ void zjs_remove_callback(int32_t id)
...
+
+    cb_map[id]->arg_sz = size;
```
grgustaf: This is also not going to work if the size could possibly be different from call to call, because you only have one copy of this struct where you're writing to. So I think maybe the callback ID and arg count need to be pushed into the ring buffer along with the args, and use it as a pending callback buffer.

src/zjs_callbacks.c
```diff
@@ -315,6 +334,20 @@ void zjs_remove_callback(int32_t id)
     }
 }
 
+void zjs_signal_callback_args(int32_t id, void* args, uint32_t size)
+{
+    if (args && size) {
+        int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id, 0, (uint32_t*)args, (uint8_t)size / 4);
```
grgustaf: You already have id here and you could put arg count in the field where you're using 0... then you have enough info in here I think to use this ring buffer as your full "pending callbacks" buffer, and key everything off this.

jprestwo: I think we discussed this on IRC but I will reiterate it here. The only problem with using ONLY the ring buffer as the pending callback queue is that you have the know the size of the ring buffer element before hand, which is why I use the queue to hold that size. I agree that this is not correct because you could push two of the same callback ID's with different argument sizes. We essentially need a Zephyr API that lets you get the size of the next ring buffer element, then we could create an array that is big enough to hold the argument data. You had said that we could just have a set size (like 4 or 5) but this seem like more of a hack and not very efficient. I think it would be quite simple to add a Zephyr API which lets you get the size of the next ring buffer element (or we could just hack it and look into the ring buffer's data and find the size.

grgustaf: No, that's not what I said. Zephyr *can* tell you the size of the next item in the ring buffer. If you pass it a '0' for the size, and there were arguments stored, it will tell you your buffer is not big enough and return you the size you need. So it's a bit inconvenient, but it already does this.

Having the set size like 4 or 5 that I talked about is not inefficient. That space is not being wasted inside the ring buffer. It is only when you go to read it out, you have your waiting buffer be that size so that it can handle any possible output. (Like having a 256 char buf to receive a string even though most of the time the string will be 10-15 chars.)

However, you did point out that in some cases like the timer functions the number of arguments is defined by the app writer, not by us. So there we have two options: either a configurable MAX_ARGS that they need to increase in Zephyr if they're going to use more... or we malloc this read buffer when our default is not big enough.

jprestwo: ah, yes your right. I misread the implementation and did not see the line setting size32 to length:
```
if (header->length > *size32) {
    *size32 = header->length;
    return -EMSGSIZE;
}
```
So yes, instead of pulling off the queue, we can just do a while loop over the ring buffer get function and set the size to zero. If the size comes back as > 0 then there is an item in the buffer and we can pull off of it, completely doing away with the queue and solving the duplicate argument problem. Sorry for the confusion.

grgustaf: Just be careful because one other weirdness is that even if you pass size 0, if there were *no* arguments, and it all you wrote was the type/value stuff, it will still pop it off. (It's still able to read even though you didn't supply a buffer basically.)

src/zjs_aio.c
```diff
@@ -278,8 +268,7 @@ static jerry_value_t zjs_aio_pin_read_async(const jerry_value_t function_obj,
 
     handle->pin_obj = this;
     handle->callback_id = zjs_add_callback(argv[0], this, handle,
-                                           zjs_aio_pre_callback,
-                                           zjs_aio_free_callback);
+            zjs_aio_free_callback);
```
grgustaf: indent is strange

src/zjs_callbacks.c
```diff
@@ -86,6 +92,8 @@ void zjs_init_callbacks(void)
         }
         memset(cb_map, 0, size);
     }
+    zjs_port_ring_buf_init(&ring_buffer, 1024, (uint32_t*)args_buffer);
```
grgustaf: Why did you switch from the DECLARE_POW2 method? That seemed sort of safer to ensure we get optimized access, in case Joe Maintenance Programmer alters things. :)

src/zjs_callbacks.c
```diff
@@ -315,17 +312,21 @@ void zjs_remove_callback(int32_t id)
...
 {
-    if (id != -1 && cb_map[id]) {
-#ifdef DEBUG_BUILD
-        if (cb_map[id]->type == CALLBACK_TYPE_JS) {
-            DBG_PRINT("signaling JS callback id %ld\n", id);
-        } else {
-            DBG_PRINT("signaling C callback id %ld\n", id);
+    if (args && size) {
+        // signal with args
+        int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id, 0, (uint32_t*)args, (uint8_t)size / 4);
```
grgustaf: This should be (size + 3) / 4 I think to avoid rounding error. E.g. if you passed in one byte of args, this would turn into 0 dwords.

src/zjs_callbacks.c
```diff
@@ -315,17 +312,21 @@ void zjs_remove_callback(int32_t id)
...
+        if (ret != 0) {
+            DBG_PRINT("error putting into ring buffer, ret=%u\n", ret);
+        }
+    } else {
+        // signal without args
+        uint32_t dummy = 0;
+        int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id, 0, (uint32_t*)&dummy, (uint8_t)1);
```
grgustaf: Here you can/should pass 0 for size and NULL is fine for dummy as it won't be used. This way you're wasting an extra 4 bytes in the ring buffer.

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
-            uint32_t argc = 0;
-            jerry_value_t ret_val;
-            jerry_value_t* args = NULL;
-
-            if (cb_map[i]->js->pre) {
-                args = cb_map[i]->js->pre(cb_map[i]->js->handle, &argc);
-            }
-
-            DBG_PRINT("calling callback list id %ld with %lu args\n", cb_map[i]->js->id, argc);
+    if(cb_map[id]) {
```
grgustaf: space

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
+        if (cb_map[id]->type == CALLBACK_TYPE_JS) {
+            if (cb_map[id]->js->func_list == NULL && jerry_value_is_function(cb_map[id]->js->js_func)) {
+                jerry_value_t ret_val;
+                // Single function callback
+                if (sz) {
+                    ret_val = jerry_call_function(cb_map[id]->js->js_func, cb_map[id]->js->this, data, sz);
+                } else {
+                    ret_val = jerry_call_function(cb_map[id]->js->js_func, cb_map[id]->js->this, NULL, 0);
+                }
```
grgustaf: Should be able to drop this else clause. They should be passing data == NULL already anyway right? And if you pass size 0 it shouldn't read data anyway. (BTW you might want to update zjs_callbacks.h jsdoc to define the data/size parameter for this function and there you could be explicit that you pass NULL if it's 0 size, if you want.

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
 
-            for (j = 0; j < cb_map[i]->js->num_funcs; ++j) {
-                ret_val = jerry_call_function(cb_map[i]->js->func_list[j], cb_map[i]->js->this, args, argc);
-                if (jerry_value_has_error_flag(ret_val)) {
-                    DBG_PRINT("callback %ld returned an error for function[%i]\n", i, j);
+                for (i = 0; i < cb_map[id]->js->num_funcs; ++i) {
+                    if (sz) {
+                        ret_val = jerry_call_function(cb_map[id]->js->func_list[i], cb_map[id]->js->this, data, sz);
+                    } else {
+                        ret_val = jerry_call_function(cb_map[id]->js->js_func, cb_map[id]->js->this, NULL, 0);
```
grgustaf: I think you meant to call func_list[i] here too? And again this should be the same as above so the if/else can be dropped I think.

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
+                    cb_map[id]->js->post(cb_map[id]->js->handle, &ret_val);
                 }
             }
-            if (cb_map[i]->js->post) {
-                cb_map[i]->js->post(cb_map[i]->js->handle, &ret_val);
+        } else if (cb_map[id]->type == CALLBACK_TYPE_C && cb_map[id]->c->function) {
+            if (sz) {
+                cb_map[id]->c->function(cb_map[id]->c->handle, data);
+            } else {
+                cb_map[id]->c->function(cb_map[id]->c->handle, NULL);
```
grgustaf: Same thing here I think?

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
-            zjs_call_callback(i);
+    if (ring_buf_initialized) {
+        uint16_t type;
+        uint8_t value;
+        uint8_t size = 0;
+        uint32_t dummy;
+        while (zjs_port_ring_buf_get(&ring_buffer,
+                                     &type,
+                                     &value,
+                                     &dummy,
```
grgustaf: Again here you can drop dummy var and just pass NULL. 98% sure, trust but verify. :)

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
+                                     &size) == -EMSGSIZE) {
+            if (size == 0) {
+                break;
+            } else {
+                // Pull from ring buffer
+                uint16_t id;
+                uint8_t value1;
+                uint8_t sz = size;
+                jerry_value_t data[sz];
```
grgustaf: Brilliant, I forgot about that. I thought you'd have to malloc. Nice!

src/zjs_callbacks.c
```diff
@@ -387,64 +386,86 @@ void print_callbacks(void)
...
+                int ret = zjs_port_ring_buf_get(&ring_buffer,
+                                                &id,
+                                                &value1,
+                                                (uint32_t*)data,
+                                                &sz);
+                if (ret != 0) {
+                    DBG_PRINT("(JS) error pulling from ring buffer: ret = %u\n", ret);
+                    break;
+                }
+                zjs_call_callback(id, data, sz);
```
grgustaf: Oops, this is a bug. If there are no args, the while will not return -EMSGSIZE... it will work! That was a gotcha I mentioned. So you need to check for that case and still call the callback. If I'm not mistaken, you're not calling any 0-arg functions right now.

(Seems surprising guess I'll go test it when I'm done here.)

So... only call buf_get the second time *IF* there were args and it was needed. But either way, call the callback.

src/zjs_callbacks.h
```diff
@@ -178,7 +174,7 @@ int32_t zjs_add_c_callback(void* handle, zjs_c_callback_func callback);
  *
  * @param i             ID of callback
```
grgustaf: Add the descriptions here like I mentioned above.

src/zjs_callbacks.c
grgustaf: James answered that it was to simplify implementing the linux port. We should look into whether it changes things or not, to make sure we're getting the good performance, but we could save that task for later. Also, this accidentally changed the buffer from 4096 bytes to 1024 bytes vs. the first version of the patch, but maybe that's enough. We'll need to profile to get a sense of a good default at some point.

src/zjs_callbacks.c
jprestwo: I don't think you can put in "nothing" into the ring buffer. The Zephyr implementation actually doesn't check for NULL and de-references the pointer, so that will segfault, but I don't think passing in no data was a use case for their ring buffer implementation.

src/zjs_callbacks.c
grgustaf: FWIW I opened issue #337 to track that last point.

src/zjs_callbacks.c
grgustaf: You can't put "nothing", but you can put just type/value, and not any additional 32-bit chunks (size = 0). Please download this code and run `make qemu` in `experiments/ringbuf`.

https://github.com/grgustaf/zephyr.js-1/tree/ringbuf-experiment

Ask me in IRC if anything is unclear.

src/zjs_callbacks.c
grgustaf: Still a good idea just to be safe. Then you don't *need* to document that they have to use something divisible by four; it will work even if they don't.

src/zjs_callbacks.c
```diff
@@ -387,64 +385,76 @@ void print_callbacks(void)
...
-    for (i = 0; i < cb_size; i++) {
-        if (cb_map[i] && cb_map[i]->signal) {
-            cb_map[i]->signal = 0;
-            zjs_call_callback(i);
+    if (ring_buf_initialized) {
+        uint16_t type;
+        uint8_t value;
+        uint8_t size = 0;
+        uint32_t dummy;
+        // setting size = 0 will check if there is an item in the ring buffer
```
grgustaf: Yes, but it will also FETCH the item in the ring buffer if it has JUST a type and value set, not any arguments.

src/zjs_callbacks.c
grgustaf: So actually once you resolve this you don't need this second call to buf_put. You can just use the same code as above.

If there are no args, size is 0. size + 3 / 4 is still 0. And as my experiment above will show you, you *can* write an object to the ring buffer with size 0. It will write a 32-bit value with 16-bit "type" (id), 8-bit "value" (you've given 0), and an 8-bit "size" (0 in this case).

src/zjs_callbacks.c
```diff
@@ -313,19 +313,9 @@ void zjs_remove_callback(int32_t id)
...
-    } else {
-        // signal without args
-        uint32_t dummy = 0;
-        int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id, 0, (uint32_t*)&dummy, (uint8_t)1);
-        if (ret != 0) {
-            DBG_PRINT("error putting into ring buffer, ret=%u\n", ret);
-        }
+    int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id, 0, (uint32_t*)args, (uint8_t)((size + 3) / 4));
+    if (ret != 0) {
+        DBG_PRINT("error putting into ring buffer, ret=%u\n", ret);
```
grgustaf: Sorry, I hate to keep making comments... but I think this needs to be a real error, not just debug. If we weren't able to put then the callback is lost, and the user needs to have an indication (at least if they're watching serial) that their queue is full and may need to be increased or their algorithm throttle work.

jprestwo: I totally agree, I meant to use ERR_PRINT, but I am still getting used to ERR_PRINT being a macro now days :)

src/zjs_callbacks.c
```diff
@@ -424,35 +414,39 @@ void zjs_call_callback(int32_t id, void* data, uint32_t sz)
...
-        // setting size = 0 will check if there is an item in the ring buffer
-        while (zjs_port_ring_buf_get(&ring_buffer,
-                                     &type,
-                                     &value,
-                                     &dummy,
-                                     &size) == -EMSGSIZE) {
-            if (size == 0) {
-                break;
-            } else {
+        while (1) {
```
grgustaf: Hmm, we could probably get into an infinite loop here if some callback triggers another callback to be signaled? We probably can't find out how many callbacks are in the buffer. We could have a fixed maximum then, or can we do better?

grgustaf: Or maybe what we need to do is read all the callbacks out first and then call them, stopping when we get to the end. O_o

jprestwo: I like the idea of having a max that can be serviced at once. That makes more sense to me. Reading them all out at once would be a difficult change and a lot of overhead. What do you think? Just add a count and only service at most N callbacks?

jprestwo:
@jimmy-huang, I reverted one of the commits that caused the problem we were seeing today. If you could try out what I have pushed up now (with the commit reverted) and see if it fixes the problem, it did for me.

@grgustaf, unfortunately the commit I reverted was the one that removed the "dummy" uint32_t data that we pushed on when the callback had zero arguments. The reason I had to add this back was because when pulling off the ring buffer, if the item had a length of zero, it would not actually pull any of the data off because it would always hit that first if block and return. This seems like something that should be fixed in Zephyr, but for now we will just have to go with the way I had it and push on 4 bytes of dummy data for zero argument callbacks.

src/zjs_callbacks.c
grgustaf: Yeah, I think that would work. And I think it should be at least 64 because you can fit that many callbacks in your 1024-byte buffer even if each takes up 16 bytes. I thought I remembered you'd added this in a follow-on patch but I don't see it now and am confused. :)

grgustaf: Ah that's the one you reverted. Okay.

grgustaf:
In the code snippet you gave, header->length I believe is 0, so it is not bigger than *size32 (0), which is why it does get past that and work (as evidenced by test 3 of my sample experiment).


Here's the code from `sys_ring_buf_put` where it makes sure there's room for one extra block than requested (so 1 block when you ask for 0) and then sets length to 0 as I'm suggesting above.

```c
	if (space >= (size32 + 1)) {
                ...
		header->length = size32;
```

jprestwo:
@grgustaf, ah yes you are right. That must not be the cause then. For some reason with zero argument callbacks, data is not getting pulled out of the ring buffer so it is filling up, but when I used the "dummy" mechanism it seemed to work. I just need to look more into it.

src/zjs_callbacks.c
```diff
@@ -387,64 +396,113 @@ void print_callbacks(void)
...
+                                        &id,
+                                        &value,
+                                        NULL,
+                                        &size);
+            if (ret == -EMSGSIZE || ret == 0) {
+                // item in ring buffer with size > 0, has args
+                // pull from ring buffer
+                int ret1;
+                uint16_t id1;
+                uint8_t value1;
```
grgustaf: You're only overwriting these three in the -ENOMSGSIZE case so you don't actually need new variables, you can overwrite the old ones.

src/zjs_callbacks.c
```diff
@@ -387,64 +396,113 @@ void print_callbacks(void)
...
+                uint8_t sz = size;
+                jerry_value_t data[sz];
+                if (ret == -EMSGSIZE) {
+                    ret1 = zjs_port_ring_buf_get(&ring_buffer,
+                                                &id1,
+                                                &value1,
+                                                (uint32_t*)data,
+                                                &sz);
```
grgustaf: Indents off here.

grgustaf:
+1 after plenty of review, and testing AIO.js, Timers.js, BTGrove demo, I2C.js

PR #291, opened 2016-10-13
[Demo] Improve BLE demo stability

jimmy-huang:
It seems that the temperature value change update is causing random
update issue and dropped connection, changed the demo code to
only send updates if a value change is detected.

Also increase the interval between updates to a minimum of 2sec.

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -14,8 +14,8 @@
 #include "zjs_ipm.h"
 
 #define QUEUE_SIZE       10  // max incoming message can handle
-#define SLEEP_TICKS       1  // sleep time in cpu ticks
-#define UPDATE_INTERVAL 200  // interval in between notifications
+#define SLEEP_TICKS       1  // 10ms sleep time in cpu ticks
+#define UPDATE_INTERVAL 200  // 2sec interval in between notifications
 
 // AIO
 #define ADC_DEVICE_NAME "ADC_0"
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -13,9 +13,9 @@
 #include "zjs_common.h"
 #include "zjs_ipm.h"
 
-#define QUEUE_SIZE      10  // max incoming message can handle
-#define SLEEP_TICKS      1  // sleep time in cpu ticks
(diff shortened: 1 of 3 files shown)


arc/src/main.c
```diff
@@ -13,9 +13,9 @@
 #include "zjs_common.h"
 #include "zjs_ipm.h"
 
-#define QUEUE_SIZE      10  // max incoming message can handle
-#define SLEEP_TICKS      1  // sleep time in cpu ticks
-#define UPDATE_INTERVAL 50  // interval in between notifications
+#define QUEUE_SIZE       10  // max incoming message can handle
+#define SLEEP_TICKS       1  // sleep time in cpu ticks
+#define UPDATE_INTERVAL 200  // interval in between notifications
```
kenchris: State in what unit? I would not guess 200 to be very 2 seconds

samples/WebBluetoothDemo.js
```diff
@@ -148,9 +148,8 @@ ble.on('stateChange', function(state) {
 });
 
 ble.on('advertisingStart', function(error) {
-    print('advertisingStart: ' + (error ? error : 'success'));
-
     if (error) {
+        print("Failed to advertise physical web");
```
kenchris: ... Physical Web URL.

samples/WebBluetoothDemo.js
```diff
@@ -163,25 +162,31 @@ ble.on('advertisingStart', function(error) {
...
+    var lastTemp = 0;
 
     tmp36.on("change", function(data) {
         var voltage = (data / 4096.0) * 3.3;
         var celsius = (voltage - 0.5) * 100 + 0.5;
         celsius = celsius | 0;
 
-        print("Temperature change: " + celsius + " degrees Celsius");
-        TemperatureCharacteristic.valueChange(celsius);
+        if (celsius != lastTemp) {
```
kenchris: !== ?

jimmy-huang:
Updated PR based on @kenchris feedback

poussa:
LGTM, based on the feedback from @kenchris this is improving the demo.

PR #293, opened 2016-10-14
[callbacks] Fix for DBG_PRINT

brianjjones:
Fixes a DBG_PRINT which fails to compile when debugging is turned
on because j is an int, not a long unsigned.


jimmy-huang:
+1

poussa:
+1

PR #294, opened 2016-10-15
[memory] Creating functions to free memory

brianjjones:
Creating a free all memory function for each section of code that
uses malloc.  This will allow ashell to free any memory that
JerryScript failed to free when the JavaScript is halted and
restarted.

### DIFF
--- a/src/zjs_timers.c
+++ b/src/zjs_timers.c
@@ -108,7 +108,7 @@ static bool delete_timer(int32_t id)
     return false;
 }
 
-void zjs_timers_free_timers()
+void zjs_timers_cleanup()
 {
     for (zjs_timer_t **ptm = &zjs_timers; *ptm; ptm = &(*ptm)->next) {
         zjs_timer_t *tm = *ptm;
--- a/src/zjs_timers.c
+++ b/src/zjs_timers.c
@@ -108,6 +108,23 @@ static bool delete_timer(int32_t id)
     return false;
 }
 
+void zjs_timers_free_timers()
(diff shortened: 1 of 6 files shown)


src/zjs_timers.c
```diff
@@ -108,6 +108,23 @@ static bool delete_timer(int32_t id)
     return false;
 }
 
+void zjs_timers_free_timers()
+{
+    for (zjs_timer_t **ptm = &zjs_timers; *ptm; ptm = &(*ptm)->next) {
+        zjs_timer_t *tm = *ptm;
+        int i;
```
poussa: Move 'i' down to the for loop (for (int i = 0...) since that is where it is only used.

jimmy-huang:
shouldn't we name these functions _destruct() or _clean()?

brianjjones:
I'm fine with either name if that makes more sense to people. @grgustaf you got a function naming preference?

brianjjones:
This compiles for me, not sure why travis-ci is flagging it.  Looks like it isn't quite set up right?

poussa:
Sorry my bad. I enabled travis accidentally today. Now it is off again. Need to set it up properly.

grgustaf:
They look good, but we don't need them unless they'll be called, so maybe we wait till you get your jerry_cleanup stuff sorted out?

brianjjones:
This request has been replaced with https://github.com/01org/zephyr.js/pull/336 which solves the problem more elegantly.

PR #299, opened 2016-10-18
[Makefile] Denote generated files with a comment and clean up on clean

grgustaf:
The prj.conf, prj.mdef, arc/prj.conf, and src/Makefile files are all
generated at build time, so add a comment to each to that effect and
delete them in make clean.


grgustaf:
Mistakenly requested against master.

PR #304, opened 2016-10-19
[deps] Updated JerryScript and Zephyr to latest (as of 10/19/16)

jprestwo:


Makefile
```diff
@@ -17,9 +17,11 @@ TRACE ?= off
 # Specify pool malloc or heap malloc
 MALLOC ?= pool
 
+# TODO: The latest Zephyr seems to require running make mrproper for each build
+#       the mrproper target should be removed when this is fixed
 # Build for zephyr, default target
 .PHONY: zephyr
-zephyr: analyze generate
+zephyr: mrproper analyze generate
```
jprestwo: It turns out that running "make mrproper" is not very detrimental to the build time (unlike e.g. make clean), so we may just want to leave this here.

pfalcon:
How is one supposed to upgrade from current devel to this patch? Here's my transcript:

~~~
$ make update
Found dependency: jerryscript
error: pathspec 'fd98d649b' did not match any file(s) known to git.
^Cmake: *** wait: No child processes.  Stop.
make: *** Waiting for unfinished jobs....
make: *** wait: No child processes.  Stop.
pfalcon@x230:~/projects-3rdparty/Embedded/Zephyr/zephyr.js$ make setup
Found dependency: jerryscript
error: pathspec 'fd98d649b' did not match any file(s) known to git.
You are not currently on a branch. Please specify which
branch you want to merge with. See git-pull(1) for details.

    git pull <remote> <branch>

Error attempting git pull on jerryscript
Found dependency: zephyr
error: pathspec '34ffa15669' did not match any file(s) known to git.
^Cmake: *** [update] Interrupt

pfalcon@x230:~/projects-3rdparty/Embedded/Zephyr/zephyr.js$ make update
Found dependency: jerryscript
Found dependency: zephyr
error: pathspec '34ffa15669' did not match any file(s) known to git.
^Cmake: *** [update] Interrupt

pfalcon@x230:~/projects-3rdparty/Embedded/Zephyr/zephyr.js$ make setup
Found dependency: jerryscript
Found dependency: zephyr
error: pathspec '34ffa15669' did not match any file(s) known to git.
~~~

As can be seen, I was scared by errors and broke and restarted process several times. Not very friendly. Problems like that could have been alleviated is standard git submodules were used (yup, they may have their own problems, but always less than somebody's adhoc solution).

pfalcon:
Building this patch with `make BOARD=frdm_k64f JS=samples/ButtonLEDs-k64f.js` leads to:

~~~
  AR      libzephyr.a
Building libjerry-core.a
make -C /home/pfalcon/projects-3rdparty/Embedded/Zephyr/zephyr.js/deps/jerryscript -f targets/zephyr/Makefile.zephyr BOARD=frdm_k64f jerry
targets/zephyr/Makefile.zephyr:52: build/frdm_k64f/zephyr/.config: No such file or directory
gcc: error: unrecognized argument in option ‘-mabi=aapcs’
gcc: note: valid arguments to ‘-mabi=’ are: ms sysv
gcc: warning: ‘-mcpu=’ is deprecated; use ‘-mtune=’ or ‘-march=’ instead
~~~

pfalcon:
`make linux` seems to work though.

pfalcon:
To exclude unrelated, JrS upstream master itself also builds fine against Zephyr master.

pfalcon:
Ah, right, I probably didn't re-run all the environment setup scripts for the new dependency versions. Retrying.

pfalcon:
Nope, still the same.

grgustaf:
I see the same problem, we'll figure it out. @pfalcon could you clarify what you mean by standard git submodules? Maybe we're missing some process we should adopt. :)

grgustaf:
Aha instead of our deps/ I guess. Hmm, I didn't know about this. So perhaps yes.

poussa:
Yeah, we should switch to git submodules.

pfalcon:
> Aha instead of our deps/ I guess. Hmm, I didn't know about this. So perhaps yes.

Yep, git's native mechanism of pulling external repositories into a working tree (git submodule --help). Here's how IoT.js uses them for example: https://github.com/Samsung/iotjs/tree/master/deps

pfalcon:
Ok, so I have changes to build this. Will spool them next (includes patch for JrS). There're already visible improvements - pressing one of buttons on frdm_k64f just hanged a running application, now it doesn't, and even can be used (after correcting pin number, will submit too).

poussa:
Can you send the JrS patch to upstream? We don't want to maintain those if possible.

pfalcon:
> Can you send the JrS patch to upstream? We don't want to maintain those if possible.

Sure, that went as https://github.com/Samsung/jerryscript/pull/1404

And then https://github.com/01org/zephyr.js/pull/313 one-liner on top of this patch.

One note is it's definitely not the only way to fix the issue, but it's arguably the simplest one: one extra var to pass on Z.js side, factor out common default suffix on JrS side. Another approach would be to rework the way JrS and Zephyr (and then likely Z.js and Z) integrate at all. But that would be much more involved way ;-).

poussa:
Thanks, let's wait if/when JrS patch lands and start merging these. I'll take s closer look tomorrow.

jprestwo:
@sergioamr,

Hey Sergio, you have had some experience with this and I cannot for the life of me figure out the issue. I updated JerryScript and Zephyr to the latest, but when I run an application I get an immediate crash:
```
***** CPU exception 7
Current thread ID = 0xa800e120
Faulting segment:address = 0x00000008:0x40010af8
eax: 0x00000008, ebx: 0xa8006420, ecx: 0x00000001, edx: 0xa800e928
esi: 0x00000000, edi: 0xa8006774, ebp: 0a8006420, esp: 0xa800e7b4
eflags: 0x00010293
Fatal task error! Aborting task.
```

Do you have any suggestions on fixing this?

poussa:
@jprestwo I have the same crash in this branch. I think it is some config flag missing or something. Don't have time now to dig in but @jimmy-huang, @brianjjones and @grgustaf may also have a helping hand.

pfalcon:
To note it, https://github.com/Samsung/jerryscript/pull/1404 was merged. So, would be nice to update JrS revision in this patch to at least fdf9e71caad5af1018fb312edde81494d6af05b9.

jprestwo:
@poussa, @grgustaf, @jimmy-huang,

This runs on Zephyr without crashing now. The problem was that JerryScript was crashing when trying to run `print()` in JavaScript. I fixed this by re-defining our own `print()` function that uses `printk` on Zephyr and `printf` on Linux

jimmy-huang:
@jprestwo this is good, we should be using console.log() instead of print() anyway

jprestwo:
@grgustaf, @jimmy-huang, @poussa,

Do not merge this at this time. `snprintf` causes the A101 to crash when using newlib. I have filed a JIRA on Zephyr: https://jira.zephyrproject.org/browse/ZEP-1118

jprestwo:
@poussa, I have updated JerryScript to the latest as of today (10/25/16) at 11am.

sergioamr:
Let me have a look, it rings a bell but it might just be that we were talking about it on the IRC :D

poussa:
I spend some time on this today. 

@jprestwo can you rebase and check the travis. Also, the make clean needs to work properly. Now it leaves outdir/<board-name> in place. When removing outdir by hand, you then need to run make mrproper in zephyr tree.

@pfalcon The new JrS Zephyr build system with outputexports was not working for me when ccache was enabled. I had to teach Zephyr not to use it (USE_CCACHE=0) since JrS cmake was using 

```
 -DEXTERNAL_CMAKE_C_COMPILER=ccache /opt/zephyr-sdk/sysroots/i686-pokysdk-linux/usr/bin/arm-poky-eabi/arm-poky-eabi-gcc \
 ````
And complaining

```
CMake Error: The source directory "/opt/zephyr-sdk/sysroots/i686-pokysdk-linux/usr/bin/arm-poky-eabi/arm-poky-eabi-gcc" is a file, not a directory.
```

**Comment by pfalcon** [comment]
> @pfalcon The new JrS Zephyr build system with outputexports was not working for me when ccache was enabled. I had to teach Zephyr not to use it (USE_CCACHE=0) since JrS cmake was using

I see, can't add much, I didn't try ccache with Zephyr/Zephyr.js so far. @daniel-thompson, FYI ^^. @poussa , if you think that disabling ccache is acceptable, let's go that way for now. Otherwise, we can upgrade to JrS revision previous to outputexports for now. We with @daniel-thompson definitely will look what may be an issue, but it may take some time. Thanks.

**Comment by daniel-thompson** [comment]
@pfalcon, @poussa : *Fixing* the ccache build should be pretty quick (just use $(subst ...) to get rid of ccache again). Note that this will restore old behaviour and is not the same as making the JerryScript build *use* ccache. I don't think the jerryscript half of the build system has ever honoured USE_CCACHE=1 anyway. I have a patch for this which I am testing now.

@poussa, @jprestwo: Not quite sure what is meant by "the make clean needs to work properly". Expecting "clean" to remove **all** build artefacts is really problematic for zephyr because it clobbers the .config file (in which the user may have enabled extra debug sessions). Normal zephyr applications do not remove all build artefacts using the clean rule.

**Comment by pfalcon** [comment]
@daniel-thompson: Thanks for looking into that.

> Expecting "clean" to remove all build artefacts is really problematic for zephyr because it clobbers the .config file (in which the user may have enabled extra debug sessions). Normal zephyr applications do not remove all build artefacts using the clean rule.

Just my 2 cents: I'm personally not sure that it's 100% grounded to let Zephyr build system's "traits" bleed thru to a higher-level project which integrates Zephyr. If in higher-level project "make clean" makes "remove any generated files" (like it usually is), then bringing in Zephyr's behavior is pretty confusing.

**Comment by daniel-thompson** [comment]
This is not *just* zephyr build system traits, to be honest I think the expectation the clean cleans up everything is bogus (autotools, cmake, etc).

For sure no system that uses kbuild nukes the config files during a clean... to do so is deeply user hostile!


[1] Or
Likewise 

Actually the same applied to autotools based build systems as well. make clean doesn't nuke configs there either.

**Comment by pfalcon** [comment]
@daniel-thompson : It depends on what you consider "config". While for linux kernel it's definitely .config, for Zephyr it's more like prj.conf (== config fragment from which .config is generated). I add my user settings to prj.conf, not .config. Maybe I got it wrong of course. And I don't want to spend too much time on that here, just took a chance to bring it up. I'm curious what decision Zephyr.js team will make though.

File: Makefile [hunk 2]
```diff
@@ -154,16 +159,15 @@ endif
 .PHONY: clean
 clean:
 	@if [ -d deps/jerryscript ]; then \
-		make -C $(JERRY_BASE) -f targets/zephyr/Makefile.zephyr clean; \
-		make -C $(JERRY_BASE) -f targets/zephyr/Makefile clean; \
+		make -C $(JERRY_BASE) -f targets/zephyr/Makefile clean BOARD=$(BOARD); \
```
grgustaf: This seems a bit odd. I usually wouldn't pass a BOARD to make clean, so this will probably only clean a101?

Makefile
```diff
@@ -154,16 +159,15 @@ endif
...
 		rm -rf deps/jerryscript/build/$(BOARD)/; \
 		rm -rf deps/jerryscript/build/lib; \
 	fi
 	@if [ "$(TARGET)" != "linux" ]; then \
 		if [ -d deps/zephyr ] && [ -e src/Makefile ]; then \
-			make -f Makefile.zephyr clean; \
+			cd deps/zephyr; make clean BOARD=$(BOARD); \
```
grgustaf: Same thing here with BOARD.

src/zjs_common.h
```diff
@@ -6,8 +6,14 @@
...
 #include <stdio.h>
-
+#ifndef ZJS_LINUX_BUILD
+#include <misc/printk.h>
+#define PRINT printk
+#else
 #define PRINT printf
+#endif
+
+
```
grgustaf: two extra lines here

Makefile
grgustaf: Okay, yeah make pristine rm's the outdir; I added a ZJS pristine target to call that too.

daniel-thompson:
On 27/10/16 18:03, Paul Sokolovsky wrote:
> completely, rm -rf is probably the best choice (see also our discussion
> with daniel-thompson below).

Actually I am happy to concede that I think I've made a mistake 
regarding "make pristine"... That **should** nuke the whole of the 
build/ directory but today I don't think it does.

grgustaf:
Hey, @daniel-thompson, since you're still awake, I think we noticed a problem with your JrS patch that went in today. Makefile.zephyr calls include(Makefile.export) near the top of the file. But later there's a rule to make that file with "make outputexports". So that's not working. I get an error that it doesn't exist, and I think that's why other stuff is failing. We're giving up for now and merging this patch based on an older version of JrS but will try to help figure that out.

grgustaf:
Oh shoot, now I see the conversation @pfalcon referred to and I missed the whole thing. /me catches up.

pfalcon:
> Makefile.zephyr calls include(Makefile.export) near the top of the file. But later there's a rule to make that file with "make outputexports".

How make works is that if it sees "include foo.mk", and foo.mk doesn't exist, then it warns, but continues parsing a makefile. Then if it finds rule how to make foo.mk, it runs it and retries the include. That's the idea behind it, and we with @daniel-thompson tested it to work in other contexts. There may be some integrational issue of course, so for now staying with a revision before that change makes sense, and we'll be happy to look into the issue.

grgustaf:
@poussa, we also spent a bunch of time looking at this today. It's not the outdir I think that causes the problem, it's a .config getting created in the zephyr tree. @ibriano helped us look at this today, too. He thought @daniel-thompson 's Samsung/jerryscript#1407 patch today that removed lots of arch-specific stuff was a really good step, but there are still some problems.

grgustaf:
@pfalcon, okay the (great?) thing about make is there's always an infinity of quirks left to not know about. :) Alright, I didn't see that Makefile.export ever got created, but I did see the make outputexports line run and couldn't quite understand why, so maybe that's what I was seeing. I'll try to look into it again soon. Thanks!

grgustaf:
+1, seems to be working let's get it in. I tested with the btgrove demo and travis passes.

pfalcon:
> patch today that removed lots of arch-specific stuff was a really good step, but there are still some problems.

Yep, that's a great step towards making Zephyr.js support *Zephyr*, not just few boards in Zephyr ;-). Sorry for the issues which surface, I guess doing upgrades step by step (and not altogether) makes sense after all, and as I wrote, we'll be happy to look into issues (not sure if I'm able to do that tomorrow, but definitely will be next week).

grgustaf:
I developed some scripts to automatically build against each commit of Zephyr and JerryScript that come out. I'll be trying to clean those up and check them in, in a form they can be used by others. Another spinoff that will be the ability to do a "travis lite" locally that should be a lot quicker, from already having the Zephyr SDK installed and such. Finally, I plan to have an IRC bot report status changes like "build broken by new Zephyr commit xyz" and such.

BTW @pfalcon join us on Freenode #zjs if you like. Then we'll have a reason to actually talk on there. :)

pfalcon:
> the (great?) thing about make is...

Btw, I'm *really* happy Z.js uses make and not something else. JrS with its CMake wrapped into a Python script is much more confusing. Yesterday, when I looked how to build JrS with "minimal" config, but with Date object enabled, it took a bit more time than I'd wish (my attitude towards that is that make, at least everyone knows, including its issues; but if there're so many alternative tools, there may be something wrong with each of them, and *unknown* issues ;-) ).

> join us on Freenode #zjs if you like

Thanks for letting me know, I will.

PR #305, opened 2016-10-19
[console.log] Added console.log to be used instead of print

jprestwo:

### DIFF
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -1,0 +1,66 @@
+// Copyright (c) 2016, Intel Corporation.
+#ifdef BUILD_MODULE_CONSOLE
+
+#include "zjs_common.h"
+#include "zjs_util.h"
+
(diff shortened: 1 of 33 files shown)


src/zjs_console.c
```diff
@@ -0,0 +1,66 @@
...
+                                 const jerry_value_t this,
+                                 const jerry_value_t argv[],
+                                 const jerry_length_t argc)
+{
+    if (!jerry_value_is_string(argv[0])) {
+        ERR_PRINT("first parameter must be a string\n");
+        return ZJS_UNDEFINED;
+    }
+    jerry_size_t jlen = jerry_get_string_size(argv[0]);
+    char buffer[jlen + 1];
```
poussa: We need to limit how much stack we use for this in order to avoid stack overflow and crash.

grgustaf:
LGTM, I should probably try compiling though. :)

PR #308, opened 2016-10-20
[Tests] Update print to console.log instead

haoxli:
Due to PR #305


jimmy-huang:
+1

PR #317, opened 2016-10-21
[zjs_k64f_pins] SW3 button is confirmed to be on PTA4.

pfalcon:
This is also what https://developer.mbed.org/platforms/FRDM-K64F/ lists.

Signed-off-by: Paul Sokolovsky paul.sokolovsky@linaro.org

### DIFF
--- a/src/zjs_k64f_pins.c
+++ b/src/zjs_k64f_pins.c
@@ -71,9 +71,9 @@ jerry_value_t zjs_k64f_init()
     zjs_obj_add_number(obj, PTE + 26, "LEDG");  // verified
     zjs_obj_add_number(obj, PTB + 21, "LEDB");  // verified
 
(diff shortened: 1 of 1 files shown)


pfalcon:
SW3 doesn't work with currently used Zephyr 1.5.0 (causes application/board lockup on press), but works well with Z upgraded to master with https://github.com/01org/zephyr.js/pull/304 . I confirmed it to be on PTA4 on my board, and that's what listed at https://developer.mbed.org/platforms/FRDM-K64F/ . Given that user buttons are pretty important user interaction, I reject an idea that different board revisions may have SW3 on different GPIOs. Though, if there're suspicions like that, let me know, and I'll investigate further.

pfalcon:
Ping.

poussa:
+1

PR #318, opened 2016-10-21
[console.log] Fixed all JS files to use console.log

jimmy-huang:


grgustaf:
+1

PR #321, opened 2016-10-24
[travis] Add TrafficLight, I2C, and WebBT samples as standard builds

grgustaf:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
 
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install make gcc g++ python3-ply ncurses-dev -qq
+  - sudo apt-get install make gcc g++ python3-ply ncurses-dev sysvbanner -qq
 
 install: >
   if [ "$TARGET" != "linux" ]; then
@@ -35,8 +35,11 @@ before_script: >
 
 script: >
   if [ "$TARGET" == "zephyr" ]; then
+    banner "Test 1"
     make V=1
+    banner "Test 2"
     make V=1 "JS=samples/TrafficLight.js"
+    banner "Test 3"
     make V=1 "JS=samples/I2C.js"
   elif [ "$TARGET" == "zephyr-large" ]; then
     make V=1 "JS=samples/WebBluetoothGroveLcdDemo.js"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -30,10 +31,15 @@ install: >
...
+  source zjs-env.sh $PARTSIZE && make update && source deps/zephyr/zephyr-env.sh
 
 script: >
-  if [ "$TARGET" == "default" ]; then
-    make V=1; else
-    make V=1 "${TARGET}";
+  if [ "$TARGET" == "zephyr" ]; then
+    make V=1
+    make V=1 "JS=samples/TrafficLight.js"
+    make V=1 "JS=samples/I2C.js"
```
baleboy: Testing each JS sample could be done in parallel on a separate VM, but is that overkill?

baleboy:
+1 with some optional comments

grgustaf:
Actually, I started out with them all in parallel (still visible in first commits) but then saw that we are limited to like 4-5 VMs at a time unless we pay for an account. That would have doubled the time to complete the tests. Most of the overhead is in getting the VM ready so I think it makes sense to run multiple tests in one VM.

As for K64F, that will make sense at some point when more tests are compatible, but for right now I have special K64F versions of some samples. So yes at some point.

grgustaf:
@baleboy, oh hmm should I put && in between my make commands so that any of them will result in the test failing? (Bash isn't my native tongue.)

baleboy:
@grgustaf actually now that you ask I'm surprised it even works without && - clearly I'm no bash expert myself. Are you sure it's not running only the first command? I scanned through the log but couldn't figure it out.

baleboy:
To clarify, afaik multi-line if blocks are fine in bash, but I thought Travis' "Yaml-to-bash" conversion was handling everything as a one-line command. But if it works, I think a failure in any of the statements will cause the whole build to fail.

baleboy:
I can see the line "Creating C string from JS application" three times in the build log, so it looks everything went fine.

grgustaf:
OK right, and I remember the build took longer so I know it's running all three. But maybe the best way to know for sure is to purposely break one of the early builds.

jimmy-huang:
+1

PR #328, opened 2016-10-25
[modules] Changing modules to use singletons

brianjjones:
Each module will now only be created once.  Additional calls to
requires("module") will return the previously created instance.
This will cut down on memory consumption and prevent unintended
behavior.

### DIFF
--- a/src/zjs_modules.c
+++ b/src/zjs_modules.c
@@ -104,6 +104,7 @@ static jerry_value_t native_require_handler(const jerry_value_t function_obj,
         }
     }
 
+    PRINT("MODULE: `%s'\n", module);
     return zjs_error("native_require_handler: module not found");
 }
 
--- a/src/zjs_modules.c
+++ b/src/zjs_modules.c
@@ -31,6 +31,7 @@
 typedef struct module {
     const char *name;
     initcb_t init;
+    jerry_value_t instance;
 } module_t;
 
 module_t zjs_modules_array[] = {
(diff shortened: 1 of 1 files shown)


src/zjs_modules.c
```diff
@@ -95,16 +96,26 @@ static jerry_value_t native_require_handler(const jerry_value_t function_obj,
...
-            return jerry_acquire_value(mod->init());
+            // We only want one intance of each module at a time
+            if (mod->instance == 0) {
+                mod->instance = jerry_acquire_value(mod->init());
+            }
+            return mod->instance;
         }
     }
 
-    PRINT("MODULE: `%s'\n", module);
```
grgustaf: This should stay there. It's meant to tell you which module failed to load in the unlikely event of such a failure. (But a typo in require() would do this maybe?)

It used to be part of the print but I pulled it out so zjs_error wouldn't need to take variable arguments.

src/zjs_modules.c
```diff
@@ -95,16 +96,26 @@ static jerry_value_t native_require_handler(const jerry_value_t function_obj,
...
 void zjs_modules_init()
 {
+    int modcount = sizeof(zjs_modules_array) / sizeof(module_t);
+
+    for (int i = 0; i < modcount; i++) {
+        module_t *mod = &zjs_modules_array[i];
+        mod->instance = 0;
+    }
+
     jerry_value_t global_obj = jerry_get_global_object();
```
grgustaf: This is actually a place where we forget to free the global object. Not sure if you want to handle that in this PR or elsewhere.

src/zjs_modules.c
```diff
@@ -95,16 +96,26 @@ static jerry_value_t native_require_handler(const jerry_value_t function_obj,
...
     return zjs_error("native_require_handler: module not found");
 }
 
 void zjs_modules_init()
 {
+    int modcount = sizeof(zjs_modules_array) / sizeof(module_t);
+
+    for (int i = 0; i < modcount; i++) {
+        module_t *mod = &zjs_modules_array[i];
+        mod->instance = 0;
```
grgustaf: I'm 99% sure that since the modules array was created as static global data it is already initialized to zero. On one hand it doesn't hurt anything, but on the other hand we're trying to pinch bytes here! Can you verify this?

src/zjs_modules.c
brianjjones: Whoops that was unintentional - probably due to rebasing.

src/zjs_modules.c
brianjjones: It absolutely is zero from the start, no question there.  But this is for when we launch a ZJS app multiple times in ashell, I need to reset the values to zero without nuking the entire struct. I'm open to suggestions for something more efficient :)

src/zjs_modules.c
brianjjones: Good catch, I'm taking care of it in this PR https://github.com/01org/zephyr.js/pull/336

poussa:
LGTM

grgustaf:
+1

PR #329, opened 2016-10-26
[docs] Updating timers.md doc file

xiaolongx:
Fix a blank page in timers.md file.

### DIFF
--- a/docs/timers.md
+++ b/docs/timers.md
@@ -82,4 +82,4 @@ called.
 Sample Apps
 -----------
 * [Timers sample](../samples/Timers.js)
-* [Spaceship2 sample](../arduino/starterkit/Spaceship2.js)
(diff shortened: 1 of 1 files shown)


docs/timers.md
```diff
@@ -82,4 +82,4 @@ called.
 Sample Apps
 -----------
 * [Timers sample](../samples/Timers.js)
-* [Spaceship2 sample](../arduino/starterkit/Spaceship2.js)
+* [Spaceship2 sample](../samples/Spaceship2.js)
```
grgustaf: Good catch, the old path was wrong. But the new path is also wrong. :)

It should be:
`../samples/arduino/starterkit/Spaceship2.js`

grgustaf:
+1, thanks for the fix. I'll just merge this onto devel and then cherry pick it manually onto master to avoid needless pain.

PR #336, opened 2016-10-27
[memory] Free global javascript objects and allow multiple restart

brianjjones:
This adds the necessary code to supply JerryScript with callbacks
to free memory when an object is garbage collected.  It also makes
the changes needed to ensure global JavaScript objects get GC'd.
Finally - it adds the code needed by ashell to be able to run
a ZJS JavaScript file multiple times without rebooting.

Signed-off-by: Brian J Jones brian.j.jones@intel.com

### DIFF
--- a/src/zjs_gpio.c
+++ b/src/zjs_gpio.c
@@ -323,6 +323,8 @@ static jerry_value_t zjs_gpio_open(const jerry_value_t function_obj,
 
         handle->pin = newpin;
         handle->pin_obj = async ? jerry_acquire_value(pinobj) : pinobj;
+        handle->devnum = devnum;
+
         // Register a C callback (will be called after the ISR is called)
         handle->callbackId = zjs_add_c_callback(handle, gpio_c_callback);
         // Set the native handle so we can free it when close() is called
--- a/src/zjs_gpio.c
+++ b/src/zjs_gpio.c
@@ -39,6 +39,7 @@ void (*zjs_gpio_convert_pin)(uint32_t orig, int *dev, int *pin) =
 struct gpio_handle {
     struct gpio_callback callback;  // Callback structure for zephyr
     uint32_t pin;                   // Pin associated with this handle
+    uint32_t devnum;
     uint32_t value;                 // Value of the pin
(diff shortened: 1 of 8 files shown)


src/zjs_ipm.c
```diff
@@ -98,4 +98,15 @@ void zjs_ipm_register_callback(uint32_t msg_id, ipm_callback_t cb)
 #endif
 }
 
+void zjs_ipm_free_callbacks() {
+    #ifdef CONFIG_X86
```
poussa: CONFIG_BOARD_ARDUINO_101 ?

poussa:
LGTM minus one comment. Did not test (yet).

src/zjs_ipm.c
brianjjones: I was simply following what the rest of the file uses.  @jimmy-huang should we be using a different flag than CONFIG_X86 for IPM now?  Or does it still make sense?

jimmy-huang: so CONFIG_X86 means it's only for the x86 side and not ARC, so I think it's correct here, the whole IPM is conditionally inside CONFIG_BOARD_ARDUINO_101, but CONFIG_X86 and CONFIG_ARC further separate between them, since ARC doesn't have any mem alloc capababilities, it makes sense to fall into X86 only.

src/zjs_gpio.c
```diff
@@ -185,6 +186,18 @@ static jerry_value_t zjs_gpio_pin_close(const jerry_value_t function_obj,
...
 
+static void zjs_gpio_free_cb(const uintptr_t native)
+{
+    struct gpio_handle* handle = (struct gpio_handle*)native;
+    zjs_remove_callback(handle->callbackId);
+    if (handle->onchange_func) {
+        jerry_release_value(handle->onchange_func);
+    }
+
+    gpio_remove_callback(zjs_gpio_dev[handle->devnum], &handle->callback);
```
brianjjones: I added the line to set handle->devnum when the handle is created.

jimmy-huang:
+1

PR #338, opened 2016-10-28
[samples] Fixed Spaceship2.js timer/GPIO bug

jprestwo:
- On Zephyr, if a GPIO pin is configured to have double edge detection
  it will still sometimes signal an interrupt with the same value as
  the last time. This causes this sample in particular to set the timer
  interval two times without clearing it. Adding these if blocks, provides
  some sort of syncronization so that the timer can only be added once
  before being cleared. This way of doing this should be in any sample
  that is similar to this one.

Signed-off-by: James Prestwood james.prestwood@intel.com

### DIFF
--- a/samples/arduino/starterkit/Spaceship2.js
+++ b/samples/arduino/starterkit/Spaceship2.js
@@ -41,31 +41,36 @@ var btnstate = false;
 // turn green LED on initially
 led1.write(true);
 
-var timer;
+var timer = null;
 var toggle;
 
 button.onchange = function (event) {
     if (event.value) {
-        // button has been pressed
-        toggle = false;
-        led1.write(false);
-        led2.write(false);
-        led3.write(true);
+        if (timer == null) {
+            // button has been pressed
+            toggle = false;
+            led1.write(false);
+            led2.write(false);
+            led3.write(true);
 
-        // start timer to toggle LED states every 250ms (1/4 second)
-        timer = setInterval(function () {
-            toggle = !toggle;
-            led2.write(toggle);
-            led3.write(!toggle);
-        }, 250);
+            // start timer to toggle LED states every 250ms (1/4 second)
+            timer = setInterval(function () {
+                toggle = !toggle;
+                led2.write(toggle);
+                led3.write(!toggle);
+            }, 250);
+        }
     }
     else {
-        // button has been released
-        led1.write(true);
-        led2.write(false);
-        led3.write(false);
+        if (timer != null) {
+            // button has been released
+            led1.write(true);
+            led2.write(false);
+            led3.write(false);
 
-        // remove the timer event
-        clearInterval(timer);
+            // remove the timer event
+            clearInterval(timer);
+            timer = null;
+        }
     }
 }


jprestwo:
@grgustaf,

We have talked about this before, but this seems to be an inherent problem with Zephyr's GPIO interrupts when both rising and falling edge detection is enabled. Occasionally, you will receive an interrupt that has the same pin value as the previous one, which should never happen when double edge detection is enabled. Specifically for the Spaceship2.js sample, this will cause the setInterval to be called twice before it is cleared. To fix this, I have added an if block which checks that the timer handle is/is not null so that it will not call setInterval again before clearing the timer.

grgustaf:
Basically to rewrite your patch in English, "We know we're going to see the wrong value sometimes, and when we do we'll ignore the event entirely."

This doesn't seem too good to me because e.g. it will leave the Spaceship's lasers firing indefinitely. So I want to look deeper into what's the right way to use the underlying APIs so we give a good example of how to do it for our users. Which also probably means getting to the bottom of why we see what we do from Zephyr and whether that's to be expected.

jprestwo:
@grgustaf, the last time I tested using JUST Zephyr's own GPIO sample, I was getting duplicate values, so I don't think its how we are using the API, I think its got to do with how Zephyr sets up the pin (debounce or something).

grgustaf:
@jprestwo, the Zephyr callback does not send us the pin value. We have to call read to get the pin value in that callback handler. So that's going back to the pin and asking for its value. It's understandable (to me, I think) that it might give the "wrong" value, because of bounce issues like you mention. I.e. presumably it did go low causing the falling edge that triggered the interrupt, but by the time I go back to read it, it has bounced back up to high.

It seems to me like the JS app writer will just have to deal with that reality.

The part that troubles me more is that we're not reliably getting the rising/falling callbacks in pairs. I would hope that Zephyr could guarantee that we will get one and only one callback each time the value rises and falls. So in other words, if it started out as low, and I get 16 "bidirectional" callbacks, it should be low again.

If we can't rely on that either, then I need to dig into this more to understand what is a reasonable way to program with GPIOs. It seems we cannot rely on that currently because when I press and release my pushbutton enough times quickly, I'm able to see an odd number of callbacks have been called sometimes. So I need to figure out if that's a bug in Zephyr or another problem the app writer is expected to deal with.

grgustaf:
Talked with Zephyr folks today and I should start by giving them a test case to reproduce the issue.

grgustaf:
I don't know how I closed this; I didn't mean to.

grgustaf:
Ah this got closed because I deleted devel branch.

grgustaf:
I've learned that Zephyr cannot guarantee that I will get balanced rising/falling edge callbacks. So I think a different tactic is needed to be sure we will turn off the lasers again once we turn them on. I've propose a new PR #484 to replace this one.

Like this patch, it is more careful to not duplicate timers. But it also stops using the event.value because that is unreliable because of button bounce. We should perhaps remove that event.value entirely so people don't rely on it.

PR #340, opened 2016-10-29
[callbacks] Return this from some methods to support chaining

gabrielschulhof:
This unifies the interface with that of node.js. The methods that support
chaining are:
- add_listener
- remove_listener
- remove_all_listener
- set_max_listeners

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -127,6 +127,7 @@ bool zjs_remove_callback_list_func(int32_t id, jerry_value_t js_func)
             }
         }
     }
+    DBG_PRINT("could not remove callback %d\n", id);
     return false;
 }
 
--- a/src/zjs_event.c
+++ b/src/zjs_event.c
@@ -92,28 +92,28 @@ static jerry_value_t add_listener(const jerry_value_t function_obj,
 {
     if (!jerry_value_is_string(argv[0])) {
         DBG_PRINT("first parameter must be event string\n");
-        return ZJS_UNDEFINED;
+        return this;
     }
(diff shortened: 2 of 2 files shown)


brianjjones:
@gabrielschulhof I'm afraid this PR fails when running samples/tests/Events.js.  I see only this in the console:

ipm_console0: 'Sensor core running ZJS ARC support image'
exit

Can you do a bit of digging to see why it fails?  If I use a clean copy of master it works.

grgustaf:
@jprestwo could you take a look at this?

jprestwo:
I have had problems of returning 'this' from a function. I never got down to what was wrong but the program would just sit there and spin. It could be a JerryScript problem. But I can take a look again and try and get down to why its not working. It could be because we have to acquire 'this' and return the acquired value? I don't believe I tried that when I was seeing this problem before.

jimmy-huang:
@jprestwo I don't think you'll need to acquire the 'this' object before, so this patch looks okay, if you can test it out if it's not breaking anything, we can merge this.

brianjjones:
@jimmy-huang I tested it earlier and it fails on samples/tests/Events.js  Unless Events.js needs to be changed to work with this new patch, I don't think it works as intended and breaks functionality.

grgustaf:
This is still targeting "devel" - it needs a rebase to master, or else we can manually merge it when it's ready if there's any confusion about how to do that.

brianjjones:
I've rebased this and created a new pull request here https://github.com/01org/zephyr.js/pull/431

jimmy-huang:
I am going to close this as #431 is already merged.

Issue #343, opened 2016-10-31
console.log() behavior is not compliant


pfalcon:
To reproduce, modify samples/HelloWorld.js (do we get an interactive mode?) to contain:

~~~
console.log(1)
console.log('foo', 'Hello, ZJS world!')
~~~

Run:
~~~
$ ./jslinux samples/HelloWorld.js 

foo
~~~

Issue: numeric argument is not printed. With multiple arguments, only 1st is silently printed.

For comparison:

~~~
$ nodejs samples/HelloWorld.js 
1
foo Hello, ZJS world!
~~~

Browser interactive console:
~~~
console.log(1)
console.log('foo', 'Hello, ZJS world!')
1
foo Hello, ZJS world!
undefined
~~~

kenchris:
https://console.spec.whatwg.org/

pfalcon:
I also tried the HelloWorld.js as modified per above on frdm_k64f, with the same result:

~~~
$ picocom -b115200 /dev/ttyACM0 
...
Terminal ready

foo
~~~

In other words, currently Z.js doesn't have means to print numbers (which is hard to believe, so maybe regression after recent JrS/Z upgrade?).

Note that upstream JrS with its print() feels well in this regard (Z.js's print() doesn't print numbers still):

~~~
$ picocom -b115200 /dev/ttyACM0 
...
Terminal ready

JerryScript build: Oct 31 2016 17:31:56
JerryScript API 1.0
Zephyr version 1.5.99
js> print(1)
1
undefined
js> print("foo", "bar")
foo bar
undefined
js> 
~~~

jprestwo:
I wrote console.log to behave the exact same as print. I did not foresee the event of providing JUST a number like `console.log(1)`, which is definitely an acceptable use case. Good catch, ill get this fixed.

pfalcon:
Thanks. I hope you have an idea how to fix it well, because I so far don't have a clear picture. But I'd suggest to consider not trying to have own implementation, but reuse JrS' print() implementation - it already handles multiple args, and Z.js carries its code anyway. Indeed, a workaround I use so far is:

~~~
-    zjs_obj_add_function(global_obj, native_print_handler, "print");
+//    zjs_obj_add_function(global_obj, native_print_handler, "print");
~~~

This stops overriding JrS's own print().

How to implement console.log() in terms of JrS's print() is another question. I checked that their implementation is static, and calling them to un-static it for reuse wouldn't sound completely ungrounded for me.

[...]

Issue #341, opened 2016-10-31
How to run tests?


pfalcon:
Please let me play "casual user" here (I'm going to dig a bit deeper after this report, but I'm afraid one can get only so far with this matter without docs). So, I see `tests/` dir in the tree, I see commit traffic around tests. However, there's no info of them in the README beyond just mentioning `tests/` dir. Ok, I try to run them. "make test", "make check", "make tests" don't work. I decide to run them manually, and assume that the most obvious way to do that is with Linux version, as something which everyone has, though seeing test-aio.js puts me off a bit. Anyway:

~~~
$ ./jslinux tests/test-buffer.js 

JerryScript: cannot run javascript
~~~

No error, no nothing, just cannot run.

~~~
$ ./jslinux tests/test-timers.js 

PASS - clearInterval: intervalID does not exist
PASS - clearTimeout: timeoutID does not exist
PASS - setTimeout: set delay
PASS - setTimeout: set delay and optional arg
PASS - setInterval: set delay
PASS - clearInterval: intervalID
Segmentation fault (core dumped)
~~~

Note the segmentation fault at the end.

pfalcon:
I tried to build a test image for frdm_k64f with:

~~~
make BOARD=frdm_k64f JS=tests/test-timers.js
~~~

Result:

~~~
PASS - clearInterval: intervalID does not exist
native_set_interval_handler: timer alloc failed
JerryScript: cannot run javascript
~~~

haoxli:
They can run as well as `samples/`:
```
make JS=tests/test-timers.js
```
Unfortunately `test-timers.js` is blocked due to #310, please try other tests, thanks.

baleboy:
I think @pfalcon 's comment is valid, a make 'test' target that translates to the test command above would be a simple way to make things easier.

grgustaf:
I've assigned to @jprestwo to see what we can do here. You're right, the linux build has only very partial support but we have failed to document what it is useful for and what the plans are.

We should think of some way to identify what tests are expected to work with jslinux. I'm surprised though that nothing I've tried in tests/ seems to work expect for test-eval.js.

grgustaf:
So some possible actions:
- Add some documentation to main README and a tests/README.
- Fix jslinux or tests so that more work with it.
- Perhaps create a linuxtest make target that would run everything expected to work, or something. This also seems it would depend on #330. (In turn, once we had that we could add these linux tests to Travis.)

jprestwo:
[...]

Issue #325, opened 2016-10-25
[RFC] Adding basic time function support


pfalcon:
I see that currently Zephyr.js doesn't seem to have any means to measure time intervals. I'd like to add basic support for this, for benchmarking, and well, to measure time intervals. At this time, no talking about datetime (real time, wall clock time).

Quick survey on how to do it in JavaScript brought:
1. Standard down to ECMAScript Date object. The issue here is that this is pretty overloaded object, used for datetime functionality too. `Date.now()` (static method) is speced as "returns the number of milliseconds elapsed since 1 January 1970 00:00:00 UTC."
2. Node.js has `process.hrtime()` returning Array of [time_since_epoch_in_seconds, nanoseconds]
3. Draft "High Resolution Time" spec from W3C: https://www.w3.org/TR/hr-time/ , https://developer.mozilla.org/en-US/docs/Web/API/Performance/now . This would appear as the most relevant to the topic of benchmarking and what we can achieve out of the box with Zephyr - it doesn't depend on some absolute epoch, but on a relative time.

My personal preference would be go Date.now() way, as it's more widely known. It may be also first step of supporting other Date features (absolute time). Note that Date is implemented (as in: available) in JerryScript, so implementing Zephyr's version there would allow reuse by other JrS-based project. There's a risk too, if JrS maintainers won't agree that for the lack of better alternative, it's acceptable to use a relative time (since boot of a board), instead of absolute Unix Epoch time. Then, Date implementation would need to be migrated to Zephyr.js.

pfalcon:
> To the extent we've thought about it at all, the vague plan would have been to try to pull in JrS version at some point, but maybe working on some config options to pull in just what we really need or the actual app uses and not everything. (We have a similar issue for Math etc.)

I see. Well, I'd say that may be spreading too thin. It's one thing to include/omit whole modules, and rather different matter to exclude/omit individual objects or their methods. Non-trivial program analysis may be required for the latter vs just grepping for "require(.+)". Anyway, indeed, to optimize something, one first need to have it implemented ;-).

> (Apologies that we've got a bit of a patch logjam this week as we work on CI and such; we appreciate your contributions and will try to get them merged ASAP.)

No problem, I see that there's a lot of important activity happens.

> I don't think I understand the relative time issue you refer to at the end there.

[...]

PR #344, opened 2016-10-31
[console] Fixed console.log/error to conform to Node.js better

jprestwo:

### DIFF
--- a/samples/Console.js
+++ b/samples/Console.js
@@ -1,0 +1,12 @@
+console.log("TESTING LOG");
+console.log(1);
+console.log(-1);
+console.log(123.456);
+console.log(-123.456);
+console.log({int_val: 1});
+
+console.error("TESTING ERROR");
+console.error(1);
+console.error(123.456);
+console.error(-123.456);
+console.error({int_val: 1});
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -62,7 +62,7 @@ static jerry_value_t do_print(const jerry_value_t function_obj,
                 fprintf(out, "%lu ", num);
             } else if (type == IS_INT) {
                 int32_t num = (int32_t)jerry_get_number_value(argv[i]);
-                fprintf(out, "%ld ", num);
+                fprintf(out, "%li ", num);
             }
         } else if (jerry_value_is_object(argv[i])) {
             fprintf(out, "[Object] ");
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -56,7 +56,13 @@ static jerry_value_t do_print(const jerry_value_t function_obj,
             int type = is_int(argv[i]);
             if (type == IS_NUMBER) {
                 double num = jerry_get_number_value(argv[i]);
-                fprintf(out, "%lf ", num);
+#ifdef ZJS_LINUX_BUILD
+                fprintf(out, "%f ", num);
+#else
+                // TODO: printing floats does not currently work on Zephyr
+                int32_t num32 = (int32_t)num;
+                fprintf(out, "[Float ~%ld] ", num32);
+#endif
             } else if (type == IS_UINT) {
                 uint32_t num = (uint32_t)jerry_get_number_value(argv[i]);
                 fprintf(out, "%lu ", num);
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -51,7 +51,6 @@ static jerry_value_t do_print(const jerry_value_t function_obj,
             char buffer[jlen + 1];
             int wlen = jerry_string_to_char_buffer(argv[0], (jerry_char_t *)buffer, jlen);
             buffer[wlen] = '\0';
-
             fprintf(out, "%s ", buffer);
         } else if (jerry_value_is_number(argv[i])) {
             int type = is_int(argv[i]);
@@ -69,6 +68,7 @@ static jerry_value_t do_print(const jerry_value_t function_obj,
             fprintf(out, "[Object] ");
         }
     }
+    fprintf(out, "\n");
     return ZJS_UNDEFINED;
 }
 
@@ -77,22 +77,7 @@ static jerry_value_t console_log(const jerry_value_t function_obj,
                                  const jerry_value_t argv[],
                                  const jerry_length_t argc)
 {
-    if (!jerry_value_is_string(argv[0])) {
-        ERR_PRINT("first parameter must be a string\n");
-        return ZJS_UNDEFINED;
-    }
-    jerry_size_t jlen = jerry_get_string_size(argv[0]);
-    if (jlen > MAX_STR_LENGTH) {
-        ERR_PRINT("console string is too long, max length=%u\n", MAX_STR_LENGTH);
-        return ZJS_UNDEFINED;
-    }
-    char buffer[jlen + 1];
-    int wlen = jerry_string_to_char_buffer(argv[0], (jerry_char_t *)buffer, jlen);
-    buffer[wlen] = '\0';
-
-    STDOUT_PRINT("%s\n", buffer);
-
-    return ZJS_UNDEFINED;
+    return do_print(function_obj, this, argv, argc, stdout);
 }
 
 static jerry_value_t console_error(const jerry_value_t function_obj,
@@ -100,22 +85,7 @@ static jerry_value_t console_error(const jerry_value_t function_obj,
                                    const jerry_value_t argv[],
                                    const jerry_length_t argc)
 {
-    if (!jerry_value_is_string(argv[0])) {
-        ERR_PRINT("first parameter must be a string\n");
-        return ZJS_UNDEFINED;
-    }
-    jerry_size_t jlen = jerry_get_string_size(argv[0]);
-    if (jlen > MAX_STR_LENGTH) {
-        ERR_PRINT("console string is too long, max length=%u\n", MAX_STR_LENGTH);
-        return ZJS_UNDEFINED;
-    }
-    char buffer[jlen + 1];
-    int wlen = jerry_string_to_char_buffer(argv[0], (jerry_char_t *)buffer, jlen);
-    buffer[wlen] = '\0';
-
-    STDERR_PRINT("%s\n", buffer);
-
-    return ZJS_UNDEFINED;
+    return do_print(function_obj, this, argv, argc, stderr);
 }
 
 void zjs_console_init(void)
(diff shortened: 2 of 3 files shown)


jprestwo:
@pfalcon, this is a fix for https://github.com/01org/zephyr.js/issues/343

It ended up being relatively simple (or so I thought), so hopefully I got it somewhere near compliant to Node.js

pfalcon:
Thanks for coming forward with the fix!

> added console test

If it's intended to be a "test", maybe it's worth to put it into tests/ , to clarify https://github.com/01org/zephyr.js/issues/341 ?

pfalcon:
Running Console.js with linux port:

~~~
TESTING LOG
1
4294967295
123.456000
-123.456000
[Object]
TESTING ERROR
1
123.456000
-123.456000
[Object]
~~~

An issue is visible - integers are printed as unsigned.

pfalcon:
Now let's compare that with nodejs output:

~~~
TESTING LOG
1
-1
123.456
-123.456
{ int_val: 1 }
TESTING ERROR
1
123.456
-123.456
{ int_val: 1 }
~~~

It looks like nodejs became smart and actually prints dictionaries' contents, instead of just "[object Object]". Running in a browser console also will get a fancy output nowadays. But I'd really like to call to use a classical object print representation from JavaScript. With all the fancy stuff mentioned above, nowadays you can get this classy representation with alert() in browser. Let's try it:

~~~
alert({int_val: 1})
[object Object]
alert(performance)
[object Performance]
~~~

So, it prints "[object", followed by "native type", which is "Object" for {} things, and may be something different for other "native" (C-coded) types. Well, I'm not sure if we can do the latter part in Z.js, i.e. not sure if we can get to "[object Gpio]" or "[object gpio]", but following that pattern makes sense IMHO.

Oh, and it should call toString() on object if it's defined, do you do that? ;-) (That can be left for later (first bugreport) of course, but I wonder if JrS' native print() does that).

kenchris:
What about following the standard https://console.spec.whatwg.org/ and potentially contribute tests back to it? If Node.js impl changes, we can file issues for the standard as well. @poussa

kenchris:
https://console.spec.whatwg.org/#formatter is probably what you want to look at

kenchris:
Well,

1. Spec is quite small and is not designed by a committee, but instead trying to standardize what browsers and node.js already do today. 
 - Lacking a feature can be annoying but could be acceptable for us if it is really a size issue.
 - Working differently to what people are used to is just bad (for that JerryScript for instance supports full ES5 and not half baked stuff)
2. Being consistent is good - plus we are heavily focusing on standards (mostly in the W3C because of business reasons).
3. We can very well influence specs and even some features can be considered Level 2 or non-nominative. Upstreaming tests is always good.
4. I wouldn't be surprised if console.log makes it into the ES standard itself at some point.

poussa:
We need to have "good enough" console print feature. The "good enough" evolves over time. Right now, we have tons of other important things to work on so the bar is pretty low. First, we did not have console.log at all. Now we have it with single argument. The next step is to have multiple args. After that, we should decide if we want to have a standard (WHATWG) or de-facto (Node.js) API or subset of any. We should not over-engineer a print function.

kenchris:
@poussa I tend to agree. But looking at the spec, does make sense - it is actually small and pretty straight ahead. Then after that, deciding to leave things out for how seems totally fine.  Better than just doing things in the dark (I am not saying that we are doing that :-))

pfalcon:
> for that JerryScript for instance supports full ES5 and not half baked stuff

Not argument's sake, but because I'm genuinely interested in understanding Z.js philosophy to be able to contribute productively. I'm sure more people will be interested in it over time, so thinking about it now, and spelling out explicitly somewhere eventually, would help. (There're also other reasons why I'd prefer to raise these questions sooner rather then later, like in my organization, I'm (as someone who looked at those projects) being asked questions like "We initially planned to work on IoT.js, but now Zephyr.js appeared, which we should use, and why?")

So, indeed, JerryScript implements full ES5. And yet Zephyr.js uses it exactly in half-baked "minimal" configuration where a lot of standard ES5 things are missing. And that's good (supposedly) because enabling *just one* of missing features - Date class increases code size by almost 15% (https://github.com/01org/zephyr.js/issues/325#issuecomment-256434825). Or maybe it's bad, and we're just wasting time, and should enable "full" JrS config from the start and wave bye-bye to the smaller of MCUs out there (I'm still sure it'll fit into Arudino 101 or FRDM-K64F). But then there's IoT.js already, which is exactly a "more pumpy" JerryScript environment, and one of the promise of Z.js is being smaller than it.

Going further, there're more interesting things. Like, there was official ECMA "ECMAScript compact profile" (ECMA-327). JrS implemented it, but it was dropped before 1.0 release. Why? Because ECMA-327 was withdrawn long ago, even despite being nice 1-page standard (counting the informative part). And even ES5 is now too generations old.

So, Zephyr.js architects/maintainers would need to think about such things in a complex, and see where Zephyr.js' place in all this (and convey that to its users). No hurry of course, but having some guidelines is already helpful as this ticket shows (well, I hope there's an agreement that so far just adding features step by step is good enough). 

Thanks.

kenchris:
Yeah it is a tough tradeoff, but luckily with time even MCUs will have more memory. Or projects like ZJS will push for that. Memory is getting much cheaper and often doubling the memory on such devices shouldn't change the BOM that much, or at least I would happily pay for it :-) There is a huge gap between the memory available on devices such as the A101 and the Edison.

ES2015+ is about twice the size of ES5 and people are already getting used to its features such as promises, let, generators, async/await (ES2017), and currently that will just not fit on something like the Arduino 101.

When memory is not enough we will have to do some kind of tree-shaking (I believe the build scripts do something similar to that today), ie. avoid overhead of things not used by the scripts themselves. Getting that detailed enough can be a challenge though - and then you cannot just drop a new JS file to the device and expect it to work.

pfalcon:
@kenchris, I see, thanks. I guess for a stage we're at currently, just trying (simple) things by firing nodejs or bringing up a browser console is practical quick way. Of course, keeping in mind that there's actually a spec, and be ready to match an implementation against it going forward - is a very good idea.

pfalcon:
Another issue with proposed patch - several arguments to console.log() are being printed each on a new line, should be with a space between each other. Included Console.js doesn't test multi-argument case.

samples/Console.js
```diff
@@ -0,0 +1,12 @@
...
+console.log(-1);
+console.log(123.456);
+console.log(-123.456);
+console.log({int_val: 1});
+
+console.error("TESTING ERROR");
+console.error(1);
+console.error(123.456);
+console.error(-123.456);
+console.error({int_val: 1});
```
grgustaf: Yeah, I agree with @pfalcon this is more of a test than a sample. To pretend for a minute that we have a good reason for having both tests/ and sample/tests, the latter is the better place for a "developer test" like this. We don't need QA to be running it regularly, and it doesn't report PASS/FAIL like the ones in tests/ but it's still nice to share and keep around for comparison with future output perhaps.

grgustaf: Do still add a copyright header...

src/zjs_console.c
```diff
@@ -14,50 +14,78 @@
...
+            jerry_size_t jlen = jerry_get_string_size(argv[0]);
+            if (jlen > MAX_STR_LENGTH) {
+                ERR_PRINT("console string is too long, max length=%u\n", MAX_STR_LENGTH);
+                return ZJS_UNDEFINED;
+            }
+            char buffer[jlen + 1];
+            int wlen = jerry_string_to_char_buffer(argv[0], (jerry_char_t *)buffer, jlen);
+            buffer[wlen] = '\0';
+
+            fprintf(out, "%s\n", buffer);
```
grgustaf: I suggest you print the arguments separated by spaces instead of newlines; this is what I expect at least and what it seems to do in e.g. Chrome Inspector. (Then add a newline at the end.)

samples/Console.js
grgustaf: Also add some test lines showing the multi-argument feature.

grgustaf:
Wow, quite a discussion in here. :) I just want to affirm that our not-quite-written-down philosophy has indeed been trying to implement things iteratively as we can define a direct need, just like @pfalcon describes. In fact, I've intended us to be "demo-driven" i.e. expanding the real-world use cases supported and providing everything needed to get them working. But we're also glad to be informed about the world of standards / implementations out there to aim at as @kenchris has been reliably clueing us in on. My primary thought there is to follow whatever API is most likely that a developer coming to this project might already know, and thus find intuitive.

Addressing some more of @pfalcon's comments, we don't have that much headroom left on A101 (~50k flash after our WebBT demo?) and that was our initial target so we were forced to make the decisions that would make it possible. But you're right, the sweet spot for this project could turn out to be devices a bit bigger. Still, I'm glad we've come at it from this side because I think it's easier to start small and build up in a modular fashion than to pare monolithic things down later. We tried to do that up front.

samples/Console.js
```diff
@@ -0,0 +1,12 @@
+console.log("TESTING LOG");
+console.log(1);
+console.log(-1);
+console.log(123.456);
+console.log(-123.456);
+console.log({int_val: 1});
+
+console.error("TESTING ERROR");
+console.error(1);
+console.error(123.456);
```
grgustaf: You probably meant to print -1 again here with error, like above?

samples/Console.js
```diff
@@ -0,0 +1,12 @@
+console.log("TESTING LOG");
+console.log(1);
+console.log(-1);
+console.log(123.456);
+console.log(-123.456);
```
grgustaf: Neither of these floats are printing for me, it's blank.

grgustaf:
James looked into it and printf doesn't have float support. This issue came up on the [Zephyr users](https://lists.zephyrproject.org/archives/list/users@lists.zephyrproject.org/thread/FZGWGLNB7CXCMKMWNZ2AZSXJCP3EQRVL/) mailing list but didn't get an answer yet.

I suggested that for now @jprestwo could print [Float] (just like [Object]) for now, or maybe round it to an int and print that like [Float ~42], so we can get something in for now.

jprestwo:
@grgustaf, I changed how floats are printed per your recommendation:
 - On linux, it prints normally
 - On Zephyr it prints: 
    `[Float ~<number>]`

grgustaf:
Works for me on A101, +1.

PR #346, opened 2016-11-01
[performance] New module implementing performance counter API.

pfalcon:
Based on recent W3C API, formally defined in https://www.w3.org/TR/hr-time/
and informally in
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now . As the
last link suggests, real-world usage boils down to just:

    t = performance.now();

even though formal spec defines more properties for "performance" object.

As defined in the spec for a browser, "performance" is a property of
global window object and thus implicitly present into context of a JS
script. We however define it as a module, so usage pattern is:

    var performance = require("performance");
    t = performance.now();

This module is intended for benchmarking, and in extreme cases, for
synchronous delays, etc. Supported both in Linux and Zephyr builds.
A testcase is included.

### DIFF
--- a/tests/test-performance.js
+++ b/tests/test-performance.js
@@ -1,0 +1,26 @@
+var performance = require("performance");
+
+var total = 0;
+var passed = 0;
+
+function assert(actual, description) {
+    total += 1;
+    var label = "\033[1m\033[31mFAIL\033[0m";
+    if (actual === true) {
+        passed += 1;
+        label = "\033[1m\033[32mPASS\033[0m";
+    }
+    console.log(label + " - " + description);
+}
+
+var before = performance.now();
+//console.log("performance.now() before:", before);
+
+setTimeout(function() {
+    var after = performance.now();
+    //console.log("performance.now() after:", after);
+    var diff = after - before;
+    // Allow for some jitter, especially on Linux
+    assert(diff >= 989 && diff <= 1011, "performance.now() result over known delay");
+    console.log("TOTAL: " + passed + " of " + total + " passed");
+}, 1000);
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -14,6 +14,7 @@ CORE_SRC = 	src/main.c \
 			src/zjs_event.c \
 			src/zjs_linux_time.c \
 			src/zjs_modules.c \
+			src/zjs_performance.c \
 			src/zjs_script.c \
 			src/zjs_script_gen.c \
 			src/zjs_timers.c \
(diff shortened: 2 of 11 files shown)


pfalcon:
This implements https://github.com/01org/zephyr.js/issues/325

poussa:
LGTM

In the follow up PR a doc entry is needed. I assume it returns microseconds, right?

pfalcon:
>  I assume it returns microseconds, right?

No, in full accordance to spec it return milliseconds, but as floating-point number. I added following to the commit message:

    The returned value is a floating-point number of milliseconds since
    arbitrary point in time. This means that its absolute value is
    meaningless and only difference between 2 values make sense (and
    represent time duration between calls which produced these 2 values).
    As the returned value is a floating-point number, it may have higher
    resolution than a millisecond (using fractional part). The actual
    resolution however depends on a particular Zephyr board, and some
    boards provide resolution only of 10 milliseconds.

Note that there's high-resolution timer also available from Zephyr (sys_cycle_get_32()), but it's explicitly defined as 32-bit, so wrap-around with it can't be ignored at all, so I don't want to even approach it (for now).

> In the follow up PR a doc entry is needed.

Ok, will work on it.

pfalcon:
Just to clarify, I tested qemu_x86 and frdm_k64f and the both have 10ms system timer resolution.

pfalcon:
Docs commit pushed to this PR.

tests/test-performance.js
```diff
@@ -0,0 +1,26 @@
...
+
+var before = performance.now();
+//console.log("performance.now() before:", before);
+
+setTimeout(function() {
+    var after = performance.now();
+    //console.log("performance.now() after:", after);
+    var diff = after - before;
+    // Allow for some jitter, especially on Linux
+    assert(diff > 990 && diff < 1010, "performance.now() result over known delay");
```
grgustaf: A101 also has 10ms timer resolution, so maybe you meant to use >= and <= here? When I see before = 30 and after = 1040, I get a failure.

Other than that +1, I'll see what you say and merge it tomorrow.

pfalcon: @grgustaf: Do you mean that you actually see 10ms jitter on Arduino 101, or mention possibility of that? I'd say I don't have concerns about Zephyr - it's `*RT*OS` after all. On frdm_k64f, I saw 1000 diff all the time (of course, there's always a chance to get 1001 if ticks incremented literally between few CPU instructions). When I wrote the above I mostly kept Linux in mind, that's why there's whole 10ms of jitter factor - and I'm afraid, on an overloaded system, you can get more than that. Anyway, I'll update as you suggest, and even make it 989 and 1011 to follow floating-point best practices. But I'd be really interested if you saw 10ms extra on a101.

grgustaf: Yeah that really was what I saw pretty consistently, 30 before, 1040 after, on reboot. Now that's **if I enabled the console prints you'd commented out**. Without them it seems to generally pass every time. So maybe the print itself is interfering somehow. I can't say I understand things to that level yet.

pfalcon: Ah, that may play it's role. Because yes, UART output is pretty slow - at 115200, it's mere ~11KB/s,  or 10 characters take whole millisecond. They (chars) should be buffered, but if they don't or buffer's up, can affect millisecond-level stuff considerably. Anyway, hope the issue is fixed now.

grgustaf:
Needs to be rebased to master for merging.

pfalcon:
> Needs to be rebased to master for merging.

Done as https://github.com/01org/zephyr.js/pull/362

PR #348, opened 2016-11-02
[Tests] Add tests for PWM APIs

cuiyanx:

### DIFF
--- a/tests/test-pwm.js
+++ b/tests/test-pwm.js
@@ -1,0 +1,144 @@
+// Copyright (c) 2016, Intel Corporation.
+
+
+console.log("Wire IO3 to IO2");
+
+var total = 0;
+var passed = 0;
+
+function assert(actual, description) {
+    total += 1;
+
+    var label = "\033[1m\033[31mFAIL\033[0m";
+    if (actual === true) {
+        passed += 1;
+        label = "\033[1m\033[32mPASS\033[0m";
+    }
+
+    console.log(label + " - " + description);
+}
+
+function expectThrow(description, func) {
+    var threw = false;
+    try {
+        func();
+    }
+    catch (err) {
+        threw = true;
+    }
+    assert(threw, description);
+}
+
+var pwm = require("pwm");
+var gpio = require("gpio");
+var pins = require("arduino101_pins");
+var pinA, pinB, msTimer, cycleTimer;
+
+pinB = gpio.open({pin: pins.IO2, direction: 'in'});
+
+// PWMPins open
+pinA = pwm.open({channel: pins.IO3});
+assert(pinA != null && typeof pinA == "object",
+      "open: defined pin and default argument");
+
+expectThrow("open: undefined pin", function () {
+    pinA = pwm.open({channel: 1024});
+});
+
+// set Period and PulseWidth with ms
+// duty cycle: 30%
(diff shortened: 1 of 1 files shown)


tests/test-pwm.js
```diff
@@ -0,0 +1,144 @@
...
+    pinA.setPulseWidth(300);
+});
+
+pinA = pwm.open({channel: pins.IO3, period: 3, pulseWidth: 1});
+assert(pinA != null && typeof pinA == "object",
+       "open: with period and pulseWidth");
+
+pinA.setPeriod(1000);
+
+expectThrow("pwmpin: set pulseWidth greater than period", function () {
```
grgustaf: This does not throw, by design. The API is not very good but reflects the state of underlying Zephyr API (which has just recently changed so we'll be able to fix this properly.)

The problem is you need to set both period and pulse width, but after you've set one but before you've set the other, you may be in an invalid state. The first call can't throw in that case because it's going to be corrected with the next line.

tests/test-pwm.js
```diff
@@ -0,0 +1,144 @@
...
+});
+
+// set Period and PulseWidth with ms
+// duty cycle: 30%
+var msTrue = 0;
+var msFalse = 0;
+var msCount = 0;
+
+expectThrow("pwmpin: set pulseWidth without period", function () {
+    pinA.setPulseWidth(300);
```
grgustaf: This does not throw, again because there's not requirement that you set period first and then pulse width first. I guess there could be. So I'd remove this test. The way we're going to fix it is that I've asked Zephyr to put both values into the same call, as in set_pwm_cycles(period, pulse_width). That way you're never in an invalid, halfway defined state. We'll change our API to match and then we'll be able to throw on errors.

tests/test-pwm.js
```diff
@@ -0,0 +1,144 @@
...
+    }, 10);
+}, 1000);
+
+expectThrow("pwmpin: set periodCycles with invalid value", function () {
+    pinA.setPeriodCycles("Value");
+});
+
+expectThrow("pwmpin: set pulseWidthCycles with invalid value", function () {
+    pinA.setPulseWidthCycles("Value");
+});
```
grgustaf: Add a newline to the end of the last line. (Not a blank line.)

tests/test-pwm.js
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) 2016, Intel Corporation.
+
+
```
grgustaf: Remove extra blank line.

tests/test-pwm.js
cuiyanx: This is my testing purpose.

tests/test-pwm.js
cuiyanx: I must test this situation to get error message, and that does not throw out is not that I fixed it with the next line.

tests/test-pwm.js
grgustaf: It's working as designed, but I guess it's okay to leave this in there as a failure because it should eventually be fixed. But it will be fixed by a redesign of the API, so the test will no longer work and will have to be changed at that point.

grgustaf:
Resubmitted with whitespace fixes and rebased on master as PR #392. Closing.

PR #309, opened 2016-10-20
[Samples] Fix onchange event in ButtonEnable.js

xiaolongx:
Following gpio.md, there is no gpio.on method, but onchange event instead.

PR #350, opened 2016-11-02
[Samples] Fix onchange event in ButtonEnable.js

grgustaf:
Following gpio.md, there is no gpio.on method, but onchange event instead.


grgustaf:
+1. Rebased from PR #309.

PR #351, opened 2016-11-02
[timers] Changing how timers are released

brianjjones:
Because ZJS implements its own version of setInterval,
the new code to release global objects had a unintended
effect of causing timers that didn't save the return to
be garbage collected after the first call.  Now instead
of relying on JerryScript to GC our timers, we will clear
them ourself when jerry_cleanup runs.


src/zjs_timers.c
```diff
@@ -108,6 +108,22 @@ static bool delete_timer(int32_t id)
     return false;
 }
 
+void zjs_timers_free_timers()
+{
+    for (zjs_timer_t **ptm = &zjs_timers; *ptm; ptm = &(*ptm)->next) {
+        zjs_timer_t *tm = *ptm;
+        int i;
```
jimmy-huang: you could declare the i outside the for loop

src/ashell/jerry-code.c
```diff
@@ -133,9 +151,10 @@ void javascript_stop()
 
     /* Cleanup engine */
     jerry_cleanup();
+    zjs_ipm_free_callbacks();
+    zjs_timers_free_timers();
 
-    /* Initialize engine */
-    jerry_init(JERRY_INIT_EMPTY);
+    restore_zjs_api();
```
jimmy-huang: I understand the logic, but I am not sure restore_zjs_api should be called inside javascript_stop(), it seems it should be called after stop() and then before start() when you decided to run another app.

src/zjs_timers.c
grgustaf: I think @jimmy-huang meant inside the next for loop, where it's used?

src/zjs_timers.c
```diff
@@ -108,6 +108,22 @@ static bool delete_timer(int32_t id)
...
+        zjs_port_timer_stop(&tm->timer);
+        *ptm = tm->next;
```
grgustaf: I'm pretty sure this is buggy. E.g. the first time you're looking at zjs_timers (the first list object) and here you set *ptm (aka zjs_timers) to the second list object. After the loop iteration, the "next" statement runs which sets ptm to the pointer to the third list object, hence skipping the second.

The delete function worked because it returned after doing this line, it didn't continue in the loop.

You can simplify this whole thing by implementing the function as:
while (zjs_timers)
    delete_timer(zjs_timers->callback_id)

Slightly inefficient, but pretty obviously correct.

Or at least replace the outer for statement with a while (zjs_timers) to make sure you're not skipping any.

grgustaf:
You also need to rebase this to get rid of the three UART commits. You can do git rebase -i HEAD~5 for example and remove them.

brianjjones:
Not sure why rebasing is not working cleanly.  Closing this and starting a new PR sorry.

PR #353, opened 2016-11-02
[Sensor] Initial implementation of the Generic Sensor API

jimmy-huang:
Initial implementation for the W3C Generic Sensor API and add
support for the Accelerometer and Gyroscope sensors using
the BMI160 inertia sensor built-in on the Arduino 101,
the current implementation is polling from the ARC and channeling
the data back to X86.

Also adding two sample JS code testing both APIs.

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -619,6 +619,38 @@ static void fetch_sensor(struct device *dev) {
     }
 }
 
+/*
+ * The values in the following map are the expected values that the
+ * accelerometer needs to converge to if the device lies flat on the table. The
+ * device has to stay still for about 500ms = 250ms(accel) + 250ms(gyro).
+ */
+struct sensor_value acc_calib[] = {
+    {SENSOR_VALUE_TYPE_INT_PLUS_MICRO, { {0, 0} } },      /* X */
+    {SENSOR_VALUE_TYPE_INT_PLUS_MICRO, { {0, 0} } },      /* Y */
+    {SENSOR_VALUE_TYPE_INT_PLUS_MICRO, { {9, 806650} } }, /* Z */
+};
+
+static bool auto_calibration(struct device *dev)
+{
+    /* calibrate accelerometer */
+    if (sensor_attr_set(dev, SENSOR_CHAN_ACCEL_ANY,
+                        SENSOR_ATTR_CALIB_TARGET, acc_calib) < 0) {
+        return false;
+    }
+
+    /*
+     * Calibrate gyro. No calibration value needs to be passed to BMI160 as
+     * the target on all axis is set internally to 0. This is used just to
+     * trigger a gyro calibration.
+     */
+    if (sensor_attr_set(dev, SENSOR_CHAN_GYRO_ANY,
+                        SENSOR_ATTR_CALIB_TARGET, NULL) < 0) {
+        return false;
+    }
+
+    return true;
+}
+
 static void handle_sensor(struct zjs_ipm_message* msg)
 {
     uint32_t error_code = ERROR_IPM_NONE;
@@ -638,6 +670,10 @@ static void handle_sensor(struct zjs_ipm_message* msg)
                 error_code = ERROR_IPM_OPERATION_FAILED;
                 PRINT("failed to initialize BMI160 sensor\n");
             } else {
+                if (!auto_calibration(bmi160)) {
+                    PRINT("failed to perform auto calibration\n");
+                }
+
                 accel_poll = gyro_poll = false;
                 DBG_PRINT("BMI160 sensor initialized\n");
             }
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -366,7 +366,6 @@ static void handle_glcd(struct zjs_ipm_message* msg)
             glcd = device_get_binding(GROVE_LCD_NAME);
 
             if (!glcd) {
-                PRINT("failed to initialize Grove LCD\n");
                 error_code = ERROR_IPM_OPERATION_FAILED;
             } else {
                 DBG_PRINT("Grove LCD initialized\n");
@@ -477,7 +476,7 @@ static void send_sensor_data(enum sensor_channel channel,
 {
     struct zjs_ipm_message msg;
     msg.id = MSG_ID_SENSOR;
-    msg.type = TYPE_SENSOR_EVENT_VALUE_CHANGE;
+    msg.type = TYPE_SENSOR_EVENT_READING_CHANGE;
     msg.flags = 0;
     msg.user_data = NULL;
     msg.error_code = ERROR_IPM_NONE;
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -472,7 +472,7 @@ static inline int sensor_value_snprintf(char *buf, size_t len,
 #endif
 
 static void send_sensor_data(enum sensor_channel channel,
-                             double array[])
+                             union sensor_reading reading)
 {
     struct zjs_ipm_message msg;
     msg.id = MSG_ID_SENSOR;
@@ -481,7 +481,7 @@ static void send_sensor_data(enum sensor_channel channel,
     msg.user_data = NULL;
     msg.error_code = ERROR_IPM_NONE;
     msg.data.sensor.channel = channel;
-    memcpy(msg.data.sensor.value, array, sizeof(double) * 3);
+    memcpy(&msg.data.sensor.reading, &reading, sizeof(union sensor_reading));
     zjs_ipm_send(MSG_ID_SENSOR, &msg);
 }
 
@@ -554,8 +554,11 @@ static void process_accel_data(struct device *dev)
     if (ABS(dval[0] - accel_last_value[0]) > threshold ||
         ABS(dval[1] - accel_last_value[1]) > threshold ||
         ABS(dval[2] - accel_last_value[2]) > threshold) {
-        memcpy(accel_last_value, dval, sizeof(double) * 3);
-        send_sensor_data(SENSOR_CHAN_ACCEL_ANY, dval);
+        union sensor_reading reading;
+        reading.x = dval[0];
+        reading.y = dval[1];
+        reading.z = dval[2];
+        send_sensor_data(SENSOR_CHAN_ACCEL_ANY, reading);
     }
 
 #ifdef DEBUG_BUILD
@@ -582,11 +585,14 @@ static void process_gyro_data(struct device *dev)
     dval[1] = convert_sensor_value(&val[1]);
     dval[2] = convert_sensor_value(&val[2]);
 
-    if (ABS(dval[0] - accel_last_value[0]) > 0 ||
-        ABS(dval[1] - accel_last_value[1]) > 0 ||
-        ABS(dval[2] - accel_last_value[2]) > 0) {
-        memcpy(gyro_last_value, dval, sizeof(double) * 3);
-        send_sensor_data(SENSOR_CHAN_GYRO_ANY, dval);
+    if (ABS(dval[0] - gyro_last_value[0]) > 0 ||
+        ABS(dval[1] - gyro_last_value[1]) > 0 ||
+        ABS(dval[2] - gyro_last_value[2]) > 0) {
+        union sensor_reading reading;
+        reading.x = dval[0];
+        reading.y = dval[1];
+        reading.z = dval[2];
+        send_sensor_data(SENSOR_CHAN_GYRO_ANY, reading);
     }
 
 #ifdef DEBUG_BUILD
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -432,7 +432,7 @@ static void handle_glcd(struct zjs_ipm_message* msg)
 #endif
 
 #ifdef	BUILD_MODULE_SENSOR
-#ifndef DEBUG_BUILD
+#ifdef DEBUG_BUILD
 static inline int sensor_value_snprintf(char *buf, size_t len,
                                         const struct sensor_value *val)
 {
@@ -485,7 +485,7 @@ static void send_sensor_data(enum sensor_channel channel,
     zjs_ipm_send(MSG_ID_SENSOR, &msg);
 }
 
-#define ABS(x) ((x) > 0) ? (x) : -(x)
+#define ABS(x) ((x) >= 0) ? (x) : -(x)
 
 static double convert_sensor_value(const struct sensor_value *val)
 {
@@ -545,7 +545,7 @@ static void process_accel_data(struct device *dev)
     dval[2] = convert_sensor_value(&val[2]);
 
     if (dval[0] == 0 && dval[1] == 0 && dval[2] == 0) {
-        // FIXME: BUG? why sometimes it reports 0, 0, 0 on all axis
+        // FIXME: BUG? why sometimes it reports 0, 0, 0 on all axes
         return;
     }
 
@@ -585,9 +585,9 @@ static void process_gyro_data(struct device *dev)
     dval[1] = convert_sensor_value(&val[1]);
     dval[2] = convert_sensor_value(&val[2]);
 
-    if (ABS(dval[0] - gyro_last_value[0]) > 0 ||
-        ABS(dval[1] - gyro_last_value[1]) > 0 ||
-        ABS(dval[2] - gyro_last_value[2]) > 0) {
+    if (dval[0] != gyro_last_value[0] ||
+        dval[1] != gyro_last_value[1] ||
+        dval[2] != gyro_last_value[2]) {
         union sensor_reading reading;
         reading.x = dval[0];
         reading.y = dval[1];
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -12,6 +12,9 @@
 #ifdef BUILD_MODULE_I2C
 #include <i2c.h>
 #endif
+#ifdef BUILD_MODULE_SENSOR
+#include <sensor.h>
+#endif
 #ifdef BUILD_MODULE_GROVE_LCD
 #include <display/grove_lcd.h>
 #endif
@@ -57,6 +60,13 @@ static struct device *i2c_device[MAX_I2C_BUS];
 static struct device *glcd = NULL;
 static char str[MAX_BUFFER_SIZE];
 #endif
+
+// BMI160 sensor
+static struct device *bmi160 = NULL;
+static bool accel_poll = false;
+static bool gyro_poll = false;
+static double accel_last_value[3];
+static double gyro_last_value[3];
 
 // add strnlen() support for security since it is missing
 // in Zephyr's minimal libc implementation
@@ -356,6 +366,7 @@ static void handle_glcd(struct zjs_ipm_message* msg)
             glcd = device_get_binding(GROVE_LCD_NAME);
 
             if (!glcd) {
+                PRINT("failed to initialize Grove LCD\n");
                 error_code = ERROR_IPM_OPERATION_FAILED;
             } else {
                 DBG_PRINT("Grove LCD initialized\n");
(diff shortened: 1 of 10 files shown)


arc/src/main.c
```diff
@@ -322,6 +324,7 @@ static void handle_glcd(struct zjs_ipm_message* msg)
             glcd = device_get_binding(GROVE_LCD_NAME);
 
             if (!glcd) {
+                PRINT("failed to initialize Grove LCD\n");
```
kenchris: Seems unrelated

src/zjs_sensor.c
```diff
@@ -0,0 +1,512 @@
...
+        if (jerry_value_has_error_flag(rval)) {
+            PRINT("zjs_sensor_set_state: error calling onstatechange\n");
+        }
+        jerry_release_value(rval);
+    }
+    jerry_release_value(func);
+
+    if (old_state == SENSOR_STATE_ACTIVATING &&
+        state == SENSOR_STATE_ACTIVATED) {
+        func = zjs_get_property(obj, "onactivate");
```
kenchris: Great that you are actively following the spec, this was just added two days ago :-)

https://github.com/w3c/sensors/commit/a85830bef0566b21176ac5af1ba0b1d8241b06ca

src/zjs_sensor.c
```diff
@@ -0,0 +1,512 @@
...
+#include "zjs_sensor.h"
+#include "zjs_callbacks.h"
+#include "zjs_ipm.h"
+#include "zjs_util.h"
+
+#define ZJS_SENSOR_TIMEOUT_TICKS                      500
+
+static struct nano_sem sensor_sem;
+
+enum sensor_state {
```
kenchris: I believe these might become internals: Check

https://github.com/w3c/sensors/issues/104#issuecomment-248316955

(that part is not fixed in spec yet, ie the checkbox isnt checked yet)

kenchris: We discussed these changes back in the W3C TPAC meeting in Lisbon and they are slowly making it into the spec

src/zjs_sensor.c
```diff
@@ -0,0 +1,512 @@
...
+
+        double option_freq;
+        if (zjs_obj_get_double(options, "frequency", &option_freq)) {
+            // TODO: figure out a list of frequencies we can support
+            // and have Zephyr trigger value changes instead of
+            // polling on the ARC.
+            // For now, frequency is ignored,  we just report new event
+            // as soon as we detect a value change.
+            if (option_freq <= 0) {
+                PRINT("zjs_sensor_create: invalid frequency, default to 50hz\n");
```
kenchris: defaulting

src/zjs_sensor.c
```diff
@@ -0,0 +1,512 @@
...
+            }
+        } else {
+            PRINT("zjs_sensor_create: frequency not found, default to 50hz\n");
+        }
+
+        if (channel == SENSOR_CHAN_ACCEL_ANY) {
+            bool option_gravity;
+            if (zjs_obj_get_boolean(options, "includeGravity", &option_gravity) &&
+                option_gravity) {
+                // TODO: find out if BMI160 can be configured to include gravity
```
kenchris: I guess the math to add it manually wouldn't be too hard

src/zjs_ipm.h
```diff
@@ -95,6 +124,15 @@ typedef struct zjs_ipm_message {
             uint8_t color_b;
             void *buffer;
         } glcd;
+#endif
+
+#ifdef BUILD_MODULE_SENSOR
+        struct sensor_data {
```
kenchris: For external sensors (ie DHT22 etc) we would probably need something like "controller" here so that I can do

new Hygrometer( { controller: "dht22" } ) which would hook up to the right Zephyr driver

scripts/analyze.sh
```diff
@@ -104,6 +104,11 @@ if [ $? -eq 0 ] && [[ $MODULE != *"BUILD_MODULE_BUFFER"* ]]; then
     >&2 echo Using module: Buffer
     MODULES+=" -DBUILD_MODULE_BUFFER"
 fi
+sensor=$(grep -E AccelerometerSensor\|Gyroscope  $SCRIPT)
```
kenchris: I thought the Sensor part was being removed... I will ask Riju to look at this PR

samples/Accelerometer.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2016, Intel Corporation.
+
+// Test code to use the AccelerometerSensor (subclass of Generic Sensor) API
+// to communicate with the BMI160 inertia sensor on the Arduino 101
+// and obtaining information about acceleration applied to the X, Y and Z axis
+console.log("Acclerometer test...");
```
kenchris: spelling error

samples/Gyroscope.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2016, Intel Corporation.
+
+// Test code to use the Gyroscope (subclass of Generic Sensor) API
+// to communicate with the BMI160 inertia sensor on the Arduino 101
+// and monitor the rate of rotation around the the X, Y and Z axis
+print("Gyroscope test...");
```
kenchris: Riju, do you have tests from Blink that we can reuse?

src/zjs_sensor.c
```diff
@@ -0,0 +1,512 @@
...
+    jerry_value_t x_val = jerry_create_number(x);
+    jerry_value_t y_val = jerry_create_number(y);
+    jerry_value_t z_val = jerry_create_number(z);
+    jerry_value_t reading_obj = jerry_create_object();
+    if (channel == SENSOR_CHAN_ACCEL_ANY) {
+        zjs_set_property(reading_obj, "accelerationX", x_val);
+        zjs_set_property(reading_obj, "accelerationY", y_val);
+        zjs_set_property(reading_obj, "accelerationZ", z_val);
+    } else if (channel == SENSOR_CHAN_GYRO_ANY) {
+        zjs_set_property(reading_obj, "rotationRateX", x_val);
```
riju: same here

samples/Accelerometer.js
```diff
@@ -0,0 +1,29 @@
...
+
+
+var sensor = new AccelerometerSensor({
+    includeGravity: false, // true is not supported, will throw error
+    frequency: 60
+});
+
+sensor.onchange = function(event) {
+    console.log("acceleration (m/s^2): " +
+          " x=" + event.reading.accelerationX +
```
riju: accelerationX -> x , changed recently
https://w3c.github.io/accelerometer/#accelerometerreading

samples/Gyroscope.js
```diff
@@ -0,0 +1,26 @@
...
+          " y=" + event.reading.rotationRateY +
+          " z=" + event.reading.rotationRateZ);
+};
+
+sensor.onstatechange = function(event) {
+    console.log("state: " + event);
+};
+
+sensor.onerrorchange = function(error) {
+    console.log("error: " + error);
```
riju: nit : sensor.onerror = event => console.log(event.error.name, event.error.message);

scripts/analyze.sh
riju: Yes, we have remoed the "sensor" part. https://w3c.github.io/accelerometer/#accelerometer

samples/Gyroscope.js
riju: We have the AmbientLightSensor tests here -> https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/sensor/ambient-light-sensor.html?q=ambient&sq=package:chromium&l=1

Gyro and Accelerometer test cases will be upstreamed along with their implementation soon.

samples/Accelerometer.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2016, Intel Corporation.
+
+// Test code to use the AccelerometerSensor (subclass of Generic Sensor) API
+// to communicate with the BMI160 inertia sensor on the Arduino 101
+// and obtaining information about acceleration applied to the X, Y and Z axis
+console.log("Acclerometer test...");
+
+
+var sensor = new AccelerometerSensor({
```
riju: AccelerometerSensor -> Accelerometer
Just a heads up, very soon, we will split the Accelerometer interface to 2 interfaces, one for Accelerometer and another for LinearAccelerometer. In that case includeGravity will not be needed. You may want to change this when it is reflected in the spec.

samples/Accelerometer.js
```diff
@@ -0,0 +1,29 @@
...
+          " x=" + event.reading.accelerationX +
+          " y=" + event.reading.accelerationY +
+          " z=" + event.reading.accelerationZ);
+};
+
+sensor.onstatechange = function(event) {
+    console.log("state: " + event);
+};
+
+sensor.onerrorchange = function(error) {
```
riju: sensor.onerror = event => console.log(event.error.name, event.error.message);

src/zjs_ipm.h
```diff
@@ -95,6 +124,15 @@ typedef struct zjs_ipm_message {
...
+            enum sensor_channel channel;
+            uint32_t frequency;
+            double value[3];
```
riju: Would you prefer it to split this struct into 2 ? Like we have for SensorReading (frequency&timestamp) as base struct, and say concreteSensor's reading (which contains the actual value) like https://w3c.github.io/gyroscope/#gyroscopereading. 
Also no timestamp ? like performance.now() equivalent .

When you implement the next sensor, say AmbientLightSensor , there is only 1 double for "value".
If you implement Uncalibrated Magnetometer or Uncalibrated Gyro, then double value[6].

src/zjs_sensor.c
```diff
@@ -0,0 +1,512 @@
...
+    SENSOR_STATE_ACTIVATING,
+    SENSOR_STATE_ACTIVATED,
+    SENSOR_STATE_ERRORED,
+};
+
+typedef struct sensor_handle {
+    int32_t id;
+    enum sensor_channel channel;
+    enum sensor_state state;
+    double sensor_value[3];
```
riju: Fine for now, but not generic enough. See https://github.com/01org/zephyr.js/pull/353/commits/294d36b85640c752a6840df72dbe3073278b9aa7#r86344326

arc/src/main.c
jimmy-huang: Yes, I forgot to add it to the glcd when I am implementing sensor, so I am thinking to include it as part of the patch, but I'll remove it and throw in another patch.

scripts/analyze.sh
jimmy-huang: Ok, I'll update accordingly, I was wondering that too since it's odd that the one has, but one does not.

samples/Gyroscope.js
```diff
@@ -0,0 +1,26 @@
...
+// to communicate with the BMI160 inertia sensor on the Arduino 101
+// and monitor the rate of rotation around the the X, Y and Z axis
+print("Gyroscope test...");
+
+
+var sensor = new Gyroscope();
+
+sensor.onchange = function(event) {
+    print("rotation (rad/s): " +
+          " x=" + event.reading.rotationRateX +
```
jimmy-huang: I'll update accordingly.

samples/Gyroscope.js
jimmy-huang: This is a limitation of the jerryscript engine, it doesn't support this syntax, it's ES6 right?  It only supports ES5 for now.

samples/Accelerometer.js
jimmy-huang: Ok, I'll keep that in mind, thanks for the heads up.

samples/Accelerometer.js
jimmy-huang: See my comment above.

src/zjs_ipm.h
```diff
@@ -59,7 +77,16 @@ enum {
...
 #define TYPE_GLCD_GET_INPUT_STATE                          0x002B
+#endif
 
+// SENSOR
+#ifdef BUILD_MODULE_SENSOR
+#define TYPE_SENSOR_INIT                                   0x0030
+#define TYPE_SENSOR_START                                  0x0031
+#define TYPE_SENSOR_STOP                                   0x0032
+#define TYPE_SENSOR_EVENT_STATE_CHANGE                     0x0033
+#define TYPE_SENSOR_EVENT_VALUE_CHANGE                     0x0034
```
jimmy-huang: that's fine with me

src/zjs_ipm.h
jimmy-huang: ok yeah, this is the internal implementation, and only specific to Arduino 101 where we have to channel the data, so we can fix it up however we want, but that's a good idea.

src/zjs_sensor.c
jimmy-huang: that's one possible way to do it

samples/Gyroscope.js
riju: good to know that. iirc, its better to use event, instead of error object. not sure, right now.

```
sensor.onerror = function(event) {
   console.log(event.error.name, event.error.message);
```

src/zjs_ipm.h
jimmy-huang: @riju I updated patch to use a union to hold the reading values depending on the sensor type

arc/src/main.c
```diff
@@ -388,6 +388,250 @@ static void handle_glcd(struct zjs_ipm_message* msg)
     ipm_send_reply(msg);
 }
 
+#ifndef DEBUG_BUILD
```
grgustaf: I think you meant #ifdef? This is used only in DEBUG_BUILD below AFAICT. Bad Jimmy, didn't build with DEBUG! :)

arc/src/main.c
```diff
@@ -388,6 +388,250 @@ static void handle_glcd(struct zjs_ipm_message* msg)
...
+    msg.type = TYPE_SENSOR_EVENT_READING_CHANGE;
+    msg.flags = 0;
+    msg.user_data = NULL;
+    msg.error_code = ERROR_IPM_NONE;
+    msg.data.sensor.channel = channel;
+    memcpy(&msg.data.sensor.reading, &reading, sizeof(union sensor_reading));
+    zjs_ipm_send(MSG_ID_SENSOR, &msg);
+}
+
+#define ABS(x) ((x) > 0) ? (x) : -(x)
```
grgustaf: &gt;= perhaps?

arc/src/main.c
```diff
@@ -388,6 +388,250 @@ static void handle_glcd(struct zjs_ipm_message* msg)
...
+            val2 = val->val2;
+        }
+
+        if (val1 > 0 || (val1 == 0 && val2 > 0)) {
+            result = val1 + (double)val2 * 0.000001;
+        } else if (val1 == 0 && val2 < 0) {
+            result = (double)val2 * (-0.000001);
+        } else {
+            result = val1 + (double)val2 * (-0.000001);
+        }
```
grgustaf: What happens in this block if val1 > 0 and val2 < 0, doesn't it do the wrong thing?

It kind of seems like this logic could be simplified to be more obviously correct. I'm still not sure what it's doing but will wait for your response to this before digging deeper. :)

arc/src/main.c
```diff
@@ -388,6 +388,250 @@ static void handle_glcd(struct zjs_ipm_message* msg)
...
+        PRINT("Cannot read accelerometer channels.\n");
+        return;
+    }
+
+    dval[0] = convert_sensor_value(&val[0]);
+    dval[1] = convert_sensor_value(&val[1]);
+    dval[2] = convert_sensor_value(&val[2]);
+
+    if (dval[0] == 0 && dval[1] == 0 && dval[2] == 0) {
+        // FIXME: BUG? why sometimes it reports 0, 0, 0 on all axis
```
grgustaf: axis -> axes.  Did you report that to Zephyr folks? Maybe open a JIRA?

arc/src/main.c
```diff
@@ -388,6 +388,250 @@ static void handle_glcd(struct zjs_ipm_message* msg)
...
+        return;
+    }
+
+    // set slope threshold to 0.1G (0.1 * 9.80665 = 4.903325 m/s^2)
+    double threshold = 0.980665;
```
grgustaf: Now that's some accurate gravity, right there. Is that calibrated to the second floor of JF1?

arc/src/main.c
```diff
@@ -388,6 +388,250 @@ static void handle_glcd(struct zjs_ipm_message* msg)
...
+        return;
+    }
+
+    dval[0] = convert_sensor_value(&val[0]);
+    dval[1] = convert_sensor_value(&val[1]);
+    dval[2] = convert_sensor_value(&val[2]);
+
+    if (ABS(dval[0] - gyro_last_value[0]) > 0 ||
+        ABS(dval[1] - gyro_last_value[1]) > 0 ||
+        ABS(dval[2] - gyro_last_value[2]) > 0) {
```
grgustaf: If your threshold for difference is 0, then isn't this the same as dval[0] != gyro_last_value[0]?

src/zjs_ipm.h
```diff
@@ -4,15 +4,27 @@
...
+#endif
+#ifdef BUILD_MODULE_I2C
 #define MSG_ID_I2C                                         0x02
+#endif
+#ifdef BUILD_MODULE_GROVE_LCD
 #define MSG_ID_GLCD                                        0x03
+#endif
+#ifdef BUILD_MODULE_SENSOR
+#define MSG_ID_SENSOR                                      0x04
+#endif
```
grgustaf: I think you can go ahead and define all of these all the time, without the ifdefs. And actually I think elsewhere you use them without being in an ifdef so that would break the build anyway. Maybe review your usage of MSG_ID_* everywhere to make sure it makes sense.

src/zjs_ipm.h
```diff
@@ -26,27 +38,33 @@ enum {
...
 // AIO
+#ifdef BUILD_MODULE_AIO
 #define TYPE_AIO_OPEN                                      0x0000
 #define TYPE_AIO_PIN_READ                                  0x0001
 #define TYPE_AIO_PIN_ABORT                                 0x0002
 #define TYPE_AIO_PIN_CLOSE                                 0x0003
 #define TYPE_AIO_PIN_SUBSCRIBE                             0x0004
 #define TYPE_AIO_PIN_UNSUBSCRIBE                           0x0005
 #define TYPE_AIO_PIN_EVENT_VALUE_CHANGE                    0x0006
+#endif
```
grgustaf: Defining macros like this that are never used (if the corresponding module isn't built) doesn't really hurt anything, so I doubt we need these #ifdefs?

src/zjs_ipm.h
```diff
@@ -69,13 +96,14 @@ typedef struct zjs_ipm_message {
     uint32_t error_code;
 
     union {
-        // AIO
+#ifdef BUILD_MODULE_AIO
```
grgustaf: Even defining structs that don't get used shouldn't really hurt anything. A few microseconds faster built time but more clutter in the code.

src/zjs_sensor.c
```diff
@@ -0,0 +1,521 @@
...
+
+#define ZJS_SENSOR_TIMEOUT_TICKS                      500
+
+static struct nano_sem sensor_sem;
+
+enum sensor_state {
+    SENSOR_STATE_IDLE,
+    SENSOR_STATE_ACTIVATING,
+    SENSOR_STATE_ACTIVATED,
+    SENSOR_STATE_ERRORED,
```
grgustaf: Is this non-word Errored something you're inheriting from elsewhere? :)

grgustaf:
Trying out Accelerometer.js, when my board is sitting flat it reports something like:
acceleration (m/s^2):  x=-0.92958 y=0.25773799999999997 z=11.353628

We would expect x=0, y=0, z=9.81; so y is understandable but x and z seem surprisingly off I think?

grgustaf:
It would be very nice if the output of the sample could be printed with a fixed width, so the y and z values line up as they're spewing by, then you could watch them go up and down without having to pause the stream.

src/zjs_sensor.c
```diff
@@ -0,0 +1,521 @@
...
+        zjs_sensor_update_reading(handle->sensor_obj,
+                                  handle->channel,
+                                  handle->reading);
+    }
+}
+
+static void zjs_sensor_signal_callbacks(sensor_handle_t *handle,
+                                        union sensor_reading reading)
+{
+    if (!handle)
```
grgustaf: This isn't needed because your loop will be skipped if it's NULL.

grgustaf:
I can confirm the calibration made the flat on desk case more accurate. If I tip it up on short end or long end it shows x=-8.3 though or y=-8.3, where i'd expect -9.81. But flipped the opposite way where gravity is positive they are around 9.8. Not sure if it can get better, especially without a dynamic user involved calibration?

arc/src/main.c
jimmy-huang: yeah, lol

arc/src/main.c
jimmy-huang: ok, i'll change it

arc/src/main.c
jimmy-huang: if val1 > 0, then it falls into the first case, and if val1 > 0, then val2 can't be < 0 as the sign bit is in val1.  But anyway, this is taken from the the zephyr sample how they checked.

arc/src/main.c
jimmy-huang: I think this will be fixed with the trigger, that's why I wasn't looking into this.

arc/src/main.c
jimmy-huang: yeah, i was going to add a threshold here intead of 0, but I don't really know what value to put in for now, so i left it as 0.  I'll update it.

src/zjs_ipm.h
jimmy-huang: ok, i'll revert them.

grgustaf:
Jimmy and I found that the last two commits segfault if you try to run AIO.js. If you run the BTGrove demo, it does the same once you start to connect.

samples/Accelerometer.js
```diff
@@ -0,0 +1,30 @@
...
+// and obtaining information about acceleration applied to the X, Y and Z axis
+console.log("Accelerometer test...");
+
+
+var sensor = new Accelerometer({
+    includeGravity: false, // true is not supported, will throw error
+    frequency: 50
+});
+
+sensor.onchange = function(event) {
```
riju: @jimmy-huang : Some changes were made for the [Sensor Reading](https://w3c.github.io/sensors/#sensor-reading)
Now
```js
accleromterSensor.onchange = function () {
console.log("acceleration (m/s^2): " +
        " x=" + accleromterSensor.reading.x +
        " y=" + accleromterSensor.reading.y +
        " z=" + accleromterSensor.reading.z);
}
```
just a heads up, you can make this changes in subsequent PR also, upto you.

samples/Accelerometer.js
```diff
@@ -0,0 +1,30 @@
...
+    console.log("acceleration (m/s^2): " +
+                " x=" + event.reading.x +
+                " y=" + event.reading.y +
+                " z=" + event.reading.z);
+};
+
+sensor.onstatechange = function(event) {
```
riju: onstatechange is also removed

samples/Gyroscope.js
```diff
@@ -0,0 +1,27 @@
...
+// to communicate with the BMI160 inertia sensor on the Arduino 101
+// and monitor the rate of rotation around the the X, Y and Z axis
+console.log("Gyroscope test...");
+
+
+var sensor = new Gyroscope();
+
+sensor.onchange = function(event) {
+    console.log("rotation (rad/s): " +
+                " x=" + event.reading.x +
```
riju: Please see the comment above for accelerometer

jimmy-huang:
this needs to be rebased once #415 is merged

jimmy-huang:
@grgustaf, updated pr and rebased, please review

PR #358, opened 2016-11-03
[Documentation] Update udev rules

cuiyanx:
It should be "/etc/udev/rules.d" on Ubuntu 16.04


grgustaf:
Thanks, good catch. +1

PR #360, opened 2016-11-03
[general] Add version file for 0.1-pre

grgustaf:
Also, update deps/repos.txt to clearly spell out that we're using
Zephyr 1.5 and JerryScript 1.0 for this branch.


grgustaf:
Not sure why its testing a 'push' nor why it's failing in one case. Will close this and try pushing from a different branch in case that's the problem.

PR #361, opened 2016-11-03
[general] Add version file for 0.1-pre

grgustaf:
Also, update deps/repos.txt to clearly spell out that we're using
Zephyr 1.5 and JerryScript 1.0 for this branch.


grgustaf:
Manually merged this because I couldn't figure out why Travis was attempting to check out a commit that didn't exist in the upstream zjs-0.1 branch. Possibly because I had force pushed and removed that commit when planning to instead submit it via pull request. It's a 99.99% safe commit.

PR #362, opened 2016-11-03
[performance] New module implementing performance counter API.

pfalcon:
~~~
Based on recent W3C API, formally defined in https://www.w3.org/TR/hr-time/
and informally in
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now . As the
last link suggests, real-world usage boils down to just:

    t = performance.now();

even though formal spec defines more properties for "performance" object.

As defined in the spec for a browser, "performance" is a property of
global window object and thus implicitly present into context of a JS
script. We however define it as a module, so usage pattern is:

    var performance = require("performance");
    t = performance.now();

The returned value is a floating-point number of milliseconds since
arbitrary point in time. This means that its absolute value is
meaningless and only difference between 2 values make sense (and
represent time duration between calls which produced these 2 values).
As the returned value is a floating-point number, it may have higher
resolution than a millisecond (using fractional part). The actual
resolution however depends on a particular Zephyr board, and some
boards provide resolution only of 10 milliseconds.

This module is intended for benchmarking, and in extreme cases, for
synchronous delays, etc. Supported both in Linux and Zephyr builds.
A testcase is included.
~~~

### DIFF
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -14,6 +14,7 @@ CORE_SRC = 	src/main.c \
 			src/zjs_event.c \
 			src/zjs_linux_time.c \
 			src/zjs_modules.c \
+			src/zjs_performance.c \
 			src/zjs_script.c \
 			src/zjs_script_gen.c \
 			src/zjs_timers.c \
@@ -30,7 +31,8 @@ JERRY_LIB_PATH = 	-L$(JERRY_BASE)/build/lib/
 
 LINUX_FLAGS = -std=gnu99 -Wpointer-sign
 
-LINUX_DEFINES = -DZJS_LINUX_BUILD -DBUILD_MODULE_EVENTS -DBUILD_MODULE_CONSOLE
+LINUX_DEFINES = -DZJS_LINUX_BUILD -DBUILD_MODULE_EVENTS -DBUILD_MODULE_CONSOLE \
+		-DBUILD_MODULE_PERFORMANCE
 
 ifeq ($(VARIANT), debug)
 LINUX_DEFINES += -DDEBUG_BUILD
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -1,0 +1,58 @@
+ZJS API for Benchmarking
+========================
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+"Performance" module implements a subset of "High Resolution Time" specification
+from W3C, intended primarily for benchmarking purposes. The key point of this
+module is that it is very light-weight by implementing just one function
+(unlike for example Date object).
+
+Web IDL
+-------
+This IDL provides an overview of the interface; see below for documentation of
+specific API functions.
+
+```javascript
+double now();
+```
+
+API Documentation
+-----------------
+### now
+
+`now();
+`
+
+The module exposes just one function, `now()`. It returns current time in
+milliseconds, as a floating-point number, since an arbitrary point in
+time. It is thus not useful as an absolute value, but subtracting values
+from two calls will give a time duration between these two calls.
+
+As the value returned is floating point, it may have higher resolution
+than a millisecond. However, the actual resolution depends on a platform
+and its configuration. For example, default Zephyr configuration for
+many boards provides resolution of only ones or tens of milliseconds.
+
+The intended usage of this function is for benchmarking and other testing
+and development needs.
+
+
+Examples
+--------
+
+    var performance = require("performance");
+
+    t = performance.now()
+    do_long_operation();
+    console.log("Long operation took:", performance.now() - t, "ms");
+
+
+Sample Apps
+-----------
+* [Performance module unit test](../tests/test-performance.js)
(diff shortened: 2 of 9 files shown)


pfalcon:
@grgustaf asked to resubmit https://github.com/01org/zephyr.js/pull/346 against master, here it is.

tests/test-performance.js
```diff
@@ -0,0 +1,26 @@
...
+    var label = "\033[1m\033[31mFAIL\033[0m";
+    if (actual === true) {
+        passed += 1;
+        label = "\033[1m\033[32mPASS\033[0m";
+    }
+    console.log(label + " - " + description);
+}
+
+var before = performance.now();
+//console.log("performance.now() before:", before);
```
jimmy-huang: Remove commented out code.

tests/test-performance.js
```diff
@@ -0,0 +1,26 @@
...
+
+setTimeout(function() {
+    var after = performance.now();
+    //console.log("performance.now() after:", after);
```
jimmy-huang: Remove commented out code.

Makefile.linux
```diff
@@ -30,7 +31,7 @@ JERRY_LIB_PATH = 	-L$(JERRY_BASE)/build/lib/
 
 LINUX_FLAGS = -std=gnu99 -Wpointer-sign
 
-LINUX_DEFINES = -DZJS_LINUX_BUILD -DBUILD_MODULE_EVENTS -DBUILD_MODULE_CONSOLE
+LINUX_DEFINES = -DZJS_LINUX_BUILD -DBUILD_MODULE_EVENTS -DBUILD_MODULE_CONSOLE -DBUILD_MODULE_PERFORMANCE
```
jimmy-huang: wrap this to new line when over 80 chars

docs/performance.md
```diff
@@ -0,0 +1,58 @@
...
+========================
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+"Performance" module implements subset of "High Resolution Time" specification
```
jimmy-huang: a subset?

docs/performance.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+Examples
+--------
+
+    var performance = require("performance");
+
+    t = performance.now()
+    do_long_operation();
+    console.out("Long operation took:", performance.now() - t, "ms");
```
jimmy-huang: console.log?

src/zjs_performance.c
```diff
@@ -0,0 +1,32 @@
...
+// ZJS includes
+#include "zjs_util.h"
+
+static jerry_value_t zjs_performance_now(const jerry_value_t function_obj,
+                                         const jerry_value_t this,
+                                         const jerry_value_t argv[],
+                                         const jerry_length_t argc)
+{
+    if (argc != 0)
+        return zjs_error("performance.now: no args expected");
```
jimmy-huang: use "zjs_performance_now" for easier grep the src

src/zjs_performance.c
```diff
@@ -0,0 +1,32 @@
...
+
+// ZJS includes
+#include "zjs_util.h"
+
+static jerry_value_t zjs_performance_now(const jerry_value_t function_obj,
+                                         const jerry_value_t this,
+                                         const jerry_value_t argv[],
+                                         const jerry_length_t argc)
+{
+    if (argc != 0)
```
jimmy-huang: I don't think you should check argc, because javascript allows you to pass in additional args, they'll just be ignored.

tests/test-performance.js
pfalcon: I will. But note that the moment the test fails, someone would need to re-add this code to diagnose it. Or just uncomment it, which saves time. Apparently, that already helped @grgustaf in https://github.com/01org/zephyr.js/pull/346#discussion_r86405975 . Of course, this version should be final and never fail. But, I thought the same about my original version too, and @grgustaf had problems.

tests/test-performance.js
pfalcon: Done

Makefile.linux
pfalcon: Done

docs/performance.md
pfalcon: Fixed

docs/performance.md
pfalcon: Ah, always mix that up... Fixed.

src/zjs_performance.c
pfalcon: I'm not sure what you mean by "javascript allows" here. WebIDL seems to be a strictly typed representation and it doesn't specify that now accepts arbitrary parameters: https://www.w3.org/TR/hr-time/#the-performance-interface . I'd personally consider it an error to pass parameters to function which doesn't accept them. But then I consider following an error:

~~~
console.log(obj.misspelled_property)
~~~

and JavaScript doesn't. So, I'll be happy to follow if, with the above in mind, you'll confirm that the check should be removed.

src/zjs_performance.c
pfalcon: I'd think that error messages are targeted at users, not developers. If you confirm that it's ok to leak internal things like C function names to users and potentially confuse them, I'll be happy to comply, though given the previous comment, this code may be removed at all, so need re-confirming anyway.

Thanks for the review!

pfalcon:
New rebased/squashed version of the patch uploaded. 2 issues waiting for the confirmation of course of action still. Thanks.

grgustaf:
+1, works great on A101 too. Merging.

PR #363, opened 2016-11-03
End of devel

grgustaf:
Merge in the last couple changes that had made it into devel, into master. Then we'll abandon devel branch and commit directly to master again as the main development branch.


jimmy-huang:
lgtm

Issue #310, opened 2016-10-20
[Timer] Timer alloc failed


cuiyanx:
The timer has limited number when be created, and be shared by setTimeout(), setInterval() and GPIOPin.openAsync(). The maximum of timers is 4 at now. 

Error Log:
    native_set_interval_handler: timer alloc failed
    JerryScript: cannot run javascript

Sometimes it well be stucked, look like stoped. but it well be running and ending after added any functions above setInterval() function.

grgustaf:
This works if you build with MALLOC=heap. @jprestwo maybe look at why this is failing with pools and see if the values should possibly be tuned to allow more than 4 timers. (I didn't verify that's the max.)

cuiyanx:
Although they change timer using MALLOC, but it still do not fix issue 310, and using MALLOC can help to verify our tests result, so keep eyes on 310, i think they should have other fix patch for it.

jprestwo:
So the issue is that there are not enough pool blocks for how big timers are. Timer allocations are 60 bytes, and we have just 10 64 byte slots, meaning we can have at most 10 timers and that is if there is nothing else taking up those 64 byte slots (like the previously mentioned GPIOPin.openAsync or other module allocations). This is just going to require some tuning, but we are going to have to make a decision on the max number of timers we allow. Obviously its quite low now.

The best solution would be to dynamically create a pool configuration based on the amount of RAM available after all the modules are compiled in. This is a bit more difficult but I think its worth looking into.

xiaolongx:
This issue cannot reproduce on both Arduino 101 and FRDM_K64F with commit 2ff271b, but another issue found in FRDM_K64F(#447)

grgustaf:
Apparently the immediate issue is fixed; we'll still want to revisit the effective limit on timers and how we could improve that or make it tunable by the developer.

PR #365, opened 2016-11-04
[timers] Changing how timers are released

brianjjones:
Because ZJS implements its own version of setInterval,
the new code to release global objects had a unintended
effect of causing timers that didn't save the return to
be garbage collected after the first call. Now instead
of relying on JerryScript to GC our timers, we will clear
them ourself when jerry_cleanup runs.

### DIFF
--- a/src/zjs_timers.c
+++ b/src/zjs_timers.c
@@ -93,10 +93,9 @@ static bool delete_timer(int32_t id)
     for (zjs_timer_t **ptm = &zjs_timers; *ptm; ptm = &(*ptm)->next) {
         zjs_timer_t *tm = *ptm;
         if (id == tm->callback_id) {
-            int i;
             zjs_port_timer_stop(&tm->timer);
             *ptm = tm->next;
-            for (i = 0; i < tm->argc; ++i) {
+            for (int i = 0; i < tm->argc; ++i) {
                 jerry_release_value(tm->argv[i]);
             }
             zjs_remove_callback(tm->callback_id);
@@ -108,9 +107,19 @@ static bool delete_timer(int32_t id)
     return false;
 }
 
-static void zjs_timer_free_cb(uintptr_t handle)
+void zjs_timers_free_timers()
 {
-    delete_timer(((zjs_timer_t *)handle)->callback_id);
+    while (zjs_timers) {
+        zjs_timer_t *tm = zjs_timers;
+        zjs_port_timer_stop(&tm->timer);
+        zjs_timers = tm->next;
+        for (int i = 0; i < tm->argc; ++i) {
+            jerry_release_value(tm->argv[i]);
+        }
+        zjs_remove_callback(tm->callback_id);
+        zjs_free(tm->argv);
+        zjs_free(tm);
+    }
 }
 
 static jerry_value_t add_timer_helper(const jerry_value_t function_obj,
@@ -131,7 +140,7 @@ static jerry_value_t add_timer_helper(const jerry_value_t function_obj,
     zjs_timer_t* handle = add_timer(interval, callback, this, repeat, argv, argc - 2);
     if (handle->callback_id == -1)
         return zjs_error("native_set_interval_handler: timer alloc failed");
-    jerry_set_object_native_handle(timer_obj, (uintptr_t)handle, zjs_timer_free_cb);
+    jerry_set_object_native_handle(timer_obj, (uintptr_t)handle, NULL);
 
     return timer_obj;
 }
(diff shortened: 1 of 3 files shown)


brianjjones:
@jimmy-huang In reference to your question about why I put the restore_zjs_api where I did:
I feel this is the best place for it as we shouldn't call it unless the javascript has been stopped. Note in ashell the user can either call stop directly or choose to run another JS file, which calls stop for you. Basically, there are multiple places where stop could be called but I only want restore_zjs_api called when stop occurs. So I think it's better it stays where it is.

@grgustaf I updated the logic and tested with multiple timers and it works fine for me.  I see them all being released as they should be.

Sorry I had to close and restart - rebase just wasn't working for me.

src/zjs_timers.c
```diff
@@ -131,7 +140,7 @@ static jerry_value_t add_timer_helper(const jerry_value_t function_obj,
     zjs_timer_t* handle = add_timer(interval, callback, this, repeat, argv, argc - 2);
     if (handle->callback_id == -1)
         return zjs_error("native_set_interval_handler: timer alloc failed");
-    jerry_set_object_native_handle(timer_obj, (uintptr_t)handle, zjs_timer_free_cb);
```
grgustaf: This isn't right, we should still free the timer immediately if it gets garbage-collected before the javascript_stop occurs.

src/zjs_timers.h
```diff
@@ -5,5 +5,6 @@
 
 void zjs_timers_process_events();
 void zjs_timers_init();
+void zjs_timers_free_timers();
```
grgustaf: The name is kind of awkward. Maybe zjs_timers_cleanup? Also, add a comment above it here describing what it does - stops and frees all timers. We should move toward @jprestwo's method of describing the APIs here in their header files and using jsdoc format.

src/zjs_timers.c
```diff
@@ -108,9 +107,19 @@ static bool delete_timer(int32_t id)
...
 }
 
-static void zjs_timer_free_cb(uintptr_t handle)
+void zjs_timers_free_timers()
 {
-    delete_timer(((zjs_timer_t *)handle)->callback_id);
+    while (zjs_timers) {
+        zjs_timer_t *tm = zjs_timers;
+        zjs_port_timer_stop(&tm->timer);
+        zjs_timers = tm->next;
```
grgustaf: I'd move this to the bottom of the while() to make it more clear that this is the loop incrementer. Sorry for a second pass of comments, but they're quick ones. :)

src/zjs_timers.c
brianjjones: If I give JrS the ability to GC the timer it will do so if you don't assign the return value of setInterval. The other option was to just keep the previous change I made, allow JrS to handle the GC of timers, and require users always save the return value of setInterval in their JavaScript. i.e. use - var timers = setInterval(myFunc, 1000);  At least as far as I know, there's no way for this code to know the GC is intended or not.

brianjjones: Also I'm no JavaScript expert, but I thought the only reason setInterval should be GC'd is if clearInterval is called or the script ends.  We could pass the zjs_timer_free_cb for setTimeout though.  Would that be more complete in your opinion?

grgustaf: Setting this to NULL doesn't mean that JrS isn't GC'ing the object, it just means you're not informed about it.

But anyway you're right, I forgot about that - the timer should not just get GC'd like that. However, I figured out the issue I think - see my next comment.

src/zjs_timers.c
```diff
@@ -131,7 +140,7 @@ static jerry_value_t add_timer_helper(const jerry_value_t function_obj,
     zjs_timer_t* handle = add_timer(interval, callback, this, repeat, argv, argc - 2);
     if (handle->callback_id == -1)
         return zjs_error("native_set_interval_handler: timer alloc failed");
-    jerry_set_object_native_handle(timer_obj, (uintptr_t)handle, zjs_timer_free_cb);
+    jerry_set_object_native_handle(timer_obj, (uintptr_t)handle, NULL);
 
     return timer_obj;
```
grgustaf: Ah shoot this is what we're doing wrong. We're not supposed to return an object, we're supposed to return an ID number. So it's okay for that ID to get "GC'd", that doesn't make the timer go away.

Instead of creating an object above we should be assigning a unique number: 1, 2, 3... I guess to be really nice we would reset it back to 1 again after cleanup.

I guess your call whether you want to try to solve that now or we can follow up with a later patch.

grgustaf:
I verified Timers.js runs, but test-timers.js doesn't. However that is bug #310 and I verified that if you build with MALLOC=heap it works.

+1, merging.

PR #366, opened 2016-11-04
Update samples/arduino/basics/Blink-k64f.js

xiaolongx:
Should use pins.LEDG instead of pins.D5 in sample Blink-k64f.js


grgustaf:
+1. Good catch, thanks.

PR #388, opened 2016-11-04
[submodules] Use git submodules

jprestwo:


jprestwo:
@poussa, yes, it just simplifies things so much. And removes that huge code block in the Makefile that parses repos.txt. Plus it makes things easier if a developer wants to use any commit for zephyr/jerry. They don't have to edit any files, just go in and check out the commit they want.

pfalcon:
As someone who complained about current adhoc setup, +1 too. (But didn't test yet.)

grgustaf:
LGTM, yeah thanks for the pointer, @pfalcon

PR #390, opened 2016-11-04
[deps] Remove old deps scripts and update deps README as README.md

grgustaf:


jimmy-huang:
+1

PR #395, opened 2016-11-08
[Tests] Add tests for GPIO APIs

cuiyanx:
I am recoding test-gpio.js tests case for fixing bug.


tests/test-gpio.js
```diff
@@ -73,35 +73,39 @@ bValue = pinB.read();
 assert(bValue, "activeLow: true");
```
haoxli: Notice 100644 → 100755, please remove execution permission

PR #398, opened 2016-11-09
[travis] Add tests so that all modules get built

grgustaf:
James had a good point that we could have a simple test that just
requires everything; but this seemed more fun for the moment. Maybe
we could even generate that test from the module list to keep from
missing any? Needs a bit more thought.


jimmy-huang:
+1

PR #400, opened 2016-11-09
[script] Multi JS file support

brianjjones:
catjs allows you to use simple ES6 module notation to import and
export javascript.  This will work as a placeholder until ES6
module support is added to JerryScript.

### DIFF
--- a/samples/TempLCD/BMP280Lib.js
+++ b/samples/TempLCD/BMP280Lib.js
@@ -1,0 +1,104 @@
+// Copyright (c) 2016, Intel Corporation.
+// JavaScript library for the BMP280 sensor
+
+export 	 function BMP280() {
+
+    // API object
+    var bmp280API = {};
+
+    var i2c = require("i2c");
+    bmp280API.i2cDevice = i2c.open({ bus: 0, speed: 100 });
+
+    var bmp280 = {
+        ADDRESS: 0x77,
+        NORMAL_MODE: 0x03,
+        REGISTER_DIG_T1: 0x88, // Unsigned Short
+        REGISTER_DIG_T2: 0x8A, // Signed Short
+        REGISTER_DIG_T3: 0x8C, // Signed Short
+        REGISTER_CHIPID: 0xD0,
+        REGISTER_CONTROL: 0xF4,
+        REGISTER_CONFIG: 0xF5,
+        REGISTER_TEMPDATA: 0xFA, // Signed 20bit. Always positive
+        STANDBY: 2 << 5,
+        FILTER: 0x00,
+        SPI_3W_DISABLE: 0x00,
+        PRESS_OVER: 1 << 2,
+        TEMP_OVER: 1 << 5
+    }
+
+    var dig_T1;
+    var dig_T2;
+    var dig_T3;
+
+    function uint16toInt16(value) {
+        if ((value & 0x8000) > 0) {
+            value = value - 0x10000;
+        }
+        return value;
+    }
+
+    bmp280API.readCoefficients = function() {
+        // These need to be read as a burst read in order to get accurate numbers,
+        // so dig_TAll contains dig_T1 through dig_T3
+        var dig_TAll = this.i2cDevice.burstRead(bmp280.ADDRESS, 24,
+                                                bmp280.REGISTER_DIG_T1);
+        dig_T1 = dig_TAll.readUInt16LE();
+        dig_T2 = dig_TAll.readUInt16LE(2);
+        dig_T3 = dig_TAll.readUInt16LE(4);
+
+        // Convert these to signed
+        dig_T1 = uint16toInt16(dig_T1);
+        dig_T2 = uint16toInt16(dig_T2);
+        dig_T3 = uint16toInt16(dig_T3);
+    }
+
+    bmp280API.readTemperature = function() {
+        // Read the raw temperature value.
+        var tempBuffer = this.i2cDevice.burstRead(bmp280.ADDRESS, 4,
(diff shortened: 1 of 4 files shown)


samples/TempLCD/BMP280Lib.js
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2016, Intel Corporation.
+// JavaScript library for the BMP280 sensor
+
+export 	 function BMP280() {
```
jimmy-huang: spaces gap here

samples/TempLCD/BMP280Lib.js
```diff
@@ -0,0 +1,104 @@
...
+        var tempDec = tempBuffer.readUInt32BE();
+
+        // Selects the first 20 bits of 32 bit integer and maintains the sign
+        tempDec >>>= 12;
+
+        // Convert raw temperature to Celsius using the algorithm found here
+        // https://www.bosch-sensortec.com/bst/products/all_products/bmp280
+        var var1 = (((tempDec >> 3) - (dig_T1 << 1)) * (dig_T2)) >> 11;
+        var var2 = (((((tempDec >> 4) - (dig_T1)) * ((tempDec >> 4) - (dig_T1))) >> 12) *
+                    (dig_T3)) >> 14;
```
jimmy-huang: align here

samples/TempLCD/BMP280Lib.js
```diff
@@ -0,0 +1,104 @@
...
+        return T;
+    }
+
+    function init() {
+        var reader = bmp280API.i2cDevice.read(bmp280.ADDRESS, 1,
+                                              bmp280.REGISTER_CHIPID);
+
+        // Check that we have a BMP280 on the I2C bus
+        if (reader.toString('hex') !== "58")
+            print("BMP280 not found! ChipId " + reader.toString('hex') !== "58");
```
jimmy-huang: console.log

samples/TempLCD/GroveLCDLib.js
```diff
@@ -0,0 +1,163 @@
...
+        // Write blank space to LCD
+        this.writeText(blanks);
+    }
+
+    groveLCDAPI.resetCursor = function (col, row) {
+        // Reset the cursor position
+        row = row || 0;
+        col = col || 0;
+
+        if (row == 0) {
```
jimmy-huang: ===?

samples/TempLCD/GroveLCDLib.js
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) 2016, Intel Corporation.
+// JavaScript library for the Grove LCD
+
+export function GroveLCD() {
```
grgustaf: Well, this comes from the I2C demo code where the point was to show how to use the I2C API.

jprestwo:
@brianjjones: I have been thinking about this and there may be a way to do it without all this pre-processing and JS code replacement (e.g. export function). The way that I understand Node.js module working is that you add all your module functions to an object, and add that object to "module.exports". We could do something similar to this but require that the name of the object be the "require" name. For example:
```
var myModule;
myModule.helloWorld = function() {
    print("hello world");
};

module.exports.myModule = myModule;
```

During our initialization, code we could create the global object "module" with element "exports". Then in the `require` C function, if we search and don't find any C modules, we look in `module.exports.<name>` and if its found we return that object. This would require that we concatenate all the JS files into one file still, but it would be more similar to how Node.js modules work. This is all in theory though, it may just not be possible but I figured I would throw it out there.

samples/TempLCD/GroveLCDLib.js
jimmy-huang: @grgustaf, well, I don't agree with that, the I2C.js demo is exactly what that's for, this patch is show the demo of multi-js, the bmp280 is a indeed good example because we don't have drivers for it, so it gives application an example how to extend the it.  Now if this I2C talks with another LCD like the ones we ordered that does more fancy things, that would be cool.

samples/TempLCD/BMP280Lib.js
brianjjones: whoops that was from when I was testing my script for stripping the export

samples/TempLCD/GroveLCDLib.js
brianjjones: @jimmy-huang  this is basically breaking up the current I2CBMP280.js, I'm reusing the code from it. I'd be more than happy to create a new Lib for the other monitor I have, but that will be at another time.

samples/TempLCD/BMP280Lib.js
```diff
@@ -0,0 +1,104 @@
...
+    var dig_T3;
+
+    function uint16toInt16(value) {
+        if ((value & 0x8000) > 0) {
+            value = value - 0x10000;
+        }
+        return value;
+    }
+
+    bmp280API.readCoefficients = function() {
```
brianjjones: They might want to, and it doesn't hurt anything (to my knowledge)

samples/TempLCD/GroveLCDLib.js
jimmy-huang: @grgustaf maybe what I pointed out is off-the-topic, I totally get what the code is trying to show about how to import 2 files, what I am saying that we could've chosen a much simpler module, and maybe that's what we should do, where if someone is looking at our example, they don't have to understand all these I2C code that doesn't totally make sense to read, and it just defeats the purpose.

brianjjones:
Closing this PR in favor of #423

PR #402, opened 2016-11-10
Mdef construction

grgustaf:


poussa:
Note that Zephyr unified kernel does not need mdef files any more.

jimmy-huang:
+1

PR #411, opened 2016-11-10
[build] Fixed convert.sh to work if uglify is not installed

jprestwo:

### DIFF
--- a/scripts/convert.sh
+++ b/scripts/convert.sh
@@ -77,9 +77,7 @@ do
         fi
     elif [ "$char" = $'\n' ] && [ $UGLIFY -eq 0 ]; then
         # Handle new lines if uglify is not installed
-        printf "\"" >> $OUTPUT
-        printf "\n" >> $OUTPUT
-        printf "\"" >> $OUTPUT
+        printf "\\\n\"\n\"" >> $OUTPUT
     else
         printf "$char" >> $OUTPUT
     fi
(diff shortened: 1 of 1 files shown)


scripts/convert.sh
```diff
@@ -73,6 +75,11 @@ do
         if [ "$char" = $'r' ] || [ "$char" = $'n' ]; then
             printf "\\\\$char" >> $OUTPUT
         fi
+    elif [ "$char" = $'\n' ] && [ $UGLIFY -eq 0 ]; then
+        # Handle new lines if uglify is not installed
+        printf "\"" >> $OUTPUT
+        printf "\n" >> $OUTPUT
+        printf "\"" >> $OUTPUT
```
grgustaf: Wouldn't printf "\"\n\"" >> $OUTPUT be the same as these three lines?

And what is this doing? I'm confused. :D

grgustaf: Better yet I think `printf '"\n"' >> $OUTPUT` does the same thing.

jprestwo: This is for lines that are just new lines. Its adding a " for the end of the previous line, then new line, then adding a " for the next line below. But I needed to do more testing obviously because some scripts don't work. They compile and run, but don't seem to work.

jprestwo:
@grgustaf, ok, I consolidated those 3 printf's into one and fixed it so it actually worked :)

grgustaf:
+1

PR #412, opened 2016-11-10
[ashell] Shrinking ring buffer size

brianjjones:
The memory dedicated to the ring buffer is too large.  It is
causing ashell to fail to build because it won't fit.

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -61,7 +61,7 @@ struct zjs_callback_map {
 static uint8_t args_buffer[ZJS_CALLBACK_BUF_SIZE];
 static struct zjs_port_ring_buf ring_buffer;
 #else
-SYS_RING_BUF_DECLARE_POW2(ring_buffer, 4);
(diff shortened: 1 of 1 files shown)


jimmy-huang:
@jprestwo can you take a look at this? Does this have any performance impact on 101?  I believe this should be changed dynamically at build time by passing it something like RING_BUFFER_SIZE, instead of hard-coding it.

jprestwo:
@jimmy-huang, a compile time variable is a fine idea. The value of that would still need to be determined though. I think to be safe we may want to increase this to 5. When its 4, that is just 16 bytes, so if there is a callback with multiple arguments it could fill up quickly. Really, the best way to figure this out is by testing. We should test GPIO buttons being pressed quickly, like using something similar to AutoButton.js, and see if we can fill up the ring buffer.

grgustaf:
For now, +1.

This kind of setting is too picky to go in a compile flag like that. But this and several other tunables should be configurable through something akin to prj.conf at some point.

PR #414, opened 2016-11-10
[build] fixed build break from missing ZJS_PRINT and QEMU socket code

jprestwo:


jprestwo:
Two issues here:
 - I had added Makefile code to allow the OCF bindings run under QEMU. The code worked at the time, but now it does not, so I am removing until a solution is found.
 - There was a missing PRINT that did not get changed to ZJS_PRINT when you turn on TRACE

jimmy-huang:
+1

grgustaf:
+1

PR #415, opened 2016-11-10
[ARC] Optimize ARC build using modules

jimmy-huang:
We seem to hit a seg fault at least with ADC read when ARC is
too big, we should be really only build ARC according to what
MODULES is needed to run the samples, this will save a ton of
space.  The downside is that now the user must build and flash
both the X86 and ARC with make all and dfu-all.


grgustaf:
LGTM

PR #416, opened 2016-11-10
[travis] Fix git check to use rev-list to be sure it finds a good commit

grgustaf:
The previous method using HEAD~49 was based on the checkout having gone
to a depth of 50, but under certain circumstances it could fail as it did
for the test build of e9b8279 apparently.


jimmy-huang:
+1

PR #417, opened 2016-11-11
[makefile] Make analyze/generate targets dependent on JS file existing

grgustaf:


jimmy-huang:
+1

PR #419, opened 2016-11-11
[cleanup] Remove old .gitignore file in deps

grgustaf:


jimmy-huang:
+1, i can't believe we still had these

PR #420, opened 2016-11-11
[makefile] Report build error when ZEPHYR_BASE is not set

grgustaf:
I really wanted to warn but set it to deps/zephyr by default, but
couldn't figure out how to do that. This is a good improvement
though since it will only report an error if you actually need
Zephyr.


jimmy-huang:
+1

PR #421, opened 2016-11-11
[callbacks] optimized callbacks

jprestwo:
- Removed the union of C/JS structs so callbacks only do 1 allocation
 - Changed callback ID's to int16_t's instead of int32_t's
 - Changed all modules using calbacks to use int16_t's

### DIFF
--- a/src/zjs_aio.c
+++ b/src/zjs_aio.c
@@ -19,7 +19,7 @@ static struct nano_sem aio_sem;
 
 typedef struct aio_handle {
     jerry_value_t pin_obj;
-    int32_t callback_id;
+    int16_t callback_id;
     double value;
     jerry_value_t jvalue;
 } aio_handle_t;
--- a/src/zjs_aio.c
+++ b/src/zjs_aio.c
@@ -19,7 +19,7 @@ static struct nano_sem aio_sem;
 
 typedef struct aio_handle {
     jerry_value_t pin_obj;
-    int16_t callback_id;
+    zjs_callback_id callback_id;
     double value;
     jerry_value_t jvalue;
 } aio_handle_t;
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -27,20 +27,28 @@
 
 #define INITIAL_CALLBACK_SIZE  16
 #define CB_CHUNK_SIZE          16
-
-#define CALLBACK_TYPE_JS    0
-#define CALLBACK_TYPE_C     1
-#define JS_TYPE_SINGLE      0
-#define JS_TYPE_LIST        1
-
 #define CB_LIST_MULTIPLIER  4
 
+// flag bit value for JS callback
+#define CALLBACK_TYPE_JS    0
+// flag bit value for C callback
+#define CALLBACK_TYPE_C     1
+// flag bit value for single JS callback
+#define JS_TYPE_SINGLE      0
+// flag bit value for list JS callback
+#define JS_TYPE_LIST        1
+// Bits in flags for once, type (C or JS), and JS type (single or list)
 #define ONCE_BIT    0
 #define TYPE_BIT    1
 #define JS_TYPE_BIT 2
-#define GET_ONCE(f)     (f & (1 << ONCE_BIT))
-#define GET_TYPE(f)     (f & (1 << TYPE_BIT))
-#define GET_JS_TYPE(f)  (f & (1 << JS_TYPE_BIT))
+// Macros to set the bits in flags
+#define SET_ONCE(f, b)      f |= (b << ONCE_BIT)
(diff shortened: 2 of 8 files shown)


src/zjs_aio.c
```diff
@@ -19,7 +19,7 @@ static struct nano_sem aio_sem;
 
 typedef struct aio_handle {
     jerry_value_t pin_obj;
-    int32_t callback_id;
+    int16_t callback_id;
```
pfalcon: Are you sure you save anything here? From jerryscript:
~~~
typedef uint32_t jerry_size_t;
~~~

pfalcon:
> Changed callback ID's to int16_t's instead of int32_t's

Isn't that called "arbitrary restriction"? Are you sure there never would be more than 64K callbacks in the queue? Think Linux port for example. Or bright MCU future with free MBs of RAM. At the very least, I'd suggest to make that a typedef.

src/zjs_aio.c
jprestwo: This doesn't have anything to do with jerryscript. The callback ID used to be stored as a int32_t, but there is not a need for that many callbacks, so we are re-evaluating and changing it to a int16_t to save 2 bytes.

jprestwo:
@pfalcon, I cant see any reason to need more than 64k callbacks, but there is no harm in changing the calllback ID to a typedef'ed type, that can be changed at compile time to be 16 or 32 bits.

src/zjs_aio.c
pfalcon: I see. A compiler would by default align structure fields using architecture's natural alignment, which for any 32-bit platform would be 4 bytes. Sticking 16-bit value between 32-bit and 64-bit value wouldn't save you anything. You'd need to group 2 16-bit values together to fill the slot.

You can ask compiler for a packed structure, then you save, but at the expense of the combination of the following: speed, code size, crashing due to alignment fault, or just getting invalid data.

src/zjs_callbacks.c
```diff
@@ -248,33 +235,33 @@ int32_t zjs_add_callback_list(jerry_value_t js_func,
...
-    new_cb->js->post = post;
-    new_cb->js->handle = handle;
-    new_cb->js->func_list = NULL;
-    new_cb->js->max_funcs = 0;
-    new_cb->js->num_funcs = 0;
-    new_cb->js->once = once;
+    // For a single callback, instead of allocating for just one function, we
+    // can use the pointer to hold the function since it will never grow
+    jerry_value_t func_val = jerry_acquire_value(js_func);
+    memcpy(&new_cb->func_list, &func_val, sizeof(jerry_value_t));
```
grgustaf: This is evil, but func_list is part of a union already so add a jerry_value_t to the union and just set it using that option.

(If I'm understanding right.)

jprestwo: oh, ya I hadn't thought of that. Didn't like my nasty hack? :) ill add its own value to the union, much better way of doing it

src/zjs_callbacks.c
```diff
@@ -288,47 +271,42 @@ int32_t add_callback(jerry_value_t js_func,
...
                 int i;
-                for (i = 0; i < cb_map[id]->js->num_funcs; ++i) {
-                    jerry_release_value(cb_map[id]->js->func_list[i]);
+                for (i = 0; i < cb_map[id]->num_funcs; ++i) {
+                    jerry_release_value(cb_map[id]->func_list[i]);
                 }
-                zjs_free(cb_map[id]->js->func_list);
-            } else {
-                jerry_release_value(cb_map[id]->js->js_func);
+                zjs_free(cb_map[id]->func_list);
```
grgustaf: This looks wrong, you only get here if this is NULL. Instead you should be checking some kind of flag to see if you have saved a single function instead of a func_list and release that.

src/zjs_callbacks.c
```diff
@@ -30,33 +30,33 @@
 
 #define CALLBACK_TYPE_JS    0
 #define CALLBACK_TYPE_C     1
+#define JS_TYPE_SINGLE      0
+#define JS_TYPE_LIST        1
```
grgustaf: You never seem to use this JS_TYPE_LIST, and you use the JS_TYPE_SINGLE once but don't explicitly set it anywhere, need to make more use of these I think to distinguish which value in the union you should be looking for.

jprestwo:
@grgustaf, quick question: should I change all the modules that use/save callback ID's to use the `zjs_callback_id` type instead of `int32_t`'s? Or should we do that with another commit? My only hesitation is that it would touch nearly every file, so it may lead to a time consuming merge conflict with https://github.com/01org/zephyr.js/pull/450 (only time consuming, but not hard). Either way is fine with me.

grgustaf:
LGTM

PR #423, opened 2016-11-11
[modules] Add support for Node.js modules

brianjjones:
Users can now import Javascript modules using
var myObj = require("modulename.js");

A sample is also included to demonstrate how this is done.

### DIFF
--- a/modules/BMP280.js
+++ b/modules/BMP280.js
@@ -1,0 +1,105 @@
+// Copyright (c) 2016, Intel Corporation.
+// JavaScript library for the BMP280 sensor
+
+function BMP280() {
+
+    // API object
+    var bmp280API = {};
+
+    var i2c = require("i2c");
+    bmp280API.i2cDevice = i2c.open({ bus: 0, speed: 100 });
+
+    var bmp280 = {
+        ADDRESS: 0x77,
+        NORMAL_MODE: 0x03,
+        REGISTER_DIG_T1: 0x88, // Unsigned Short
+        REGISTER_DIG_T2: 0x8A, // Signed Short
+        REGISTER_DIG_T3: 0x8C, // Signed Short
+        REGISTER_CHIPID: 0xD0,
+        REGISTER_CONTROL: 0xF4,
+        REGISTER_CONFIG: 0xF5,
+        REGISTER_TEMPDATA: 0xFA, // Signed 20bit. Always positive
+        STANDBY: 2 << 5,
+        FILTER: 0x00,
+        SPI_3W_DISABLE: 0x00,
+        PRESS_OVER: 1 << 2,
+        TEMP_OVER: 1 << 5
+    }
+
+    var dig_T1;
+    var dig_T2;
+    var dig_T3;
+
+    function uint16toInt16(value) {
+        if ((value & 0x8000) > 0) {
+            value = value - 0x10000;
+        }
+        return value;
+    }
+
+    bmp280API.readCoefficients = function() {
+        // These need to be read as a burst read in order to get accurate numbers,
+        // so dig_TAll contains dig_T1 through dig_T3
+        var dig_TAll = this.i2cDevice.burstRead(bmp280.ADDRESS, 24,
+                                                bmp280.REGISTER_DIG_T1);
+        dig_T1 = dig_TAll.readUInt16LE();
+        dig_T2 = dig_TAll.readUInt16LE(2);
+        dig_T3 = dig_TAll.readUInt16LE(4);
+
+        // Convert these to signed
+        dig_T1 = uint16toInt16(dig_T1);
+        dig_T2 = uint16toInt16(dig_T2);
+        dig_T3 = uint16toInt16(dig_T3);
+    }
+
+    bmp280API.readTemperature = function() {
+        // Read the raw temperature value.
+        var tempBuffer = this.i2cDevice.burstRead(bmp280.ADDRESS, 4,
+                                                  bmp280.REGISTER_TEMPDATA);
+        // Get the number from the buffer object
+        var tempDec = tempBuffer.readUInt32BE();
+
+        // Selects the first 20 bits of 32 bit integer and maintains the sign
+        tempDec >>>= 12;
+
+        // Convert raw temperature to Celsius using the algorithm found here
+        // https://www.bosch-sensortec.com/bst/products/all_products/bmp280
+        var var1 = (((tempDec >> 3) - (dig_T1 << 1)) * (dig_T2)) >> 11;
+        var var2 = (((((tempDec >> 4) - (dig_T1)) * ((tempDec >> 4) - (dig_T1))) >> 12) *
+                    (dig_T3)) >> 14;
+
+        var t_fine = var1 + var2;
+        var T = ((t_fine * 5 + 128) >> 8) / 100 ;
+
+        return T;
+    }
+
+    function init() {
+        var reader = bmp280API.i2cDevice.read(bmp280.ADDRESS, 1,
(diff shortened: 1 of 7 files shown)


src/zjs_modules.c
```diff
@@ -123,8 +123,40 @@ static jerry_value_t native_require_handler(const jerry_value_t function_obj,
             return mod->instance;
         }
     }
+    ZJS_PRINT("Native module not found, searching for JavaScript module %s\n", module);
```
jprestwo: May just want to change this to a DBG_PRINT(). We really need to go though and change all prints to DBG_PRINTS or ERR_PRINTS unless we have a good reason for always printing.

src/zjs_modules.c
```diff
@@ -123,8 +123,40 @@ static jerry_value_t native_require_handler(const jerry_value_t function_obj,
...
+    for (int i = 0; i < 4; i++) {
+        // Strip the ".js"
+        module[len-i] = '\0';
+    }
 
-    ZJS_PRINT("MODULE: `%s'\n", module);
+    jerry_value_t found_obj = zjs_get_property(exports_obj, module);
+
+    if (jerry_value_is_object(found_obj)) {
+        ZJS_PRINT("JavaScript module %s loaded\n", module);
```
jprestwo: Same as above, change to DBG_PRINT

src/zjs_modules.c
brianjjones: I generally agree with that, but this only shows up at build time - I think it should print no matter what so people can look at the build log to see what got included.  Thoughts?

brianjjones: We decided that since the native modules get printed during the build, the external JS ones should be too.  This only happens at build time so it won't add a bunch of noise to the output.

jprestwo:
This all LGTM. Not sure why the build is failing. I assume (at least on the Linux build), it has something to do with the concatenation to /tmp/zjs.js. And the ashell build just creates too big of a binary, which is semi-expected.

brianjjones:
It builds now, I put in a fix so that 'make linux' does things the old way rather than looking for the new /tmp/zjs.js file

modules/BMP280.js
```diff
@@ -0,0 +1,105 @@
...
+        }
+        return value;
+    }
+
+    bmp280API.readCoefficients = function() {
+        // These need to be read as a burst read in order to get accurate numbers,
+        // so dig_TAll contains dig_T1 through dig_T3
+        var dig_TAll = this.i2cDevice.burstRead(bmp280.ADDRESS, 24,
+                                                bmp280.REGISTER_DIG_T1);
+        dig_T1 = dig_TAll.readUInt16LE();
```
jimmy-huang: readUInt16LE not need to pass in argument here like offset?

modules/BMP280.js
```diff
@@ -0,0 +1,105 @@
...
+
+        return T;
+    }
+
+    function init() {
+        var reader = bmp280API.i2cDevice.read(bmp280.ADDRESS, 1,
+                                              bmp280.REGISTER_CHIPID);
+
+        // Check that we have a BMP280 on the I2C bus
+        if (reader.toString('hex') !== "58")
```
jimmy-huang: it might be nice to be consistent with single quote/double quote, but no big deal

modules/GroveLCD.js
```diff
@@ -0,0 +1,164 @@
...
+        // Write blank space to LCD
+        this.writeText(blanks);
+    }
+
+    groveLCDAPI.resetCursor = function (col, row) {
+        // Reset the cursor position
+        row = row || 0;
+        col = col || 0;
+
+        if (row == 0) {
```
jimmy-huang: use ===

modules/GroveLCD.js
```diff
@@ -0,0 +1,164 @@
...
+            // Put the cursor on the first row at col
+            col |= 0x80;
+        } else {
+            // Put the cursor on the second row at col
```
jimmy-huang: if row offset is not 0 or 1, shouldn't you throw an error?

modules/GroveLCD.js
```diff
@@ -0,0 +1,164 @@
...
+        if (colorObj.green && colorObj.green > -1 && colorObj.green < 256) {
+            this.i2cDevice.write(glcd.BACKLIGHT_ADDR,
+                                 new Buffer([glcd.REGISTER_G, green]));
+        }
+    }
+
+
+    groveLCDAPI.writeText = function(word, col, row) {
+
+        if (col !== undefined && row !== undefined) {
```
jimmy-huang: if col and row is not integer, or word is not string, should throw error

modules/GroveLCD.js
```diff
@@ -0,0 +1,164 @@
...
+        this.i2cDevice.write(glcd.DISPLAY_ADDR, wordBuffer);
+    }
+
+    function init() {
+
+        // Clear the screen
+        groveLCDAPI.clear();
+
+        // Set our preferences for the Grove LCD
+        groveLCDAPI.i2cDevice.write(glcd.DISPLAY_ADDR,
```
jimmy-huang: not sure you need to set up these initial display preferences here, since usually it's the app to decide at run time how they would like the display to be set up.

modules/BMP280.js
pfalcon: +1 from me for double-quotes as the default. Let's follow mother of all languages, C, it's around for almost 50 years so wouldn't let to go wrong with such a choice.

brianjjones: updated to be consistent

modules/GroveLCD.js
brianjjones: Sure I'll add that

modules/GroveLCD.js
brianjjones: On second thought, how about I just make it defaultInit? That way the user can choose to just use this setup rather than make their own.  They can also do it their own way as well.

jimmy-huang: It's really hard to decide what's really the "best" default set up is, one app might have 1 row of display, while the other app might use 2 rows, and one might wanna have the cursor, and the other one might not, so I rather leave it up to the app use to configure, instead of us forcing them an ideal setup.

brianjjones: I'm fine with that, just figured some of the user base wasn't going to be familiar with I2C so thought I'd give an example.

jimmy-huang: No I got your point, I am just saying that ot makes more sense to move them into the samples/I2CBMP280.js, not inside the module, that way they can still reference it as on how to set it up.

jimmy-huang: so I guess what you are missing in your api is what the equivalent of setDisplayState, and setFunction()

modules/BMP280.js
```diff
@@ -9,7 +9,7 @@ function BMP280() {
     var i2c = require("i2c");
     bmp280API.i2cDevice = i2c.open({ bus: 0, speed: 100 });
 
-    var bmp280 = {
+    bmp280API.bmp280Addrs = {
```
jimmy-huang: see my comments above, if you expose the setDisplayState and setFunction, then you don't need to expose the I2C calls.

jimmy-huang:
+1

kenchris:
So what was the stance on ES6 modules? Node will have them and most newer tooling around the web already uses them like WebPack2 and Splittable which was announced today. https://medium.com/@cramforce/introducing-splittable-1c882babca7e#.jemxn8n32

Maybe we should consider using a module bundler like Rollup https://github.com/rollup/rollup

It also supports good treeshaking which would be beneficial in our case.

PR #426, opened 2016-11-11
[build] Added VARIANT to the .last_build file

jprestwo:
- This means that building in release mode THEN in debug mode will
   cause the debug build to clean before rebuilding. This already
   worked for Zephyr targets, but not on Linux


jimmy-huang:
+1

PR #425, opened 2016-11-11
[makefile] Add -Werror to Linux builds

grgustaf:
Also, fix warning now caught in performance module.

Signed-off-by: Geoff Gustafson <geoff@linux.intel.com>

PR #428, opened 2016-11-12
Compress ints

grgustaf:
This should be rebased after #425 goes in.

I'm not really sure if this is a good idea, but it was a fun one. Hopefully if we don't use it, it won't actually take up any space.


jimmy-huang:
I don't know who's going to use this, but it's free and it has a unit test, so +1 :)

PR #429, opened 2016-11-12
[general] Print ERR_PRINT messages in release builds

grgustaf:
We don't want to miss errors when we're in release builds. Things for
debug only belong in DBG_PRINT.


jimmy-huang:
+1

PR #431, opened 2016-11-12
[callbacks] Return this from some methods to support chaining

brianjjones:
Events now return this to allow chaining.

This is a re-based and slightly modified version of #340 by Gabriel Schulhof


jimmy-huang:
+1

PR #437, opened 2016-11-14
[Makefile] Don't force BOARD=qemu_x86 for "qemu".

pfalcon:
This allows to build QEMU target for BOARD=qemu_cortex_m3, and in future,
for other QEMU target. This however now requires explicit BOARD when
build for x86:

make BOARD=qemu_x86 qemu


grgustaf:
+1

Issue #368, opened 2016-11-04
RFC: Making GPIO access board neutral


pfalcon:
Currently, Zephyr.js officially supports only 2 boards: arduino_101 and frdm_k64f. Of course, people will want to run it on other boards, and for example I'm already being approached with questions "how to run it on ARM board X?". So far I could answer only "Just try to build it for your board and report any issues upstream". I finally took a look at what would it take to support another board, specifically for GPIO module, and see that the issues are many. Current zjs_gpio.c implementation has wild hard-coding's like:

~~~
#ifdef CONFIG_BOARD_FRDM_K64F
#define GPIO_DEV_COUNT 5
#else
#define GPIO_DEV_COUNT 1
#endif
~~~

So any board but frdm_k64f would get support for just 1 GPIO port. 

Note that support each board one by one is never going to scale. There's only 2 supported now and there're incompatibilities/inconsistencies already: https://github.com/01org/zephyr.js/issues/284 . So, I'd like to propose the following straightforward solution:

* Zephyr.js should be able to support GPIO access for any Zephyr-supported board in a generic manner.
* However, if some party desires to add conveniences for a particular board, like currently done for arduino_101 and frdm_k64f, they should be able to (in particular, all current infrastructure for arduino_101 and frdm_k64f should remain compatible with a proposed additions).

And the way to support generic GPIO access for any Zephyr board is to just use Zephyr's way of identifying GPIO ports and pins, without trying to add any abstractions on top of it. So, Zephyr uses model where there're explicit GPIO ports, identified by strings (effectively device names) and pin numbers within port. This model is of course based on the actual reality of it - (almost?) every MCU vendor splits GPIOs into ports with relative numbers of pins within each port, so GPIO "coordinates" are "two-dimensional".

The specific proposal is:

~~Add "port" property to GPIOInit structure.~~ (reconsidered)

Update `dictionary GPIOInit` in https://github.com/01org/zephyr.js/blob/master/docs/gpio.md to:

~~~
dictionary GPIOInit {
-    unsigned long pin;
+    union {unsigned long, string} pin;
    boolean activeLow = false;
    string direction = "out";  // in, out
    string edge = "any";       // none, rising, falling, any
    string pull = "none";      // none, up, down
};
~~~

If a `pin` is a string, a new model is used, and the string format is `"GPIO_1.21"`, where part before the dot is Zephyr GPIO port device name, after the dot - pin number within this port. 

[...]

PR #438, opened 2016-11-14
[GPIO] Allow to work with Zephyr-native GPIO port and pins.

pfalcon:
For this allow "pin" property passed to open() to take a string value
like "GPIO_1.21". Part before the dot is Zphyr native GPIO port device
name, after the dot - ping number within that port. This notation allows
to access GPIO of any board supported by Zephyr kernel, without any
additional patching and modules on Zephyr.js (additional convenience
modules *can* be added for more boards, the main point is that they
are now optional (later perhaps implements in JS code)).

### DIFF
--- a/src/zjs_gpio.c
+++ b/src/zjs_gpio.c
@@ -5,6 +5,7 @@
 #include <gpio.h>
 #include <misc/util.h>
 #include <string.h>
+#include <stdlib.h>
 
 // ZJS includes
 #include "zjs_gpio.h"
@@ -39,7 +40,7 @@ void (*zjs_gpio_convert_pin)(uint32_t orig, int *dev, int *pin) =
 struct gpio_handle {
     struct gpio_callback callback;  // Callback structure for zephyr
     uint32_t pin;                   // Pin associated with this handle
-    uint32_t devnum;
+    struct device *port;            // Pin's port
     uint32_t value;                 // Value of the pin
     zjs_callback_id callbackId;             // ID for the C callback
     jerry_value_t pin_obj;          // Pin object returned from open()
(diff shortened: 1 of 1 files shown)


pfalcon:
This patch implements ideas in https://github.com/01org/zephyr.js/issues/368

src/zjs_gpio.c
```diff
@@ -325,12 +379,12 @@ static jerry_value_t zjs_gpio_open(const jerry_value_t function_obj,
         // Zephyr ISR callback init
```
daniel-thompson: I'm interested whether handles can be made unconditional so lookup_pin() can be inlined into the open function and all other usages grab the struct *device from the handle.

On a similar vein, it is possible to replace the attributes of the pinobj with "getter" methods? The current approach looks deeply inefficient in terms of RAM usage. If the handle were unconditional then getter methods could integate it and RAM usage should be smaller (not to mention that for bi-directional bit banging we need the "direction" to be hooked up to a setter method anyway).

grgustaf: @daniel-thompson, yes it seems more sensible to just always create a handle. Also, thanks for pointing out the getter/setter issue. Yes I think it is possible and I will look into moving in that direction. Similiarly, the functions above should also be coming from a prototype object instead of duplicated for each object. I submitted the first patch making that kind of change as PR #471.

I don't think it has to hold up this patch currently though.

src/zjs_gpio.c
```diff
@@ -49,6 +50,52 @@ struct gpio_handle {
     uint32_t last;
 };
 
+static jerry_value_t lookup_pin(const jerry_value_t pin_obj, struct device **port, int *pin)
+{
+    char pin_id[12];
```
grgustaf: So maybe bump this to 32? That ought to be pretty safe.

pfalcon: @daniel-thompson : Can you give an example of GPIO device name longer than 8 chars?

@grgustaf : I'll make it 32 as you suggest.

src/zjs_gpio.c
pfalcon: @daniel-thompson : I of course noticed matters which you talk about, but they definitely aren't related to the topic of this patch. This patch tries to support GPIOs of any Zephyr board. You're talking about optimizing how the module works. That's definitely useful to consider separately, though I'd call a "premature optimization" cliche (as there're talks about switching to completely different GPIO, etc. APIs).

src/zjs_gpio.c
pfalcon: > So maybe bump this to 32?

Done.

src/zjs_gpio.c
```diff
@@ -49,6 +50,52 @@ struct gpio_handle {
...
+        // this is usually GPIO_0, GPIO_1, etc., but some Zephyr
+        // ports have completely different naming, so don't assume
+        // anything! "num" is a numeric pin number within the port,
+        // usually within range 0-31.
+        char *pinstr = strrchr(pin_id, '.');
+        bool error = true;
+        if (pinstr && pinstr[1] != 0) {
+            *pinstr = 0;
+            newpin = strtol(pinstr + 1, &pinstr, 10);
+            if (*pinstr == 0)
```
pfalcon: Done.

pfalcon:
@grgustaf : Thanks for your review. I've updated the patch based on your comments. I really appreciate looking into this, this patch is pretty critical for us, without it, we simply can't try Z.js on any other boards besides a101/frdm, and there's big interest to try it on more boards, and to work towards better support it on more boards.

grgustaf:
+1, merged

daniel-thompson:
On 29/11/16 16:55, Paul Sokolovsky wrote:
> @daniel-thompson <https://github.com/daniel-thompson> : I of course
> noticed matters which you talk about, but they definitely aren't related
> to the topic of this patch. This patch tries to support GPIOs of any
> Zephyr board. You're talking about optimizing how the module works.
> That's definitely useful to consider separately, though I'd call a
> "premature optimization" cliche (as there're talks about switching to
> completely different GPIO, etc. APIs).

To a degree...

I was trying to figure out why the lookup function ended up being called 
so often that you decided that it needed to use static allocation for 
the strings. I concluded *that* was because there was no context 
structure to cache the result.

BTW I used git grep '\".*GPIO' to hunt down the >8 character strings.

pfalcon:
@daniel-thompson : Sorry for the delay looking into this issue.

> BTW I used git grep '\".*GPIO' to hunt down the >8 character strings.

I see, you mean things like:

~~~
arch/arc/soc/em9d/Kconfig.defconfig:	default "GPIO_PORTA"
~~~

> I was trying to figure out why the lookup function ended up being called
so often that you decided that it needed to use static allocation for
the strings.

Well, dynamic allocation is anti-pattern in deeply embedded systems, so nope, I didn't to do because it's called often, but because a stack allocation is well fitting there (as size-bound and small). I thought about stack usage though, that's why I set it to a conservative size. As I mentioned, I agree with you about caching of handle (unconditionally). I'd even say those were good reasons to write it like that from the start. But if it wasn't done like that, patching it just because I worked with that code didn't sound right either (changing too much at once is a known cause of regressions).

Thanks for your review and helpful hints! (It's on my TODO to bring up matter of (in)consistency of device names on Zephyr mailing list.)

PR #442, opened 2016-11-14
[travis] Reduce VMs used to two in an effort to speed up builds

grgustaf:


jimmy-huang:
+1

PR #445, opened 2016-11-15
[Tests] Update test-pwm.js for bug fix

cuiyanx:

### DIFF
--- a/tests/test-pwm.js
+++ b/tests/test-pwm.js
@@ -33,15 +33,15 @@ var gpio = require("gpio");
 var pins = require("arduino101_pins");
 var pinA, pinB, msTimer, cycleTimer;
 
-pinB = gpio.open({pin: pins.IO2, direction: 'in'});
+pinB = gpio.open({ pin: pins.IO2, direction: 'in' });
 
 // PWMPins open
-pinA = pwm.open({channel: pins.IO3});
+pinA = pwm.open({ channel: pins.IO3 });
 assert(pinA != null && typeof pinA == "object",
       "open: defined pin and default argument");
 
 expectThrow("open: undefined pin", function () {
-    pinA = pwm.open({channel: 1024});
+    pinA = pwm.open({ channel: 1024 });
 });
 
 // set Period and PulseWidth with ms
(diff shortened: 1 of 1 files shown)


tests/test-pwm.js
```diff
@@ -54,7 +54,7 @@ expectThrow("pwmpin: set pulseWidth without period", function () {
     pinA.setPulseWidth(300);
 });
 
-pinA = pwm.open({channel: pins.IO3, period: 3, pulseWidth: 1});
+pinA = pwm.open({ channel: pins.IO3, period: 3, pulseWidth: 1 });
 assert(pinA != null && typeof pinA == "object",
```
jimmy-huang: use !== and ===, this is not related to this patch, but you might want to fix these here.

tests/test-pwm.js
```diff
@@ -76,7 +76,7 @@ msTimer = setInterval(function () {
     msCount = msCount + 1;
 }, 50);
 
-setTimeout(function() {
+setTimeout(function () {
     assert(msTrue == 6 && msFalse == 14 && msCount == 20,
```
jimmy-huang: same here

tests/test-pwm.js
```diff
@@ -106,7 +106,7 @@ setTimeout(function() {
     cycleTimer = setInterval(function () {
        Flag = pinB.read();
 
-       if(Flag == oldFlag){
+       if (Flag == oldFlag) {
```
jimmy-huang: same here

tests/test-pwm.js
```diff
@@ -132,7 +132,7 @@ setTimeout(function() {
            }
        }
     }, 10);
-}, 1000);
+}, 1040);
```
jimmy-huang: why is 1040 here? can you explain?

cuiyanx: The msTimer is detection pulse, so we must Keep enough time to check it.
       

>  1000=50X20 < total-time < 50X21=1050

I have got 1040 for robustness of this case.

jimmy-huang: ok, good to know, thanks.

PR #453, opened 2016-11-16
[Sensor] Added support for trigger mode

jimmy-huang:
Added trigger support, since currently in Zephyr, triggers can
only be signaledfrom interrupt from x86 side, all the x86 side
code is wrapped with BUILD_MODULE_SENSOR_TRIGGER #ifdef so that
it can be turn off later when Zephyr changes.

The trigger are set to sample in 50Hz, any higher frequency
will cause the ring buffer to start dropping change events.

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -563,7 +563,7 @@ static void process_gyro_data(struct device *dev)
     double dval[3];
 
     if (sensor_channel_get(dev, SENSOR_CHAN_GYRO_ANY, val) < 0) {
-        ZJS_PRINT("failed to read gyroscope channels.\n");
+        ZJS_PRINT("failed to read gyroscope channels\n");
         return;
     }
 
@@ -587,7 +587,7 @@ static void process_gyro_data(struct device *dev)
     sensor_value_snprintf(buf_x, sizeof(buf_x), &val[0]);
     sensor_value_snprintf(buf_y, sizeof(buf_y), &val[1]);
     sensor_value_snprintf(buf_z, sizeof(buf_z), &val[2]);
-    ZJS_PRINT("sending gyro : X=%s, Y=%s, Z=%s\n", buf_x, buf_y, buf_z);
+    ZJS_PRINT("sending gyro: X=%s, Y=%s, Z=%s\n", buf_x, buf_y, buf_z);
 #endif
 }
 
@@ -781,7 +781,7 @@ static void handle_sensor(struct zjs_ipm_message* msg)
                 error_code = ERROR_IPM_OPERATION_FAILED;
                 ZJS_PRINT("failed to initialize BMI160 sensor\n");
             } else {
-                if (auto_calibration(bmi160) != 0) {
+                if (auto_calibration(bmi160)) {
                     ZJS_PRINT("failed to perform auto calibration\n");
                 }
                 DBG_PRINT("BMI160 sensor initialized\n");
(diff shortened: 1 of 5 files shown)


arc/src/main.c
```diff
@@ -574,7 +563,7 @@ static void process_gyro_data(struct device *dev)
     double dval[3];
 
     if (sensor_channel_get(dev, SENSOR_CHAN_GYRO_ANY, val) < 0) {
-        ZJS_PRINT("Cannot read gyroscope channels.\n");
+        ZJS_PRINT("failed to read gyroscope channels.\n");
```
grgustaf: remove period for consistency i guess?

arc/src/main.c
```diff
@@ -598,20 +587,26 @@ static void process_gyro_data(struct device *dev)
     sensor_value_snprintf(buf_x, sizeof(buf_x), &val[0]);
     sensor_value_snprintf(buf_y, sizeof(buf_y), &val[1]);
     sensor_value_snprintf(buf_z, sizeof(buf_z), &val[2]);
-    ZJS_PRINT("Sending gyro : X=%s, Y=%s, Z=%s\n", buf_x, buf_y, buf_z);
+    ZJS_PRINT("sending gyro : X=%s, Y=%s, Z=%s\n", buf_x, buf_y, buf_z);
```
grgustaf: let's remove the space before the : like good Americans

arc/src/main.c
```diff
@@ -667,30 +781,36 @@ static void handle_sensor(struct zjs_ipm_message* msg)
                 error_code = ERROR_IPM_OPERATION_FAILED;
                 ZJS_PRINT("failed to initialize BMI160 sensor\n");
             } else {
-                if (!auto_calibration(bmi160)) {
+                if (auto_calibration(bmi160) != 0) {
```
grgustaf: You don't really need the != 0, but that's just my personal preferred style, not going to enforce it. :)

scripts/analyze.sh
```diff
@@ -187,18 +187,29 @@ if [ $? -eq 0 ] && [[ $MODULE != *"BUILD_MODULE_BUFFER"* ]]; then
...
     echo "CONFIG_SENSOR=y" >> arc/prj.conf.tmp
+    echo "CONFIG_GPIO=y" >> prj.conf.tmp
+    echo "CONFIG_GPIO_QMSI=y" >> prj.conf.tmp
+    echo "CONFIG_GPIO_QMSI_0_PRI=2" >> prj.conf.tmp
+    echo "CONFIG_GPIO_QMSI_1=y" >> prj.conf.tmp
+    echo "CONFIG_GPIO_QMSI_1_NAME=\"GPIO_1\"" >> prj.conf.tmp
+    echo "CONFIG_GPIO_QMSI_1_PRI=2" >> prj.conf.tmp
+    echo "CONFIG_SPI=y" >> arc/prj.conf.tmp
+    echo "CONFIG_SPI_DW_ARC_AUX_REGS=y" >> arc/prj.conf.tmp
+    echo "CONFIG_SPI_DW_INTERRUPT_SEPARATED_LINES=y" >> arc/prj.conf.tmp
```
grgustaf: This GPIO and SPI here is specific to the Arduino 101 hardware, we'd eventually need to make that dependent on architecture.

jprestwo: Do the same thing I did for UART:
```
if [ $BOARD = "arduino_101" ]; then
```
This is precisely the reason I changed the script to pass in BOARD from the Makefile.

jimmy-huang: ok, sounds good

arc/src/main.c
jimmy-huang: ok

grgustaf:
+1

PR #459, opened 2016-11-18
[tests] Sync tests from master

haoxli:
These tests should be tested in zjs-0.1


jimmy-huang:
+1

grgustaf:
It's been suggested that we call zjs-0.1 good as of this week. So let's get this in before QA does what might be their final pass on zjs-0.1.

PR #462, opened 2016-11-18
[build] Adding configuration file support for the build

brianjjones:
Now make can be given CONFIG=zjs.conf (or whatever file you choose)
to indicate you want certain modules to be built in.  This allows
you to build modules that aren't included in your Javascript file.
This will also allow us to build ashell, linux, etc with only the
modules we want rather than all of them.  In the future we can
remove the #ifdef BUILD_MODULE* from the src files as they simply
won't be included in the build rather than commented out.

### DIFF
--- a/fragments/zjs.conf.dev
+++ b/fragments/zjs.conf.dev
@@ -1,17 +1,17 @@
-# To include a module remove the # in front of it
+# To force inclusion of a module remove the # in front of it
 
+ZJS_AIO=y
+ZJS_ARDUINO101_PINS=y
+ZJS_BLE=y
+ZJS_BUFFER=y
+ZJS_CONSOLE=y
 ZJS_EVENTS=y
-#ZJS_OCF=y
 ZJS_GPIO=y
+#ZJS_GROVE_LCD=y
+ZJS_I2C=y
+#ZJS_OCF=y
 #ZJS_PERFORMANCE=y
 ZJS_PWM=y
-ZJS_UART=y
-ZJS_BLE=y
-ZJS_AIO=y
-ZJS_I2C=y
-#ZJS_GROVE_LCD=y
-ZJS_ARDUINO101_PINS=y
-ZJS_TIMERS=y
-ZJS_BUFFER=y
 #ZJS_SENSOR=y
-ZJS_CONSOLE=y
+ZJS_TIMERS=y
+ZJS_UART=y
--- a/fragments/zjs.conf.dev
+++ b/fragments/zjs.conf.dev
@@ -1,0 +1,17 @@
+# To include a module remove the # in front of it
+
+ZJS_EVENTS=y
+#ZJS_OCF=y
+ZJS_GPIO=y
+#ZJS_PERFORMANCE=y
+ZJS_PWM=y
+ZJS_UART=y
+ZJS_BLE=y
+ZJS_AIO=y
+ZJS_I2C=y
+#ZJS_GROVE_LCD=y
+ZJS_ARDUINO101_PINS=y
+ZJS_TIMERS=y
+ZJS_BUFFER=y
+#ZJS_SENSOR=y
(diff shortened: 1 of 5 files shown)


fragments/zjs.conf.dev
```diff
@@ -0,0 +1,17 @@
...
+ZJS_UART=y
+ZJS_BLE=y
+ZJS_AIO=y
+ZJS_I2C=y
+#ZJS_GROVE_LCD=y
+ZJS_ARDUINO101_PINS=y
+ZJS_TIMERS=y
+ZJS_BUFFER=y
+#ZJS_SENSOR=y
+ZJS_CONSOLE=y
```
grgustaf: Alphabetize these to make it easy to search for one and detect whether one is already in the list.

scripts/analyze.sh
```diff
@@ -53,15 +61,31 @@ function check_for_require()
     return $?
 }
 
+function check_config_file()
+{
+    # effects: checks $1 for uncommented module to inclue from #CONFIG
```
grgustaf: typo: include

scripts/analyze.sh
```diff
@@ -53,15 +61,31 @@ function check_for_require()
...
+    #          If found, it will be put in zjs.conf.tmp and appended to src/Makefile
```
grgustaf: appended to -> inserted into
if I understand you right in IRC

scripts/analyze.sh
```diff
@@ -53,15 +61,31 @@ function check_for_require()
...
+
+    if [ -n "$CONFIG" ]; then
+        rval=$(grep -v '^#' < $CONFIG | grep $1='y\|m' | xargs)
+        if [ "$rval" ]; then
+            return 0
+        else
+            return 1
+        fi
```
grgustaf: A few tricks to know here:
* grep like most utilties will return 0 when it's happy and >= 1 if it's not.
* $? gives you the return value of the last command run
* So you could skip capturing rval and just return $?
* Also, the < is unnecessary I think.

For posterity, Brian found that piping to xargs like this strips newlines. I suggest `tr -d '\n'` is a more decipherable way to do that, though. But anyway I don't think it's needed here. If you just run the grep | grep command then return $? I believe you get what you want.

zjs.conf
```diff
@@ -0,0 +1,17 @@
...
+#ZJS_UART=y
+#ZJS_BLE=y
+#ZJS_AIO=y
+#ZJS_I2C=y
+#ZJS_GROVE_LCD=y
+#ZJS_ARDUINO101_PINS=y
+#ZJS_TIMERS=y
+#ZJS_BUFFER=y
+#ZJS_SENSOR=y
+#ZJS_CONSOLE=y
```
grgustaf: Again, alphabetize.

zjs.conf
```diff
@@ -0,0 +1,17 @@
+# To include a module remove the # in front of it
```
grgustaf: How about "To force inclusion of a module, remove..."

src/Makefile.base
```diff
@@ -26,31 +26,32 @@ ccflags-y += -DZJS_PRINT_CALLBACK_STATS
...
 
+obj-$(ZJS_EVENTS) += zjs_event.o
+obj-$(ZJS_GPIO) += zjs_gpio.o
+obj-$(ZJS_BLE) += zjs_ble.o
+obj-$(ZJS_PWM) += zjs_pwm.o
+obj-$(ZJS_PERFORMANCE) += zjs_performance.o
+obj-$(ZJS_UART) += zjs_uart.o
+obj-$(ZJS_OCF) += zjs_ocf_client.o \
+                  zjs_ocf_server.o \
+                  zjs_ocf_common.o \
```
grgustaf: Drop this line's backslash?

brianjjones:
All those sound good - adding the changes for review

PR #466, opened 2016-11-22
[build] Added flag to toggle print float flag

jprestwo:
- To print floats on the A101, a flag had to be used which added
   ~11k to our ROM footprint. This adds a compile time option to
   enable this flag. It is disabled by default so if a float is
   printed it will approximate the value e.g.

   [Float ~42]

   instead of (with the flag enabled):

   42.424242

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,9 @@ TRACE ?= off
 MALLOC ?= pool
 # Print callback statistics during runtime
 CB_STATS ?= off
+# Print floats (uses -u _printf_float flag). This is a workaround on the A101
+# otherwise floats will not print correctly. It does use ~11k extra ROM though
+PRINT_FLOAT ?= off
 # Make target (linux or zephyr)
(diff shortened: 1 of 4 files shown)


grgustaf:
+1

Makefile
```diff
@@ -37,7 +40,12 @@ endif
 # Build for zephyr, default target
 .PHONY: zephyr
 zephyr: $(PRE_ACTION) analyze generate
-	@make -f Makefile.zephyr BOARD=$(BOARD) KERNEL=$(KERNEL) VARIANT=$(VARIANT) MEM_STATS=$(MEM_STATS) CB_STATS=$(CB_STATS)
+	@make -f Makefile.zephyr	BOARD=$(BOARD) \
+								KERNEL=$(KERNEL) \
```
jimmy-huang: align this?

jimmy-huang:
+1

Issue #434, opened 2016-11-14
[REG][I2C.js] No background color in GroveLCD screen


xiaolongx:
In commit 916cb71, flashing the [I2C.js](https://github.com/01org/zephyr.js/blob/master/samples/I2C.js) with below commands, no background color in GroveLCD screen, only "HELLO" and "WORLD" displays.
```
$ make all JS=samples/I2C.js
$ make dfu-all
```
If flashing [GroveLCD.js](https://github.com/01org/zephyr.js/blob/master/samples/GroveLCD.js) firstly, and then flash [I2C.js](https://github.com/01org/zephyr.js/blob/master/samples/I2C.js), it only display background without "HELLO" and "WORLD" in GroveLCD screen.

Sample | Commit | Result
---|---|---
I2C.js | 2ff271b | PASS 
I2C.js | 916cb71 | FAIL

brianjjones:
It turns out there were some registers that needed to be set for it to turn on.  And it turns out the Grove_LCD module has been doing this all along.  So when @jimmy-huang changed the arc build to only include the specified modules, the I2C demo broke because it didn't include the Grove_LCD module.  Basically it was always broken I just didn't know it :)  I'll fix it and push a PR.

brianjjones:
The fix for this is now in master

grgustaf:
@xiaolongx please verify

xiaolongx:
This issue cannot reproduce with commit a162db1 on both Arduino_101 and FRDM_K64F, close

PR #468, opened 2016-11-22
[bugfix] Fixes issue with I2C demos not showing backlight

brianjjones:
Fix for issue #434

### DIFF
--- a/samples/I2C.js
+++ b/samples/I2C.js
@@ -136,6 +136,7 @@ function writeWord() {
     }
 
     hello = !hello;
+    resetCursor();
     i2cDevice.write(GROVE_LCD_DISPLAY_ADDR, wordBuffer);
 }
 
--- a/samples/I2C.js
+++ b/samples/I2C.js
@@ -120,7 +120,7 @@ function changeRGB(red, green, blue) {
 }
 
 function writeWord() {
-    resetCursor();
+    clear();
     var wordBuffer = new Buffer(7);
         // The first byte in the buffer is for the register address,
         // which is where the word is going to be written to.
--- a/samples/I2C.js
+++ b/samples/I2C.js
@@ -56,55 +56,54 @@ var GLCD_DS_BLINK_OFF     = 0 << 0;
  *  RGB FUNCTIONS
  *******************************************/
 
-var REGISTER_POWER  = 0x08
-var REGISTER_R      = 0x04
-var REGISTER_G      = 0x03
-var REGISTER_B      = 0x02
+var REGISTER_POWER  = 0x08;
+var REGISTER_R      = 0x04;
+var REGISTER_G      = 0x03;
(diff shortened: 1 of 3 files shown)


samples/I2C.js
```diff
@@ -56,55 +56,54 @@ var GLCD_DS_BLINK_OFF     = 0 << 0;
...
+                    new Buffer([REG_MODE1, 0x00]));
+
+    i2cDevice.write(GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([REG_MODE2, 0x05]));
+
+    i2cDevice.write(GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([REGISTER_POWER, 0xFF]));
 }
 
 function resetCursor() {
```
grgustaf: You no longer use this function anywhere. But probably it has less flicker if you do use it, rather than clear the whole screen? Were you trying to avoid the double WORLD bug?

samples/I2C.js
```diff
@@ -56,55 +56,54 @@ var GLCD_DS_BLINK_OFF     = 0 << 0;
...
+
+    // Init the backlight
+    i2cDevice.write(GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([REG_MODE1, 0x00]));
+
+    i2cDevice.write(GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([REG_MODE2, 0x05]));
+
+    i2cDevice.write(GROVE_RGB_BACKLIGHT_ADDR,
+                    new Buffer([REGISTER_POWER, 0xFF]));
```
grgustaf: This bit is the bug fix I guess? Can you explain?

samples/I2C.js
brianjjones: I didn't mean to remove it I'll add it back in.

samples/I2C.js
brianjjones: Its the code needed to initialize and turn on the backlight.  I unfortunately don't know what the significance of the numbers are as I can't find a spec sheet for it.  This code is taken from the Grove LCD library and slightly altered.

samples/I2C.js
```diff
@@ -121,27 +120,29 @@ function changeRGB(red, green, blue) {
 }
 
 function writeWord() {
-    resetCursor();
-
+    clear();
```
grgustaf: I'm hoping you don't need to clear the whole screen here? Can you move resetCursor() from below back to here and not have to change this line?

I just expect it will be slow and flicker more, i.e. bad practice for an example.

brianjjones: This is the only way I currently can clear extra text that could appear as a result of I2C write failing.  In the future I'll need to create a way to either send the error to JS or make the I2C code smarter and have it try again if the write fails.

jimmy-huang:
+1

PR #470, opened 2016-11-23
[console] Support printing arrays, functions, null, undefined

grgustaf:
Also, improve handling of spaces between arguments, and improve Console.js
test to be more thorough and clean up output.

### DIFF
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -18,7 +18,7 @@
 #define IS_INT    1
 #define IS_UINT   2
 
-int is_int(jerry_value_t val) {
+static int is_int(jerry_value_t val) {
     int ret = 0;
     double n = jerry_get_number_value(val);
     ret = (n - (int) n == 0);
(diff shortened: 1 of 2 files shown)


grgustaf:
@jprestwo you should probably review this

src/zjs_console.c
```diff
@@ -34,48 +34,99 @@ int is_int(jerry_value_t val) {
...
+        }
+    }
+    else if (jerry_value_is_null(value)) {
+        fprintf(out, "null");
+    }
+    // NOTE: important that checks for function and array were above this
+    else if (jerry_value_is_object(value)) {
+        fprintf(out, "[Object]");
+    }
+    else if (jerry_value_is_string(value)) {
```
grgustaf: @jprestwo, sorry to make a style change on you here, but I really think the else if's on their own line is much more readable in a case like this. The near-blank lines break up the page so you can quickly find where each different type is addressed.

jprestwo: Do you want me to fix this? I am confused... Isn't this your PR? Or did you mean to comment on my other floating point printing PR (https://github.com/01org/zephyr.js/pull/466).

Or were you just commenting saying that you were changing the style of what I originally wrote. Changing it is fine with me. I agree it makes the code more readable, I am just very used to doing it the way I did it out of habit.
Just making sure, do we want ALL if-else blocks like this? or only one line blocks to break it up?

grgustaf: Just meant I was changing the style here. I don't think it needs to be a policy.

jimmy-huang:
+1

PR #471, opened 2016-11-24
[buffer] Use prototype inheritance to implement Buffer functions

grgustaf:
This should save lots of runtime JS memory when there are multiple
buffers instantiated because only the single prototype needs to contain
the mappings to native API functions. I'd like to measure how much this
saves but expect we should replicate it across all APIs.

Also, added a zjs_obj_add_functions (plural) API to simplify creating a
large batch of native API mappings for an object.

Also, reorder cleanup for ashell so that jerry_cleanup() is the final
step and should leave the JS memory in its pristine state.

### DIFF
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -150,9 +150,10 @@ void javascript_stop()
     parsed_code = 0;
 
     /* Cleanup engine */
-    jerry_cleanup();
-    zjs_ipm_free_callbacks();
     zjs_timers_cleanup();
+    zjs_ipm_free_callbacks();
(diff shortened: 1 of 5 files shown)


jprestwo:
This looks good. I am very curios of the RAM savings. Just to be clear, before, every object was storing all those functions/properties and now there will just be one object prototype and when you initialize a new object it will just point to the prototype, right?

Other modules should use this, though I think buffers will benefit the most. e.g. UART, which create a new buffer for every character that it reads (inefficient, but Node says to use buffers :) )

jimmy-huang:
+1

PR #473, opened 2016-11-24
[docs] Update README.md

cuiyanx:

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ $ sudo usermod -a -G plugdev USERNAME
 ```
 
 #### Add udev rules
-Copy these two files into your /etc/udev.rules directory:
+Copy these two files into your /etc/udev/rules.d directory (/etc/udev.rules for Ubuntu 14.04):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,7 +68,7 @@ $ sudo usermod -a -G plugdev USERNAME
 ```
 
 #### Add udev rules
-Copy these two files into your /etc/udev.rules directory:
+Copy these two files into your /etc/udev/rules.d directory:
```
kenchris: I have both directories on ubuntu, do you know the difference?

cuiyanx: If setup, there are only one director on Ubuntu 16.04 host; If update, you have both.

grgustaf: How about change like this then:

... /etc/udev/rules.d directory (/etc/udev.rules for Ubuntu 14.04):

cuiyanx: Great！

jimmy-huang:
+1

PR #483, opened 2016-11-28
[callbacks] Fixed broken callbacks after optimization change

jprestwo:
- Also fixed debug build relating to the callback change


grgustaf:
Important regression fix, +1

PR #484, opened 2016-11-28
[samples] Fix Spaceship2.js timer bug

grgustaf:
Stop using the event.value because it's unreliable because of button
bounce issues. Also, double edge interrupts are better used when the
same action will be triggered either way. Here a better approach to
make sure the lasers get turned off eventually is to check the button
status repeatedly in a timer. This makes it more like Spaceship.js
but at least the lasers will turn on immediately, rather than syncing
to 250ms timer, which is important when you're leading a TIE fighter.

### DIFF
--- a/samples/arduino/starterkit/Spaceship2.js
+++ b/samples/arduino/starterkit/Spaceship2.js
@@ -18,54 +18,43 @@ var gpio = require('gpio');
 var pins = require('arduino101_pins');
 
 // set up the GPIO pins
-var button = gpio.open({
-    pin: pins.IO2,
-    direction: 'in',
-    edge: 'any'
-});
-var led1 = gpio.open({
-    pin: pins.IO4,
-    direction: 'out'
-});
-var led2 = gpio.open({
-    pin: pins.IO7,
-    direction: 'out'
-});
-var led3 = gpio.open({
-    pin: pins.IO8,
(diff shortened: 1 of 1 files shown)


grgustaf:
@jprestwo, take a look at this as a replacement for #338.

samples/arduino/starterkit/Spaceship2.js
```diff
@@ -18,54 +18,43 @@ var gpio = require('gpio');
...
-    pin: pins.IO7,
-    direction: 'out'
-});
-var led3 = gpio.open({
-    pin: pins.IO8,
-    direction: 'out'
-});
-
-var btnstate = false;
+var button = gpio.open({pin: pins.IO2, direction: 'in', edge: 'rising'});
```
jprestwo: You want to change the edge property to 'any' as it was before.

grgustaf: No actually this is intentional but changes the behavior a bit. It will immediately turn on when it gets a rising edge but shuts off at an even 250ms after that. This is a safer way to code because if one edge is missed it will correct itself on the next one. And since it keeps checking until the value is low it will always eventually shut off. The other way would have potentially left the lasers on indefinitely when there was a missed edge.

jimmy-huang:
+1

PR #485, opened 2016-11-28
[General] Fixed build warnings when VARIANT=debug

jimmy-huang:


jimmy-huang:
didn't see patch#483 already submitted, this is duplication

PR #486, opened 2016-11-28
[submodules] Ignore dirty trees or different commits in submodules

grgustaf:
If the user makes changes to e.g. their Zephyr subdirectory, this means
they will not be nagged about those changes when the look at git status.
This seems best because we frequently may have some instrumentation code
checked into Zephyr, and don't need to be constantly reminded.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,15 @@
 [submodule "deps/zephyr"]
 	path = deps/zephyr
 	url = https://gerrit.zephyrproject.org/r/zephyr
+    ignore = all
 [submodule "deps/jerryscript"]
 	path = deps/jerryscript
 	url = https://github.com/Samsung/jerryscript.git
+    ignore = all
 [submodule "deps/ihex"]
 	path = deps/ihex
 	url = https://github.com/arkku/ihex.git
 [submodule "deps/iotivity-constrained"]
 	path = deps/iotivity-constrained
 	url = https://gerrit.iotivity.org/gerrit/p/iotivity-constrained.git
+    ignore = all


pfalcon:
I didn't know about such setting, and I'd personally consider "all" setting for it a bit dangerous.

> This seems best because we frequently may have some instrumentation code checked into Zephyr, and don't need to be constantly reminded.

Even for such usecase, being reminded that you don't use pristine Zephyr, may be useful. I personally find that I rarely commit into submodules, usually just leave uncommitted changes. Being reminded that the tree is patched is generally helpful. Another common usecase I have is switching to my local branch, and switching a submodule to another revision, e.g. switching Zephyr to a newer revision which has a new board added, to test how well zephyr.js builds/works for it. Ditto for JerryScript, to see if Z.js builds well against new version. Then when I switch back to master, being reminded that submodules aren't at revision expected by master is helpful.

Anyway, those are my 2 cents, just wanted to be sure you considered such cases which really can lead to big confusion on a bad day.

grgustaf:
Hmm, maybe that is a safer default. I could always modify this on my local box to satisfy my OCD tendencies. I hate to get a warning message as if something's wrong when it's exactly what I intend.

The problem we've been having is that submodule changes have been committed accidentally a couple times. And they're easy to miss in a code review because they're just one little line without even a "code change" box around it which is what I'm normally looking for. (Really they should be in big bold red or something as it's potentially a big deal, maybe I should file a GH issue.)

Using 'dirty' instead of 'all' would still hide the submodule uncommitted changes, which you say you like to see as well. So I guess you're not too excited about the middle ground either. :)

grgustaf:
I'll go ahead and drop this for now.

PR #487, opened 2016-11-29
[Sensor] Add support for AmbientLightSensor

jimmy-huang:
Extend GenericSensor API to support AmbientLightSensor of W3C,
added support for the Grove Light sensor that comes from the
standard Grove Starter Kit for Arduino 101.  The sensor is designed
to be connected to the analog pins on the board.

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -775,7 +775,7 @@ static double cube_root_recursive(double num, double low, double high) {
     double mid = (low + high) / 2.0;
     double mid3 = mid * mid * mid;
     double diff = mid3 - num;
-    if (diff < margin || (diff * -1.0) > margin)
+    if (ABS(diff) < margin)
         return mid;
     else if (mid3 > num)
         return cube_root_recursive(num, low, mid);
@@ -810,7 +810,7 @@ static void process_light_updates()
                     reading.dval = 10000.0;
                 }
                 else {
-                    resistance = (1023.1 - analog_val) * 10.0 / analog_val;
+                    resistance = (1023.0 - analog_val) * 10.0 / analog_val;
                     base = resistance * 15.0;
                     reading.dval = 10000.0 / cube_root(base * base * base * base);
                 }
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -6,7 +6,7 @@
 #include <string.h>
 #include <device.h>
 #include <init.h>
-#ifdef BUILD_MODULE_AIO
+#if defined(BUILD_MODULE_AIO) || defined(BUILD_MODULE_SENSOR_LIGHT)
 #include <adc.h>
 #endif
 #ifdef BUILD_MODULE_I2C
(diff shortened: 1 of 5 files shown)


src/zjs_sensor.c
```diff
@@ -49,6 +49,7 @@ typedef struct sensor_handle {
 
 static sensor_handle_t *accel_handles = NULL;
 static sensor_handle_t *gyro_handles = NULL;
+static sensor_handle_t *light_handles = NULL;
```
brianjjones: Is zjs_sensor where the light sensor stuff should live?  I thought that file contained sensors found built into the arduino 101, not sure we should add grove ones to it.

arc/src/main.c
```diff
@@ -762,55 +768,132 @@ static int stop_gyro_trigger(struct device *dev)
     return 0;
 }
 
+#ifdef BUILD_MODULE_SENSOR_LIGHT
+static double cube_root_recursive(double num, double low, double high) {
```
brianjjones: Personally I think we should be returning only raw data from sensors and doing conversions in the JavaScript.  Otherwise I fear we are going to start adding too much code specific to certain sensors to ZJS.  Now that we have the ability to use JavaScript modules I think keeping the sensor specific code out of the native space makes sense.  Thoughts?

src/zjs_sensor.c
jimmy-huang: zjs_sensor are for all the w3c sensors, and this implements the w3c sensor api, it's not specific to the grove light, it's more generic light sensor

arc/src/main.c
jimmy-huang: the w3c api expects an ambient light value returned from the callback, now if the zephyr driver supports the grove light, then the driver could convert the value, but this grove light is not supported

rwaldron: > Personally I think we should be returning only raw data from sensors

That would be a violation of the specification that this purports to implement. The [AmbientLightSensor specification](https://www.w3.org/TR/ambient-light/) defines an `illuminance` property on the object located at `event.reading`, eg.

```js
let sensor = new AmbientLightSensor();
sensor.start();
    
sensor.onchange = function(event) {
  console.log(event.reading.illuminance);
};
```


`illuminance` is defined as ambient light level in lux.

brianjjones: Ah see I somehow missed that it was W3C specific.  Never-mind my comment then.

arc/src/main.c
```diff
@@ -762,55 +768,132 @@ static int stop_gyro_trigger(struct device *dev)
...
+                double resistance, base;
+                union sensor_reading reading;
+                // rescale sample from 12bit (Zephyr) to 10bit (Grove)
+                uint16_t analog_val = pin_values[i] >> 2;
+                if (analog_val > 1015) {
+                    // any thing over 1015 will be considered maximum brightness
+                    reading.dval = 10000.0;
+                }
+                else {
+                    resistance = (1023.1 - analog_val) * 10.0 / analog_val;
```
grgustaf: You said 1023.0 above?

samples/AmbientLight.js
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2016, Intel Corporation.
+
+// Test code to use the AmbientLightSensor (subclass of Generic Sensor) API
+// to communicate with the Grove Light sensor on the Arduino 101
+//
```
grgustaf: remove blank comment line

jimmy-huang:
@grgustaf updated patch

grgustaf:
+1

PR #491, opened 2016-11-30
[script] New compare feature for memorystats

brianjjones:
Now memory stats can compare your current branch to a previous
one, showing you the change in ROM/RAM usage.

### DIFF
--- a/scripts/memorystats
+++ b/scripts/memorystats
@@ -40,7 +40,11 @@ divider="--------------------------------------------------------------------"
 
 function runTests() {
 
-    # Run test on the arc side
+    # Make sure things are clean
+    make pristine > /dev/null 2>&1
+    make clean > /dev/null 2>&1
+    echo $(git log | grep commit | head -1)
+    # Run test on the arc side    
     echo "Testing with ARC..."
     make arc >& /dev/null
     if [ $? -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


brianjjones:
Usage instructions:

Flags:
-f = file to use for testing.  If not given it will use all of the files in ./samples
-c = commit id to compare your current branch against
-n = number of commits ago you want to compare your current branch to?

Examples:

// Get the memory stats of your current branch using all the samples
memorystats

// Use samples/I2C.js instead of all the samples
memorystats -f samples/I2C.js 

// Compare current branch to 5331702cef50c5b6d9c0577a4b83d11d1cd705e0
memorystats -f samples/I2C.js -c 5331702cef50c5b6d9c0577a4b83d11d1cd705e0 

// Compare current branch to 10 commits ago
memorystats -f samples/I2C.js -n 10

scripts/memorystats
```diff
@@ -18,47 +18,126 @@ if [ ! -f "deps/zephyr/scripts/sanitycheck" ]; then
...
+
+# If we are not comparing commits, print the current results
+if [[ -z "$numComp" && -z "$nameComp" ]]; then
+    sed '/Full/Q' "/tmp/memorystats_$OUTNAME.txt"
+fi
+
+# If we are comparing results, find the changes and print them
+if [ -n "$numComp" ]; then
+    results=""
+    OUTNAME=$(git log | grep commit | head -$numComp | tail -1 | cut -d' ' -f2)
```
grgustaf: A more straightforward syntax than this "-n 10" would be to let users specify git commits like HEAD~10 or HEAD^^^, which all git commits handle and interpret as 10 commits before HEAD or 3 commits before HEAD respectively. You shouldn't have to actually parse such things, just preserve them and pass them on to the git commands.

But as I said in IRC, we don't need to hold up the patch for that, just keep it in mind.

PR #495, opened 2016-12-02
[JerryScript] Optimize build with LTO and All-in-one flags

jimmy-huang:
This will further reduce final build size by 5-6k, the
default Zephyr target in JerryScript do not have these
enabled.

### DIFF
--- a/Makefile.app
+++ b/Makefile.app
@@ -1,6 +1,12 @@
 JERRY_BASE ?= $(ZJS_BASE)/deps/jerryscript
 
+EXT_JERRY_FLAGS ?= -DENABLE_ALL_IN_ONE=ON
+
+ifneq ($(BOARD), frdm_k64f)
+EXT_JERRY_FLAGS += -DENABLE_LTO=ON
+endif
+
 $(KBUILD_ZEPHYR_APP):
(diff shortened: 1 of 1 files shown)


grgustaf:
Can you explain the all-in-one thing? I looked it up Friday and I remember seeing it looked like it made little difference if you were doing LTO, from the stats at least when it was introduced. What difference does it make now and any idea what it does? :) Otherwise, LGTM.

jimmy-huang:
@grgustaf I am not 100% sure, but reading from this https://github.com/jerryscript-project/jerryscript/issues/626, it's is compiling everything into a single binary, and they introduced it basically a optimization to couple the LTO, since LTO may not always work depending on the platform.

grgustaf:
Alright I'll just merge it. I'd still be interested to see what the difference is with and without the all in one flag for us.

+1

pfalcon:
> However, LTO currently doesn't build on k64f which is ARM-based.

It also doesn't build for qemu_cortex_m3 and bunch of other ARM-based boards in Zephyr. I'd say, it's better to whitelist LTO for specific targets rather than blacklist it.

PR #496, opened 2016-12-02
[modules] Updating analyze.sh to check for all forced modules

brianjjones:
There were a couple modules that got left off, such as Buffer and
Timers.  Now these can be built adding them to the config file.

### DIFF
--- a/scripts/analyze.sh
+++ b/scripts/analyze.sh
@@ -165,6 +165,7 @@ if check_for_require ble || check_config_file ZJS_BLE; then
     MODULES+=" -DBUILD_MODULE_EVENTS"
     echo "export ZJS_EVENTS=y" >> zjs.conf.tmp
     echo "export ZJS_BLE=y" >> zjs.conf.tmp
+    buffer=true;
 fi
 
 if check_for_require aio || check_config_file ZJS_AIO; then
--- a/scripts/analyze.sh
+++ b/scripts/analyze.sh
@@ -25,7 +25,7 @@ BOARD=$1
 SCRIPT=$2
 CONFIG=$3
 
-echo "# Modules found in $SCRIPT:" > prj.conf.tmp
+echo "# Modules found in $SCRIPT:" > prj.conf.tmpq
 echo "# Modules found in $SCRIPT:" > arc/prj.conf.tmp
 tmpstr="# ZJS flags set by analyze.js"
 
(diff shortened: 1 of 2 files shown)


scripts/analyze.sh
```diff
@@ -25,7 +25,7 @@ BOARD=$1
 SCRIPT=$2
 CONFIG=$3
 
-echo "# Modules found in $SCRIPT:" > prj.conf.tmp
+echo "# Modules found in $SCRIPT:" > prj.conf.tmpq
```
grgustaf: typo?

scripts/analyze.sh
```diff
@@ -189,27 +200,32 @@ if check_for_require arduino101_pins || check_config_file ZJS_ARDUINO101_PINS; t
...
 else
     timeout=$(grep setTimeout $SCRIPT)
     if [ $? -eq 0 ]; then
         MODULES+=" -DBUILD_MODULE_TIMER"
+        echo "export ZJS_TIMERS=y" >> zjs.conf.tmp
     else
         immediate=$(grep setImmediate $SCRIPT)
         if [ $? -eq 0 ]; then
             MODULES+=" -DBUILD_MODULE_TIMER"
+            echo "export ZJS_TIMERS=y" >> zjs.conf.tmp
```
grgustaf: These two lines are repeated 3x, I think they could be here once if you restructured the if statements to check the reverse like if !0 && ! check_config && !timeout && !immediate, then timers=false. Default, timers=true. But not a big deal.

scripts/analyze.sh
```diff
@@ -189,27 +200,32 @@ if check_for_require arduino101_pins || check_config_file ZJS_ARDUINO101_PINS; t
...
         fi
     fi
 fi
-buffer=$(grep Buffer\([0-9]*\) $SCRIPT)
-if [ $? -eq 0 ] && [[ $MODULE != *"BUILD_MODULE_BUFFER"* ]]; then
+
+
```
grgustaf: extra blank line here

scripts/analyze.sh
brianjjones: Even shorter now, should do the trick

grgustaf:
+1

jimmy-huang:
+1

jimmy-huang:
@brianjjones can you update commit with signed off by?

PR #497, opened 2016-12-03
[trlite] Preserve uncommitted changes in the tree for trlite tests

grgustaf:
To do this, take the current git diff of the tree and apply that patch
within the .trlite directory.

Also, report the ZJS_BASE dir being used and the diffstats of the
uncommitted changes found to help the user be sure they're building
what they intended.

### DIFF
--- a/scripts/trlite
+++ b/scripts/trlite
@@ -18,6 +18,7 @@ if [ ! -d "$ZJS_BASE" ]; then
 fi
 
 TRLDIR=$ZJS_BASE/.trlite
+TRLPATCH=$TRLDIR/patch
 
 VERBOSE=
 if [ "$1" == "-v" ]; then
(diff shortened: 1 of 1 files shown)


jimmy-huang:
+1

pfalcon:
Sorry, I missed this patch earlier, but I can really relate to the "Preserve uncommitted changes" motive. I have to patch Zephyr.js Makefile with:

~~~
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ update:
 
 # Sets up prj/last_build files
 .PHONY: setup
-setup: update
+setup: #update
        @echo "# This is a generated file" > prj.conf
 ifeq ($(BOARD), qemu_x86)
        @cat fragments/prj.conf.qemu_x86 >> prj.conf
@@ -142,7 +142,7 @@ endif
 # initialize submodules first so clean will succeed in that case. We should find
 # a way to make clean work from the start, but for now this should be harmless.
 .PHONY: clean
-clean: update
+clean: #update
        @if [ -d $(ZEPHYR_SDK_INSTALL_DIR) ]; then \
                if [ -d deps/jerryscript ]; then \
~~~

so it stops erasing changes I've done to submodules (and I do them regularly). I received similar complaint from another guy on our team. We decided to not submit it as a bug for now, as there're more important things to do, but probably will do that later. Just FYI ;-).

grgustaf:
@pfalcon, this was a completely unrelated issue: the patch needs to copy your uncommitted changes into the pristine tree it builds with, and it's only looking at your changes in the main zjs tree. Which is the right thing for trlite I think.

But as to your point, I guess you mean you're checking out a new commit in the submodule and it's always moving it back. I agree that's not a nice behavior, I'm thinking we could just warn the user that they're building with a non-standard dependency.

grgustaf:
(In short, make update should probably run when you're missing a dependency entirely or call it manually, but not automatically just because you're building.)

PR #498, opened 2016-12-03
[aio] Use prototype inheritance to implement AIO functions

grgustaf:
Also, add a spot for a cleanup function to be registered with modules
that will be called by a new zjs_modules_cleanup function. Currently,
this is called only in ashell.

Also, fix what appeared to be a bug where module instances weren't
being released.

### DIFF
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -130,7 +130,6 @@ void javascript_eval_code(const char *source_buffer)
     jerry_release_value(ret_val);
 }
 
-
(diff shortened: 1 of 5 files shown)


grgustaf:
I added some test instrumentation and proved that the bug was real about instances not getting released. By doing a set_native_handle with a free function on the instance, it never got called until I started doing my release.

brianjjones:
LGTM +1

brianjjones:
LGTM +!

jimmy-huang:
+1

Issue #490, opened 2016-11-30
convert.sh doesn't like % in comments in PWM.js


grgustaf:
This is on line 82 of convert.sh. Probably it could be changed to:

```bash
echo -n $CHAR >> $OUTPUT
```

Also, it looks like below there's a `printf "\n"` that should probably be removed? It's not being echoed to the file, where it doesn't belong anyway AFAICT.

jprestwo:
@grgustaf, what is the results of running convert.sh normally? It seems convert and run the JS fine for me. The printf for the newline was put there for some reason, but I cant seem to remember why. I think it was for adding a new line after we write the percentage out, essentially a aesthetic thing.

grgustaf:
@jprestwo, just run "convert.sh samples/PWM.js out" and I think you will see the complaint in the first few lines of output, like I do.

Anyway, I just submitted a PR and you can review that.

grgustaf:
PR  #502 fixes this.

PR #502, opened 2016-12-06
[convert] Fix conversion of % characters, also limit output line length

grgustaf:
Because printf was being used, the % character would have needed special
treatment to avoid a warning.

Also, the dots were overflowing past the end of the line in output which
looked bad after it hit 90%.

Fixes issue #490.


brianjjones:
LGTM +1

jimmy-huang:
+1

Issue #500, opened 2016-12-05
[REG] Performance.now() result over known delay


xiaolongx:
The test result of [test-performance.js](https://github.com/01org/zephyr.js/blob/master/tests/test-performance.js) fail on both Arduino_101 and FRDM_K64F, the diff time is 1020 now in test-performance.js.

Branch | Target Device | Build ID | Date | Result
---|---|---|---|---
master | Arduino_101/FRDM_K64F | 5844f9f | Dec 5, 2016 | PASS
master | Arduino_101/FRDM_K64F | a162db1 | Dec 5, 2016 | FAIL

grgustaf:
Interesting, @pfalcon does this surprise you? Do you see the same thing?

pfalcon:
@grgustaf : Will try later today. But I guess we'll just need to bump the delta...

pfalcon:
Yes, I confirm that with the latest master (2279b157ad4fb0c12f5aa4323549844e7448c503) I get the failure.  I submitted https://github.com/01org/zephyr.js/pull/508 to address (well, work around) that. It would be nice to understand why it's regressed, but with the current extensive development of both Z and Z.js, I'd say that it's not the best application of effort. We can return to that later, when jitter/latency like that will actually affect real-world usecases.

grgustaf:
Makes sense, thanks.

xiaolongx:
This issue cannot reproduce on both Arduino 101 and FRDM_K64F with commit 31c3908.

PR #508, opened 2016-12-06
[tests] test-performance.js: Bump delta to 21ms.

pfalcon:
With upgrade to Zephyr 1.6 and refactors on Zephyr.js side, we're now
seeing delta of 20ms on both arduino_101 and frdm_k64f.

### DIFF
--- a/tests/test-performance.js
+++ b/tests/test-performance.js
@@ -19,6 +19,6 @@ setTimeout(function() {
     var after = performance.now();
     var diff = after - before;
     // Allow for some jitter, especially on Linux
-    assert(diff >= 989 && diff <= 1011, "performance.now() result over known delay");
+    assert(diff >= 979 && diff <= 1021, "performance.now() result over known delay");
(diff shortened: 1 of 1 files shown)


grgustaf:
Thanks. Do you think this is acceptable fuzziness or a problem we should bring up with the Zephyr folks?

+1

pfalcon:
> Do you think this is acceptable fuzziness or a problem we should bring up with the Zephyr folks?

I was surprised to see even 10ms latency with the original submission (then it affected arduino_101, but not frdm_k64f). But since them many things changed on z.js side, e.g. callback queue was reimplemented. I actually didn't yet have chance to look into how callback queue managed in z.js, and well, I've got to think I'd actually like look into it before e.g. calling to Zephyr folks. As I mentioned in https://github.com/01org/zephyr.js/issues/500#issuecomment-265134064 , I guess this can wait for now, but probably will be worth to re-visit some time later, when both zephyr and zephyr.js stabilize in their functionality.

Issue #501, opened 2016-12-05
Issues upgrading to the recent JerryScript upstream


pfalcon:
After https://github.com/01org/zephyr.js/pull/450 , the next logical step is to upgrade JerryScript which is quite behind the upstream (and so for example doesn't support all boards available in Zephyr). I did this in my fork, https://github.com/pfalcon/zephyr.js/commits/aeolus-2016.12 , but unfortunately, there're an issue: when building against such JrS upstream, there's an interactive question from Kconfig:

~~~
System clock's h/w timer frequency (SYS_CLOCK_HW_CYCLES_PER_SEC) [] (NEW) 
~~~

Default value for SYS_CLOCK_HW_CYCLES_PER_SEC is in 100% cases set by a particular SoC's config (e.g. arch/nios2/soc/nios2f-zephyr, arch/arc/soc/em11d, arch/x86/soc/atom, arch/arm/soc/atmel_sam3, arch/arm/soc/nordic_nrf5/nrf52, etc.), so it should never should be force-asked from a user. If it is, then apparently board/soc/arch isn't selected/passed properly on that Kconfig invocation. This needs further looking into.

All in all, I submit this issue to let everyone be in loop on it, and have a place to discuss/report progress on it.

Excerpt from build log leading to Kconfig issue:

~~~
  CC      src/zjs_timers.o
  CC      src/zjs_util.o
  CC      src/zjs_k64f_pins.o
  LD      src/built-in.o
  AR      libzephyr.a
Building libjerry-core.a
make -C /home/pfalcon/projects-3rdparty/Embedded/Zephyr/linaro-aeolus/zephyr.js/deps/jerryscript -f targets/zephyr/Makefile.zephyr BOARD=frdm_k64f jerry
targets/zephyr/Makefile.zephyr:47: build/frdm_k64f/zephyr/Makefile.export: No such file or directory
make -f ./targets/zephyr/Makefile O="build/frdm_k64f/zephyr" V= USER_LIBS="jerry-core" USER_LIB_INCLUDE_DIR="-L /mnt/hdd/projects-3rdparty/Embedded/Zephyr/linaro-aeolus/zephyr.js/deps/jerryscript/build/frdm_k64f/obj-frdm_k64f/lib" TARGET_ZEPHYR=./targets/zephyr initconfig
[...]

PR #509, opened 2016-12-06
[Makefile] Build JerryScript directly, instead of recursively via Zephyr.

pfalcon:
Such recursive approach no longer works with current upstream JerryScript
and leads to confusing errors.


pfalcon:
Fixes https://github.com/01org/zephyr.js/issues/501 . Note - JrS upgrade is not included in this PR, can be done later, or I can add it here.

grgustaf:
+1, need conflicts resolved

poussa:
+1, need conflicts resolved.

Glad to see Makefile.app to go way, that was hackish solution anyway.

pfalcon:
Patch rebased and retested.

Note that LTO breaks more things than just frdm_k64f: https://github.com/01org/zephyr.js/pull/495#issuecomment-265500924

jimmy-huang:
@pfalcon yeah, it's probably ARM in general, not just frdm_k64f, but for now we only have 2 boards, we can probably change the check if arch is not ARM.

ifneq ($(ARCH), arm)?

pfalcon:
> but for now we only have 2 boards

Well, I'm working (and submitting patches) towards supporting more (actually, any board supported by Zephyr).

> ifneq ($(ARCH), arm)?

Unfortunately, ARCH isn't available in Z.js Makefile. If the current usecase is to optimize code size for arduino_101, then perhaps following would be good for now?

~~~
ifeq ($(BOARD), arduino_101)
~~~

grgustaf:
>~~~
>ifeq ($(BOARD), arduino_101)
>~~~

Sounds right to me for now.

jimmy-huang:
yeah, that's good

grgustaf:
conflicts resolved, merging

PR #510, opened 2016-12-06
[fs] File System module

jprestwo:


src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+            DBG_PRINT("Opening file %s in mode %s throws exception if file does not exists\n", path, mode);
+            // TODO: Throw exception
+        } else {
+            DBG_PRINT("Creating new file %s\n", path);
+        }
+    }
+
+    handle->error = fs_open(&handle->fp, path);
+    if (handle->error != 0) {
+        DBG_PRINT("could not open file: %s\n", path);
```
jimmy-huang: throw exception here?

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+                               const jerry_value_t this,
+                               const jerry_value_t argv[],
+                               const jerry_length_t argc,
+                               uint8_t async)
+{
+    file_handle* handle;
+
+    if (!jerry_value_is_object(argv[0])) {
+        ERR_PRINT("first parameter was not file descriptor\n");
+        return ZJS_UNDEFINED;
```
jimmy-huang: throw exception?

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[1])) {
+            ERR_PRINT("second parameter must be callback function\n");
+            return ZJS_UNDEFINED;
```
jimmy-huang: throw exception?

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+        }
+    }
+    if (!jerry_get_object_native_handle(argv[0], (uintptr_t*)&handle)) {
+        ERR_PRINT("native handle not found\n");
+        return ZJS_UNDEFINED;
```
jimmy-huang: ditto

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[1])) {
+            ERR_PRINT("second parameter was not function\n");
+            return ZJS_UNDEFINED;
+        }
+    }
+
+    jerry_size_t jlen = jerry_get_string_size(argv[0]);
+    char path[jlen + 1];
```
jimmy-huang: check jlen for overflow

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+    int wlen = jerry_string_to_char_buffer(argv[0], (jerry_char_t *)path, jlen);
+    if (jlen != wlen) {
+        ERR_PRINT("size mismatch\n");
+        return ZJS_UNDEFINED;
```
jimmy-huang: throw exception?

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+        jerry_value_t error = jerry_create_number(ret);
+        zjs_signal_callback(id, &error, sizeof(jerry_value_t));
+    }
+
+    return ZJS_UNDEFINED;
+}
+
+static jerry_value_t zjs_unlink_sync(const jerry_value_t function_obj,
+                                    const jerry_value_t this,
```
jimmy-huang: align indentation

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+static jerry_value_t zjs_unlink_async(const jerry_value_t function_obj,
+                                     const jerry_value_t this,
+                                     const jerry_value_t argv[],
+                                     const jerry_length_t argc) {
+    return zjs_unlink(function_obj, this, argv, argc, 1);
+}
+
+static jerry_value_t zjs_read(const jerry_value_t function_obj,
+                             const jerry_value_t this,
```
jimmy-huang: align indentation

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+                             const jerry_value_t argv[],
+                             const jerry_length_t argc,
+                             uint8_t async)
+{
+    file_handle* handle;
+    int err = 0;
+
+    if (argc < 6 && async) {
+        ERR_PRINT("not enough parameters\n");
+        return ZJS_UNDEFINED;
```
jimmy-huang: throw exception, and the rest of this code.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    // if a position was specified, seek to it before writing
+
+    if (fs_seek(&handle->fp, position, SEEK_SET) != 0) {
+        ERR_PRINT("error seeking to position\n");
+        return ZJS_UNDEFINED;
+    }
+
+    DBG_PRINT("reading into fp=%p, buffer=%p, offset=%lu, length=%lu\n", &handle->fp, buffer->buffer, offset, length);
+
+    uint32_t ret = fs_read(&handle->fp, buffer->buffer + offset, length);
```
jimmy-huang: check offset, length position like negative or overflow?

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+                               const jerry_length_t argc,
+                               uint8_t async)
+{
+    file_handle* handle;
+    int position = 0;
+    jerry_value_t js_cb;
+
+    if (argc < 5) {
+        ERR_PRINT("not enough parameters\n");
+        return ZJS_UNDEFINED;
```
jimmy-huang: again, throw exceptions

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    if (!jerry_get_object_native_handle(argv[0], (uintptr_t*)&handle)) {
+        ERR_PRINT("native handle not found\n");
+        return ZJS_UNDEFINED;
+    }
+    zjs_buffer_t* buffer = zjs_buffer_find(argv[1]);
+    uint32_t offset = jerry_get_number_value(argv[2]);
+    uint32_t length = jerry_get_number_value(argv[3]);
+
+    // if a position was specified, seek to it before writing
+    if (fs_seek(&handle->fp, position, SEEK_SET) != 0) {
```
jimmy-huang: check offset and length

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+static jerry_value_t zjs_truncate(const jerry_value_t function_obj,
+                                  const jerry_value_t this,
+                                  const jerry_value_t argv[],
+                                  const jerry_length_t argc,
+                                  uint8_t async)
+{
+    fs_file_t fp;
+    if (!jerry_value_is_number(argv[1])) {
+        ERR_PRINT("second parameter must be length\n");
```
jimmy-huang: throw exception and rest of the code here

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    }
+    if (jerry_value_is_object(argv[0])) {
+        file_handle* handle;
+        if (!jerry_get_object_native_handle(argv[0], (uintptr_t*)&handle)) {
+            ERR_PRINT("native handle not found\n");
+            return ZJS_UNDEFINED;
+        }
+        fp = handle->fp;
+    } else if (jerry_value_is_string(argv[0])) {
+        jerry_size_t jlen = jerry_get_string_size(argv[0]);
```
jimmy-huang: check jlen

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+                               const jerry_value_t this,
+                               const jerry_value_t argv[],
+                               const jerry_length_t argc,
+                               uint8_t async)
+{
+    jerry_value_t js_cb;
+    uint32_t mode = MODE_A_PLUS;
+    if (!jerry_value_is_string(argv[0])) {
+        ERR_PRINT("first parameter was not path\n");
+        return ZJS_UNDEFINED;
```
jimmy-huang: throw exception here and rest of the code

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+static jerry_value_t zjs_mkdir(const jerry_value_t function_obj,
+                               const jerry_value_t this,
+                               const jerry_value_t argv[],
+                               const jerry_length_t argc,
+                               uint8_t async)
+{
+    jerry_value_t js_cb;
+    uint32_t mode = MODE_A_PLUS;
+    if (!jerry_value_is_string(argv[0])) {
+        ERR_PRINT("first parameter was not path\n");
```
jimmy-huang: do you have check if path is valid here? like string is malformed?

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+            } else {
+                js_cb = argv[2];
+            }
+        }
+        if (!jerry_value_is_string(argv[1])) {
+            ERR_PRINT("second parameter must be mode string\n");
+            return ZJS_UNDEFINED;
+        } else {
+            jerry_size_t jlen = jerry_get_string_size(argv[0]);
+            char str[jlen + 1];
```
jimmy-huang: check jlen

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+static jerry_value_t zjs_readdir(const jerry_value_t function_obj,
+                                 const jerry_value_t this,
+                                 const jerry_value_t argv[],
+                                 const jerry_length_t argc,
+                                 uint8_t async)
+{
+    jerry_value_t array;
+    jerry_size_t jlen = jerry_get_string_size(argv[0]);
+    char path[jlen + 1];
```
jimmy-huang: check jlen

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+    int num_files = 0;
+    int res;
+    fs_dir_t dp;
+    static struct fs_dirent entry;
+
+    res = fs_opendir(&dp, path);
+    if (res) {
+        ERR_PRINT("Error opening dir[%d]\n", res);
+        return res;
```
jimmy-huang: invalid return value here, should be a jerry_value, throw exception

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+        DBG_PRINT("found file %s\n", entry.name);
+    }
+
+    fs_closedir(&dp);
+
+    res = fs_opendir(&dp, path);
+    if (res) {
+        ERR_PRINT("Error opening dir[%d]\n", res);
+        return res;
```
jimmy-huang: ditto

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[1])) {
+            ERR_PRINT("second parameter was not callback function\n");
+            return ZJS_UNDEFINED;
+        }
+    }
+
+    jerry_size_t jlen = jerry_get_string_size(argv[0]);
+    char path[jlen + 1];
```
jimmy-huang: check jlen

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+                                    const jerry_value_t argv[],
+                                    const jerry_length_t argc) {
+    return zjs_stat(function_obj, this, argv, argc, 1);
+}
+
+jerry_value_t zjs_fs_init()
+{
+    jerry_value_t fs = jerry_create_object();
+
+    zjs_obj_add_function(fs, zjs_open_async, "open");
```
jimmy-huang: use the set prototype pattern for object creations, like in GPIO that @grgustaf wrote

samples/FsAsync.js
```diff
@@ -0,0 +1,37 @@
+var fs = require('fs');
+
+console.log("File System sample");
+
+fs.open('testfile.txt', 'w+', function(err, fd) {
```
jimmy-huang: it's better to be consistent on using single quote or double quotes for strings

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    if (!jerry_value_is_number(argv[4])) {
+        ERR_PRINT("fifth parameter was not position\n");
+        return ZJS_UNDEFINED;
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[5])) {
+            ERR_PRINT("sixth parameter was not callback function\n");
+            return ZJS_UNDEFINED;
+        }
+    }
```
grgustaf: Elsewhere we just do all these checks together and say "invalid arguments" or something. Stick with that pattern to save space in strings and maybe code for this check. Being so specific isn't worth the overhead.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    if (argc < 6 && async) {
+        ERR_PRINT("not enough parameters\n");
+        return ZJS_UNDEFINED;
+    }
+
+    if (!jerry_value_is_object(argv[0])) {
+        ERR_PRINT("first parameter was not file descriptor\n");
+        return ZJS_UNDEFINED;
+    }
+    if (!jerry_value_is_object(argv[1])) {
```
grgustaf: Perhaps you should use zjs_buffer_find to make sure it's a buffer specifically? Not sure if we've been consistent about doing that. Maybe we should have a zjs_is_buffer().

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    }
+
+    DBG_PRINT("reading into fp=%p, buffer=%p, offset=%lu, length=%lu\n", &handle->fp, buffer->buffer, offset, length);
+
+    uint32_t ret = fs_read(&handle->fp, buffer->buffer + offset, length);
+
+    if (ret != length) {
+        ERR_PRINT("could not read %lu bytes, only %lu were read\n", length, ret);
+        err = -1;
+    }
```
grgustaf: I don't think that needs to be an error, just a fact. They provide the maximum they want to read and if we don't have that much to give, that's fine.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+        args[0] = jerry_create_number(err);
+        args[1] = jerry_create_number(ret);
+        args[2] = argv[1];
+
+        zjs_callback_id id = zjs_add_callback_once(argv[5],
+                                                   this,
+                                                   NULL,
+                                                   NULL);
+
+        zjs_signal_callback(id, args, sizeof(jerry_value_t) * 3);
```
grgustaf: Am I understanding right that the underlying call is synchronous but we're making it artificially asynchronous? Since we're Zephyr-specific I'm not sure that makes sense either. Probably makes more sense to keep it simple and just use the model here. You're still blocking the JS from running while you wait for the read to come back so there's no benefit I can see to putting off the answer until later.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+        position = jerry_get_number_value(argv[4]);
+    } else {
+        if (async) {
+            if (!jerry_value_is_function(argv[4])) {
+                ERR_PRINT("fifth parameter was not callback function\n");
+                return ZJS_UNDEFINED;
+            }
+            js_cb = argv[4];
+        }
+    }
```
grgustaf: Same point about collapsing all these tests into one error.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    if (!jerry_get_object_native_handle(argv[0], (uintptr_t*)&handle)) {
+        ERR_PRINT("native handle not found\n");
+        return ZJS_UNDEFINED;
+    }
+    zjs_buffer_t* buffer = zjs_buffer_find(argv[1]);
```
grgustaf: Similar to a comment above - this could fail if they pass something weird so make this part of your argument validation check and error above.

src/zjs_fs.c
grgustaf: True, you need to make sure you have room in the target buffer since you are using its internal C buffer directly. We might want a helper function added to Buffer to assist, see what's needed.

src/zjs_fs.c
grgustaf: When you use JS Buffer APIs they check for overflow but right you have to do the checks yourself when using the buffer directly. As I mention below, maybe a helper function in buffer class could make it easier.

src/zjs_fs.c
grgustaf: Again collapse argument checks.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    }
+    if (argc > 2) {
+        if (async) {
+            if (!jerry_value_is_function(argv[2])) {
+                ERR_PRINT("third parameter must be callback function\n");
+                return ZJS_UNDEFINED;
+            } else {
+                js_cb = argv[2];
+            }
+        }
```
grgustaf: Again collapse argument checks - don't need detail on second/third/fourth arg.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+    if (!jerry_value_is_string(argv[0])) {
+        ERR_PRINT("first parameter was not path\n");
+        return ZJS_UNDEFINED;
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[1])) {
+            ERR_PRINT("second parameter was not callback function\n");
+            return ZJS_UNDEFINED;
+        }
+    }
```
grgustaf: Again collapse checks.

src/zjs_fs.c
grgustaf: Follow this pattern to make the changes Jimmy mentions:
`git show d4e1d64 -- 'src/zjs_aio.[ch]'`

grgustaf:
Also add API doc and maybe at least start a standard tests/test-fs.js that QA can expand on.

src/zjs_fs.c
jprestwo: Yes all of these API's are synchronous under the hood. Making them asynchronous is just following the Node.js API spec. If having the asynchronous calls doesn't matter, we can remove this logic and just have only synchronous calls. We should probably bring this up for discussion in the next meeting.

grgustaf: We discussed in IRC that one compromise solution is to offer the async APIs but allow them to be left out if not needed. Since they don't really work the way they're advertised, they're more for compatiblity with existing code. The synchronous APIs should be called out in the API docs as the ones to use for code specifically written for ZJS at this time.

src/zjs_fs.c
grgustaf: Oh also search for zjs_aio_cleanup in zjs_modules.c, you probably need to add your cleanup function there so it will get called.

samples/FsAsync.js
```diff
@@ -0,0 +1,37 @@
...
+            console.log("fd has unknown type");
+        }
+    });
+    var wbuf = new Buffer("write some bytes");
+    fs.write(fd, wbuf, 0, wbuf.length, 0, function(err, written, buffer) {
+        console.log("wrote " + written + " bytes");
+        var rbuf = new Buffer(16);
+        fs.read(fd, rbuf, 0, 16, 0, function(err, num, buf) {
+            console.log("read from file; error=" + err);
+            console.log("buffer=" + buf.toString('ascii'));
```
jprestwo: It wont, it just prints as "[Object object]", AFAIK.

src/zjs_fs.c
```diff
@@ -0,0 +1,940 @@
...
+
+        zjs_signal_callback(id, args, sizeof(jerry_value_t) * 2);
+
+        return ZJS_UNDEFINED;
+    }
+
+    return handle->fd_val;
+}
+
+static jerry_value_t zjs_open_sync(const jerry_value_t function_obj,
```
jprestwo: We discussed this but node "default" is async. The latest PR defaults to use xxxSync() as all the API's to match Node, and all the async API's are compiled out.

scripts/analyze.sh
```diff
@@ -175,6 +175,20 @@ if check_for_require aio || check_config_file ZJS_AIO; then
...
+if check_for_require fs || check_config_file ZJS_FS; then
+    >&2 echo Using module: FS
+    MODULES+=" -DBUILD_MODULE_FS -DBUILD_MODULE_BUFFER"
+    echo "CONFIG_FILE_SYSTEM=y" >> prj.conf.tmp
+    echo "CONFIG_FILE_SYSTEM_FAT=y" >> prj.conf.tmp
+    echo "CONFIG_DISK_ACCESS_FLASH=y" >> prj.conf.tmp
+    echo "CONFIG_FS_FAT_FLASH_DISK_W25QXXDV=y" >> prj.conf.tmp
+    echo "CONFIG_FLASH=y" >> prj.conf.tmp
+    echo "CONFIG_SPI=y" >> prj.conf.tmp
+    echo "CONFIG_GPIO=y" >> prj.conf.tmp
```
jprestwo: Yep, unfortunately its not direct memory access. Not sure why it needs GPIO (maybe to turn Flash on) but their example requires it.

docs/fs.md
```diff
@@ -0,0 +1,146 @@
...
+-----------------
+
+### openSync
+`object openSync(string path, string mode);`
+
+Opens a file.
+
+`path` is the name/path of the file to open
+
+`mode` is the mode to open the file in (w, r, w+ etc)
```
jprestwo: These are just to satisfy the Node API. On Zephyr they really don't do anything.

samples/FsAsync.js
```diff
@@ -0,0 +1,44 @@
...
+        console.log('wrote ' + written + ' bytes');
+        var rbuf = new Buffer(16);
+        fs.read(fd, rbuf, 0, 16, 0, function(err, num, buf) {
+            console.log('read from file; error=' + err);
+            console.log('buffer=' + buf.toString('ascii'));
+        });
+    });
+});
+
+fs.open('file.txt', 'w', function(err, fd) {
```
grgustaf: This is weird because you're not saving the returned file descriptor number. Should probably do that and at least close it again later.

samples/FsSync.js
```diff
@@ -0,0 +1,37 @@
...
+
+var fd1 = fs.openSync('file.txt', 'w');
+
+fs.writeFileSync('string.txt', 'string data', {});
+fs.writeFileSync('buffer.txt', new Buffer('buffer data'));
+
+var files = fs.readdirSync('/');
+for (var i = 0; i < files.length; ++i) {
+    console.log('files[' + i + ']=' + files[i]);
+}
```
grgustaf: What happens if you don't close a file? Can we clean them up in cleanup() function at least so they're not left hanging in ashell restarts? (Maybe you've done this already, I'll see...)

scripts/analyze.sh
```diff
@@ -216,6 +216,20 @@ if check_for_require aio || check_config_file ZJS_AIO; then
     echo "export ZJS_AIO=y" >> zjs.conf.tmp
 fi
 
+if check_for_require fs || check_config_file ZJS_FS; then
+    >&2 echo Using module: FS
+    MODULES+=" -DBUILD_MODULE_FS -DBUILD_MODULE_BUFFER"
+    echo "CONFIG_FILE_SYSTEM=y" >> prj.conf.tmp
+    echo "CONFIG_FILE_SYSTEM_FAT=y" >> prj.conf.tmp
+    echo "CONFIG_DISK_ACCESS_FLASH=y" >> prj.conf.tmp
+    echo "CONFIG_FS_FAT_FLASH_DISK_W25QXXDV=y" >> prj.conf.tmp
```
grgustaf: This looks A101-specific so should probably be conditional on that? Not sure how we'd handle this when we have multiple devices with flash filesystem support.

src/zjs_fs.c
```diff
@@ -0,0 +1,1043 @@
...
+#define MODE_AX_PLUS    (FLAGS_READABLE | FLAGS_APPENDABLE | FLAGS_NO_CREATE)
+
+#define MAX_PATH_LENGTH 128
+
+typedef struct file_handle_t {
+    fs_file_t fp;
+    uint16_t mode;
+    jerry_value_t fd_val;
+    int error;
+} file_handle;
```
grgustaf: file_handle and file_handle_t are backwards here. The idea is file_handle_t is a newly defined type that replaces "struct file_handle".

src/zjs_fs.c
```diff
@@ -0,0 +1,1043 @@
...
+static uint16_t get_mode(char* str)
+{
+    uint16_t mode = 0;
+    if (strcmp(str, "r") == 0) {
+        mode = MODE_R;
+    } else if (strcmp(str, "r+") == 0) {
+        mode = MODE_R_PLUS;
+    } else if (strcmp(str, "w") == 0) {
+        mode = MODE_W;
+    } else if (strcmp(str, "wx") == 0) {
```
grgustaf: You should document these x (no-create) values too in the doc page.

src/zjs_fs.c
```diff
@@ -0,0 +1,1043 @@
...
+        return zjs_error("native handle not found");
+    }
+    if (entry->type == FS_DIR_ENTRY_DIR) {
+        return jerry_create_boolean(true);
+    } else {
+        return jerry_create_boolean(false);
+    }
+}
+
+void free_stats(const uintptr_t native)
```
grgustaf: should be static?

src/zjs_fs.c
```diff
@@ -0,0 +1,1043 @@
...
+        return zjs_error("first parameter must be path string");
+    }
+    if (!jerry_value_is_string(argv[1])) {
+        return zjs_error("second parameter must be flags string");
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[argc - 1])) {
+            return zjs_error("last parameter must be callback function");
+        }
+    }
```
grgustaf: These are a lot of big strings, all three could probably just be "invalid argument" and they can look up the details.

src/zjs_fs.c
```diff
@@ -0,0 +1,1043 @@
...
+    if (!jerry_value_is_number(argv[3])) {
+        return zjs_error("fourth parameter was not length");
+    }
+    if (!jerry_value_is_number(argv[4])) {
+        return zjs_error("fifth parameter was not position");
+    }
+    if (async) {
+        if (!jerry_value_is_function(argv[5])) {
+            return zjs_error("sixth parameter was not callback function");
```
grgustaf: here's another couple hundred bytes of strings, excessive :)

samples/FsAsync.js
jprestwo: The FD doesnt get returned in the async version, it is provided in the callback function.

src/zjs_fs.c
jprestwo: I am conflicted on these. I added them because Node had them, but then I realized that Zephyr has no concept of permissions, so they go unused. Not sure whether to completely remove them or not. The docs do say that they don't matter on Zephyr.

samples/FsSync.js
jprestwo: I don't hold onto the file descriptors; they are part of the object handle given back to the JS script. I am not really sure how we should handle situations like this. Seems like user error to not close the file.

samples/FsAsync.js
grgustaf: ah...

samples/FsSync.js
grgustaf: It is user error, but for ashell we can't just let user errors break things.

Actually this points out, the cleanup functions aren't really needed in "production mode" (vs. ashell). So we would save some space by #ifdef'ing out the cleanup stuff in that case. In this case where we might have to keep a list of open file descriptors - that would be another thing that could be #ifdef'd.

How about at least put a TODO comment in the file that we should clean up leaked file descriptors someday.

src/zjs_fs.c
grgustaf: These aren't permissions; they are weird bonus modes Node adds above and beyond normal fopen(). The 'x' means "fail if the file already exists", i.e "eXpect to create the file".

I think the idea of this is to avoid race conditions. You don't want another process to create the file and you happen to open it right then and truncate it. But ZJS is a single-task environment, so there's no one else around to be messing with the filesystem. So it should be sufficient for the user to check existence with stat() and then create the file.

Now I see you misinterpreted the mode flags a bit, I'll go comment elsewhere.

src/zjs_fs.c
```diff
@@ -0,0 +1,1048 @@
...
+
+    handle->mode = get_mode(mode);
+
+    /*
+     * TODO: Currently Zephyr has no concept of file permissions, when you open
+     *       a file it is automatically readable and writable. This comment
+     *       is left as a reminder that if other systems/OS have permissions
+     *       we may want to implement the logic, but for now the 'mode' parameter
+     *       is unused.
+     */
```
grgustaf: You're right that there are no permissions. If there were, then opening for read or write might fail because of permissions errors. But the modes are not really about permissions, they are about what the user intends to do with the file. So we can still support them. For example, if the user asks to open a file for reading and then tries to write to it, this should fail. The modes are also important in terms of where writes will occur. If they use 'a', then writes should always go to the end of the file and no allow them to seek backwards.

Here's the easy way out... to land your patch quickly, we need the bare minimum of support for 'r' and 'w'. With those, the user can do anything, though certain operations will be a big pain because 'w' wipes out (truncates) the file if it exists. So get those two working, fail open if anything else is passed, and document that those are the only things supported at present. Add test cases that prove that they're working correctly:

For 'w' mode:
* Prove open will create a new file when it doesn't exist
* Prove open will overwrite a file that does exist
* Prove the file contents can be read back with an 'r', so they were written correctly.
* Prove that read() fails on this descriptor.
* Prove that you can seek around and modify different parts (somewhat pointless in this mode, but valid).

For 'r' mode:
* Prove that open fails with 'an exception' if the file doesn't exist.
* Prove that you can seek around and read different parts.
* Prove that write() fails.

Worth testing: can you open the same file for writing and then for reading with a different descriptor, open at the same time? Can you then make changes with the write descriptor and immediately observe them with the read one?

src/zjs_fs.c
```diff
@@ -0,0 +1,1048 @@
...
+        if (!jerry_value_is_function(argv[2])) {
+            return invalid_args();
+        }
+    }
+    if (jerry_value_is_object(argv[0])) {
+        file_handle_t* handle;
+        if (!jerry_get_object_native_handle(argv[0], (uintptr_t*)&handle)) {
+            return zjs_error("native handle not found");
+        }
+        fp = handle->fp;
```
grgustaf: For now, if supporting only 'r' and 'w' as I suggested, truncate should only be allowed if it was opened for 'w'. Later, it would be okay for 'r+' or 'w+' too. Not sure about a/a+, either not supported or only if truncate length is >= current seek position I guess.

tests/test-fs.js
```diff
@@ -111,4 +111,4 @@ fs.closeSync(fd);
 
 fs.unlinkSync('testfile.txt');
 
-console.log("TOTAL: " + passed + " of " + total + " passed");
+console.log("TOTAL: " + passed + " of " + total + " passed");
```
grgustaf: don't remove the newline :)

src/zjs_fs.c
```diff
@@ -409,19 +413,27 @@ static jerry_value_t zjs_read(const jerry_value_t function_obj,
...
-    // if a position was specified, seek to it before writing
-    if (fs_seek(&handle->fp, (uint32_t)position, SEEK_SET) != 0) {
+    if (handle->mode == MODE_A_PLUS) {
+        // mode is a+, seek to read position
+        if (fs_seek(&handle->fp, handle->rpos, SEEK_SET) != 0) {
+            return zjs_error("error seeking to position");
+        }
+    }
+    // if a position was specified, seek to it before reading
+    if (fs_seek(&handle->fp, (uint32_t)position, SEEK_CUR) != 0) {
```
grgustaf: This ignores the seek you maybe just did above. Also I think this should be SEEK_SET instead of SEEK_CUR, the position passed in I think is an absolute position from the beginning of the file, not from the current point.

Also, according to node docs position can be "null" and that means just use the current position. So you ought to check for that I guess; from this comment it seems you intended to?

src/zjs_fs.c
```diff
@@ -409,19 +413,27 @@ static jerry_value_t zjs_read(const jerry_value_t function_obj,
...
         return zjs_error("error seeking to position");
     }
+    handle->rpos += position;
```
grgustaf: Make sure to do this only in the case where you do the seek above, and again it's rpos = position I think, not += (absolute, not relative).

src/zjs_fs.c
```diff
@@ -510,6 +522,14 @@ static jerry_value_t zjs_write(const jerry_value_t function_obj,
     if (!jerry_get_object_native_handle(argv[0], (uintptr_t*)&handle)) {
         return zjs_error("native handle not found");
     }
+
+    if (handle->mode == MODE_R) {
+        return zjs_error("file is not open for writing");
+    }
+    if (handle->mode == MODE_A && position) {
+        return zjs_error("cannot seek in mode a");
```
grgustaf: Actually the nodejs docs for write() explicitly state this is ignored, not an error, for the append case.

So actually you don't need to do this check at all here.

src/zjs_fs.c
```diff
@@ -524,9 +544,16 @@ static jerry_value_t zjs_write(const jerry_value_t function_obj,
...
-    if (fs_seek(&handle->fp, (uint32_t)position, SEEK_SET) != 0) {
-        return zjs_error("error seeking to position\n");
+    if (handle->mode == MODE_A || handle->mode == MODE_A_PLUS) {
+        // if in append mode, seek to end (ignoring position parameter)
+        if (fs_seek(&handle->fp, 0, SEEK_END) != 0) {
+            return zjs_error("error seeking start");
+        }
+    } else {
+        // if a position was specified, seek to it before writing
+        if (fs_seek(&handle->fp, (uint32_t)position, SEEK_SET) != 0) {
```
grgustaf: Again you should allow position to be null above, and do this seek only if it's not null.

Yuck - apparently you're right in the arg handling code above that position is *optional* for write, even though it was not optional for read. Ew. But it can still be null.

src/zjs_fs.c
```diff
@@ -413,17 +421,21 @@ static jerry_value_t zjs_read(const jerry_value_t function_obj,
         return zjs_error("offset + length overflows buffer");
     }
 
-    if (handle->mode == MODE_A_PLUS) {
+    // a+ or read from current position (position == null)
+    if ((handle->mode == MODE_A_PLUS) || from_cur) {
```
grgustaf: I don't think you want the from_cur part here. If you're not in A_PLUS, rpos isn't needed, just trust the actually current position the OS is tracking.

src/zjs_fs.c
```diff
@@ -413,17 +421,21 @@ static jerry_value_t zjs_read(const jerry_value_t function_obj,
...
         // mode is a+, seek to read position
         if (fs_seek(&handle->fp, handle->rpos, SEEK_SET) != 0) {
             return zjs_error("error seeking to position");
         }
     }
-    // if a position was specified, seek to it before reading
-    if (fs_seek(&handle->fp, (uint32_t)position, SEEK_CUR) != 0) {
-        return zjs_error("error seeking to position");
+    if (!from_cur) {
```
grgustaf: If I'm right above, this is the only use of from_cur so you could just move the is_null down here.

grgustaf: Oh and here you want AND NOT A_PLUS or else you're still going to overwrite your seek position you set above. :)

grgustaf:
I don't see any other option than to finally merge this patch.

+1

PR #512, opened 2016-12-06
[Snapshot] Enable snapshot build feature

jimmy-huang:
The snapshot build is enabled by passing SNAPSHOT=on to make,
this will build a snapshot tool that will first parse the JS
script, and generate a zjs_snapshot_gen.c file that stores
the byte code of the running JS, and then the main program
will just run the snapshot byte code instead of parsing
the JS at run time.

For example:

$ make JS=samples/AIO.js SNAPSHOT=on all
$ make dfu-all

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 BOARD ?= arduino_101
 UPDATE ?= exit
 
-# pass TRACE=y to trace malloc/free in the ZJS API
-TRACE ?= n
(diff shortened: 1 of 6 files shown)


grgustaf:
I see the ARC ipm console message but no other output with hello world, pwm, or aio samples.

grgustaf:
Actually works when I do make clean / pristine. We should work to make the build system detect and automate this, but okay for now.

+1

PR #516, opened 2016-12-07
[tests] Update test-pwm.js for bug fix

cuiyanx:


jimmy-huang:
Is this a work-around for the bug? or a actual fix?  If this is a zephyr bug, we should file a bug as well.

cuiyanx:
This is an algorithmic problem, can not be a bug. Now time precision is 20 ms for zephyr. e.g. Timer.

time precision | set(ms) | actual(ms)
---- | ---- | ----
10ms(1~10) | 10 | 10
20ms(0~19) | 10 | 20
10ms(11~20) | 20 | 20
20ms(20~39) | 20 | 40

jimmy-huang:
+1

PR #519, opened 2016-12-07
OSX: use rsync since cp --parents is not avilable

poussa:
OSX do not have the cp --parents. rsync -R does the job.

Signed-off-by: Sakari Poussa <sakari.poussa@intel.com>


poussa:
@jprestwo PTAL.

grgustaf:
This learning something new every day is killing me. :) (--parents)

Any reason not to just use rsync everywhere?

grgustaf:
It works, so I'll go ahead and merge it, we can consider that later.

grgustaf:
+1

PR #522, opened 2016-12-08
[general] Use prototype inheritance throughout the API

grgustaf:
... except for UART because it has an extra complexity. The prototype
needs to also be an event emitter so we'll have to restructure things
a little.

Updated event, gpio, grove_lcd, i2c, and sensor modules though.

Also added sensor init/cleanup into ashell, and wrapped buffer calls
there with an ifdef in case it's not included.

Also some jsdoc comment format in zjs_event.h.

Also renamed zjs_init_uart to zjs_uart_init to follow pattern, but
abandoned fixing the prototype stuff in there for the moment.

### DIFF
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -34,11 +34,12 @@
 #include "file-utils.h"
 
 /* Zephyr.js init everything */
-#include "../zjs_timers.h"
 #include "../zjs_buffer.h"
 #include "../zjs_callbacks.h"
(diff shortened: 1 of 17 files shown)


grgustaf:
Let's hold off on merging this, i'll probably have a patch for the uart portion, and want to debug #524 to make sure prototype stuff isn't at fault.

grgustaf:
Accidentally deleted, I'll create a new one.

PR #523, opened 2016-12-08
[bugfix] Buffer will fail if freeing the last item

brianjjones:
Currently, Buffer will crash if it tries to free the last item,
because freeing it tries to look at the ->next of NULL. This adds
an extra check and also now returns once the handle has been freed,
since there's no reason to keep going.

### DIFF
--- a/src/zjs_buffer.c
+++ b/src/zjs_buffer.c
@@ -269,11 +269,7 @@ static void zjs_buffer_callback_free(uintptr_t handle)
             zjs_free((void *)handle);
             return;
         }
-        if ((*pItem) != NULL) {
-            pItem = &(*pItem)->next;
-        }
-        else
-            return;
+        pItem = &(*pItem)->next;
     }
 }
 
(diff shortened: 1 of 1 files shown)


grgustaf:
Not a bug,  *pItem was already tested in the while condition above. The if returns so you don't have to worry about that path.

brianjjones:
@grgustaf  this is only an issue if the last item is the one being freed.  The problem is that  *pItem = (*pItem)->next; causes *pItem to now be NULL.  The next line is pItem = &(*pItem)->next; which fails because NULL->next is not valid.  I updated this to only have the return once we have removed the handle.  Because it only happens if you remove the last element, and now we are returning, there's no need for the second check.

brianjjones:
In addition, there's no reason to keep iterating through the buffers once we free the handle, this saves time.

grgustaf:
OK, the return was missing. Yes, definitely correct now. +1

PR #525, opened 2016-12-09
[i2c] I2C functionality for non-ARC boards

brianjjones:
This is the code to add I2C support for non-ARC boards such as the
FRDM-K64.

### DIFF
--- a/src/Makefile.base
+++ b/src/Makefile.base
@@ -61,8 +61,10 @@ endif
 ifneq ($(BOARD), frdm_k64f)
 	obj-$(ZJS_AIO) += zjs_aio.o
 	obj-$(ZJS_GROVE_LCD) += zjs_grove_lcd.o
-	obj-$(ZJS_I2C) += zjs_i2c.o
+	obj-$(ZJS_I2C) += zjs_i2c_ipm.o
 	obj-$(ZJS_SENSOR) += zjs_sensor.o
+else
+	obj-$(ZJS_I2C) += zjs_i2c.o
 endif
 
 obj-$(CONFIG_BOARD_ARDUINO_101) += \
--- a/src/Makefile.base
+++ b/src/Makefile.base
@@ -58,7 +58,7 @@ obj-y += zjs_script_gen.o
 endif
 
 # skip for now for frdm_k64f
-ifneq ($(BOARD), frdm_k64f)
+ifeq ($(BOARD), arduino_101)
 	obj-$(ZJS_AIO) += zjs_aio.o
 	obj-$(ZJS_GROVE_LCD) += zjs_grove_lcd.o
 	obj-$(ZJS_I2C) += zjs_i2c_ipm.o
--- a/src/zjs_i2c.c
+++ b/src/zjs_i2c.c
@@ -273,6 +273,7 @@ static jerry_value_t zjs_i2c_open(const jerry_value_t function_obj,
         if (reply < 0) {
             ZJS_PRINT("I2C bus %s configure failed with error : %i\n", i2c_bus, reply);
         }
+
     } else {
         ZJS_PRINT("I2C bus I2C_%i is not a valid I2C bus.\n", (uint8_t)bus);
     }
--- a/src/zjs_i2c.c
+++ b/src/zjs_i2c.c
@@ -13,6 +13,8 @@
 #define ZJS_I2C_TIMEOUT_TICKS 500
 
 static struct k_sem i2c_sem;
+static jerry_value_t zjs_i2c_prototype;
+
 static struct device *i2c_device[MAX_I2C_BUS];
 
 static jerry_value_t zjs_i2c_read_base(const jerry_value_t this,
@@ -190,12 +192,12 @@ static jerry_value_t zjs_i2c_write(const jerry_value_t function_obj,
     } else {
         return zjs_error("zjs_i2c_write:  missing data buffer");
     }
-
+    int reply;
     uint32_t address = (uint32_t)jerry_get_number_value(argv[0]);
 
     if (bus < MAX_I2C_BUS && i2c_device[bus]) {
         // Write has to come after an Open I2C message
-        int reply = i2c_write(i2c_device[bus],
+        reply = i2c_write(i2c_device[bus],
                               dataBuf->buffer,
                               dataBuf->bufsize,
                               (uint16_t)address);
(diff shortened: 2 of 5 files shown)


grgustaf:
You should probably dig into that comment about mucking with the defconfig more and come up with a solution. We should be able to handle this in analyze.sh and such... just make the A101 a special case where the CONFIG_I2C gets redirected to ARC conf, but otherwise it goes into the main prj.conf.

src/Makefile.base
```diff
@@ -61,8 +61,10 @@ endif
 ifneq ($(BOARD), frdm_k64f)
 	obj-$(ZJS_AIO) += zjs_aio.o
 	obj-$(ZJS_GROVE_LCD) += zjs_grove_lcd.o
-	obj-$(ZJS_I2C) += zjs_i2c.o
+	obj-$(ZJS_I2C) += zjs_i2c_ipm.o
 	obj-$(ZJS_SENSOR) += zjs_sensor.o
+else
+	obj-$(ZJS_I2C) += zjs_i2c.o
 endif
```
grgustaf: Instead, we should make i2c_ipm conditional on A101 board, i2c for everything else. A101 is the outlier case here.

src/zjs_i2c.c
```diff
@@ -110,25 +67,44 @@ static jerry_value_t zjs_i2c_read_base(const jerry_value_t this,
...
-        send.type = TYPE_I2C_BURST_READ;
-    }
-
-    send.data.i2c.bus = (uint8_t)bus;
-    send.data.i2c.data = buf->buffer;
-    send.data.i2c.address = (uint16_t)address;
-    send.data.i2c.register_addr = register_addr;
-    send.data.i2c.length = size;
+        if (bus < MAX_I2C_BUS) {
+            // Read has to come after an Open I2C message
```
grgustaf: For consistent style, start these next couple comments with lower case. Not critical. :)

src/zjs_i2c.c
```diff
@@ -110,25 +67,44 @@ static jerry_value_t zjs_i2c_read_base(const jerry_value_t this,
...
+            if (i2c_device[bus]) {
+
```
grgustaf: remove newline

src/zjs_i2c.c
```diff
@@ -110,25 +67,44 @@ static jerry_value_t zjs_i2c_read_base(const jerry_value_t this,
...
+                int reply = i2c_read(i2c_device[bus],
+                                     buf->buffer,
+                                     size,
+                                     (uint16_t)address);
+
+                if (reply < 0) {
+                    ZJS_PRINT("I2C Read failed with error : %i\n", reply);
```
grgustaf: Looks like you should return zjs_error() here instead? (4x in the next few lines)

src/zjs_i2c.c
```diff
@@ -215,21 +193,21 @@ static jerry_value_t zjs_i2c_write(const jerry_value_t function_obj,
...
-    send.data.i2c.address = (uint16_t)address;
-    send.data.i2c.length = dataBuf->bufsize;
-
-    bool success = zjs_i2c_ipm_send_sync(&send, &reply);
-
-    if (!success) {
-        return zjs_error("zjs_i2c_write: ipm message failed or timed out!");
+    if (bus < MAX_I2C_BUS) {
+        // Write has to come after an Open I2C message
+        if (i2c_device[bus]) {
```
grgustaf: seems these two ifs could be combined into one with &&?

src/zjs_i2c.c
```diff
@@ -279,18 +257,30 @@ static jerry_value_t zjs_i2c_open(const jerry_value_t function_obj,
...
-    bool success = zjs_i2c_ipm_send_sync(&send, &reply);
-
-    if (!success) {
-        return zjs_error("zjs_i2c_write: ipm message failed or timed out!");
+    if (bus < MAX_I2C_BUS) {
+        char i2c_bus[6];
+        snprintf(i2c_bus, 6, "I2C_%i", (uint8_t)bus);
+        i2c_device[bus] = device_get_binding(i2c_bus);
+        if (!i2c_device[bus]) {
+            ZJS_PRINT("I2C bus %s not found.\n", i2c_bus);
```
grgustaf: This error could probably be combined with the "not a valid bus" one below, combine these two if's - like above.

src/zjs_i2c.c
```diff
@@ -279,18 +257,30 @@ static jerry_value_t zjs_i2c_open(const jerry_value_t function_obj,
...
+            k_sleep(10000);
```
grgustaf: What is this about?

src/zjs_i2c_ipm.c
```diff
@@ -0,0 +1,321 @@
...
+// Zephyr includes
+#include <i2c.h>
+#include <string.h>
+
+// ZJS includes
+#include "zjs_i2c.h"
+#include "zjs_ipm.h"
+#include "zjs_util.h"
+#include "zjs_buffer.h"
+
```
grgustaf: How about add a block comment here about why we have separate implementation there - that it's specific to A101 w/ separate ARC core, etc.

src/zjs_i2c.c
brianjjones: Its about me forgetting that from testing :)

brianjjones:
Updated with requested changes

src/zjs_i2c.c
grgustaf: It throws an exception. If they catch the exception, it won't terminate. I guess ultimately it will depend on how Zoltan has defined the API.

But I don't think you're doing anything here to let them know there was an error programmatically. How will they know and read again?

brianjjones: They don't currently, but this isn't new.  Setting up something to handle read / write fails is on the TODO.

src/zjs_i2c.c
```diff
@@ -207,8 +209,8 @@ static jerry_value_t zjs_i2c_write(const jerry_value_t function_obj,
     else {
         ZJS_PRINT("no I2C device is ready yet\n");
     }
-
-    return ZJS_UNDEFINED;
+    return (jerry_value_t)reply;
```
brianjjones: hmm didn't mean to check that in sorry, will fix

brianjjones: And no the return number is either 0 for success or negative for error, its not the number of bytes.

src/zjs_i2c.c
brianjjones: Added a return value so javascript can now determine if the write failed and call it again if they wish.

grgustaf: I'm not going to quibble more about this right now, not that important. But having a fresh think about it, this is definitely the kind of case where an exception is the right answer. Why should the read/write ever fail? There's no good reason, so it should be quite "exceptional" when it happens. Not something where we expect them to write a loop retrying because it may just fail any old time you call it.

grgustaf:
+1

PR #526, opened 2016-12-09
[makefile] Explicitly remove libjerry-core.a on "make clean".

pfalcon:
We put it there in "jerryscript" target, so there's nobody else but us
to remove it from there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,6 +158,7 @@ endif
 clean: update
 	@if [ -d $(ZEPHYR_SDK_INSTALL_DIR) ]; then \
 		rm -rf $(JERRY_BASE)/build/$(BOARD)/; \
+		rm -f outdir/$(BOARD)/libjerry-core.a; \
 		make -f Makefile.zephyr clean BOARD=$(BOARD); \
 		cd arc/; make clean; \
 	fi


pfalcon:
This came up in IRC discussion with @grgustaf . This doesn't affect the build process in any way - there's no dependency on libjerry-core.a, so it's re-made (re-copied) unconditionally as part of phony "jerryscript" target. But doing it like the patch does is "just right" (and if build process will ever be optimized to treat libjerry-core.a as a dependency, it will  make a difference).

grgustaf:
I just pushed #527 which seems to accomplish the same thing a different way, along with other fixes. Could you see if that works for you @pfalcon? Is what I'm doing to the external JERRY_BASE directory okay? Seems more correct.

pfalcon:
@grgustaf : Yes, that looks good. But I'd say this patch is still needed to be "fully correct". That's because there's no warranty that "make -f Makefile.zephyr clean" would remove any file not created by it. Feel free to push your change first, I'll rebase mine afterwards.

grgustaf:
@pfalcon if you can rebase this I'll merge it.

+1

pfalcon:
@grgustaf : Done.

PR #527, opened 2016-12-09
[makefile] Fix make clean and stop hardcoding deps/ as the path

grgustaf:


jimmy-huang:
+1

PR #532, opened 2016-12-13
[Grove LCD] Make Grove LCD work on the FRDM-K64F boards

jimmy-huang:
Create a non-ipm version of the Grove LCD that will build
and run on the Zephyr boards that do not have ARC.  Move all
existing IPM code to a new module zjs_grove_lcd_ipm that gets
build only for Arduino 101.


grgustaf:
+1

PR #533, opened 2016-12-13
[Snapshot] Update JerryScript

jimmy-huang:
Update JerryScript to support building without parser, and
enable snapshot by default.


grgustaf:
+1

PR #539, opened 2016-12-14
[trlite] Preserve cached (staged) changes in trlite builds

grgustaf:
Changes that you have already staged with git-add were not being applied
to the trlite clone of the repo, so you would get unexpected results
running trlite in that state.


jimmy-huang:
+1

PR #540, opened 2016-12-14
Changed UART docs to reflect proper usage

jprestwo:
- Documentation said that UART could use tty1, which was not true
   and would cause an error to be thrown. The error will still be
   thrown, but the docs now only mention tty0.


jimmy-huang:
+1

PR #541, opened 2016-12-14
[OCF] Make OCF only build on linux.

jprestwo:
- iotivity-constrained broke the build on Mac. This patch only includes
   the OCF API's on Linux


grgustaf:
+1

PR #542, opened 2016-12-14
[general] Added exit features to linux target

jprestwo:
- Passing in "--autoexit" will cause jslinux to terminate if no
   callbacks or timers were processed on the current or previous
   main loop.

 - Passing in "-t <time>" will cause jslinux to terminate after
   <time> milliseconds>.

 - Both new flags can be used at the same time if desired.

### DIFF
--- a/src/zjs_timers.h
+++ b/src/zjs_timers.h
@@ -3,7 +3,13 @@
 #ifndef __zjs_timers_h__
 #define __zjs_timers_h__
 
-void zjs_timers_process_events();
+/*
+ * Service the timer module.
+ *
+ * @return          1 if any timers were serviced
+ *                  0 if no timers were serviced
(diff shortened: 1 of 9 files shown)


src/zjs_timers.h
```diff
@@ -3,7 +3,13 @@
 #ifndef __zjs_timers_h__
 #define __zjs_timers_h__
 
-void zjs_timers_process_events();
+/*
```
grgustaf: For JSDoc, you're supposed to have a second star on the first line: `/**`. Not that we're actually parsing any of this yet anyway, but just in preparation.

jprestwo: fixed, and in a few other places also

grgustaf:
+1

Issue #504, opened 2016-12-06
[REG][Sensor] Catch errors when running Accelerometer.js 


xiaolongx:
Catch errors when running [Accelerometer.js](https://github.com/01org/zephyr.js/blob/master/samples/Accelerometer.js), this issue was fixed with #483, but reproduce again.

Error logs
```
acceleration (m/s^2):  x=0.004788 y=0.004788 z=0.004788
acceleration (m/s^2):  x=0.004788 y=0[ERROR] /home/xiaolong/00_xiaolong/ZJS/ww50/zephyr.js/src/zjs_callbacks.c:332 zjs_signal_callback(): error putting into ring buffer, ret=4294967174
S/ww50/zephyr.js/sracceleration (m/s^2):  x=0.004788 y=0.004788 z[ERROR] /home/xiaolong/00_xiaolong/ZJS/ww50/zephyr.js/src/zjs_callbacks.c:332 zjs_signal_callback(): error putting into ring buffer, ret=4294967174
phyr.js/sracceleration (m/s^2):  x=0.004788 y=0.004788 z=0[ERROR] /home/xiaolong/00_xiaolong/ZJS/ww50/zephyr.js/src/zjs_callbacks.c:332 zjs_signal_callback(): error putting into ring buffer, ret=4294967174
[ERROR] /home/xiaolong/00_xiaolong/ZJS/ww50/zephyr.js/src/zjs_callbacks.c:332 zjs_signal_callback(): error putting into ring buffer, ret=4294967174
```

Branch | Build ID | Date | Result
---|---|---|---
master | 870d928 | Nov 15, 2016 | FAIL
master | 103ca87 | Nov 30, 2016 | PASS
master | a162db1 | Dec 6, 2016 | FAIL

jimmy-huang:
this bug appears to be a completely different issue than last one, this one seems to have been introduced by recent upgrade to the unified kernel in Zephyr, I just tried the acceleometer sample in zephyr master, and it appears it's completely broken as well, notice that it report 0.004788 on all axis, which is basically 0, 0, 0 in the zephyr sample.  I think we'll have to wait for zephyr to fix the bmi160 driver first, see https://jira.zephyrproject.org/browse/ZEP-1358

xiaolongx:
No error messages display on Arduino 101, but it report 0.004788 on all axis.
Serial logs:
```
Accelerometer test...
ipm_console0: 'Sensor core running ZJS ARC support image'
state: activating
state: activated
[...]

PR #543, opened 2016-12-14
[Sensor] Updated BMI160 configs

jimmy-huang:
This will fix the error flooding Accelerometer events in
bug #504, although it doesn't fix the bug where Accelerometer
will report 0 on all axis, that's a separate bug.


poussa:
LGTM. 

Fixes the flooding but incorrect values still due to the Zephyr bug.

PR #545, opened 2016-12-15
[modules] Fix warnings from recent update to modules code

grgustaf:
We've turned off -Werror for a little while so it's harder to notice
when this happens.


brianjjones:
+1 lgtm

PR #546, opened 2016-12-15
[trlite] Preserve staged AND unchanged changes in trlite

grgustaf:
My last change included staged changes, but accidentally dropped
unstaged changes. Now finally we'll have both.

### DIFF
--- a/scripts/trlite
+++ b/scripts/trlite
@@ -19,7 +19,6 @@ if [ ! -d "$ZJS_BASE" ]; then
 fi
 
 TRLDIR=$ZJS_BASE/.trlite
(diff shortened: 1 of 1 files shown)


grgustaf:
@brianjjones if you could sanity check this for me. Make a changes to two files. Do `git add` on one of the changes, not on the other. So you should then see one change with `git diff` and the other with `git diff --cached`. Then run trlite. Break out of it once it actually starts building. Then cd into .trlite (it's basically a copy of your build tree) and manually view the two files you changed. Both code changes should be there.

brianjjones:
Works for me +1

PR #553, opened 2016-12-15
Console time apis

jprestwo:


src/zjs_console.c
```diff
@@ -152,6 +159,82 @@ static jerry_value_t console_error(const jerry_value_t function_obj,
...
+                                  const jerry_value_t this,
+                                  const jerry_value_t argv[],
+                                  const jerry_length_t argc)
+{
+    if (!jerry_value_is_string(argv[0])) {
+        ERR_PRINT("invalid parameters\n");
+        return ZJS_UNDEFINED;
+    }
+    int sz = jerry_get_string_size(argv[0]);
+    char label[sz + 1];
```
grgustaf: You've got to watch the limit the maximum size of the string here, as usual, or we can overflow the stack.

src/zjs_console.c
```diff
@@ -152,18 +160,167 @@ static jerry_value_t console_error(const jerry_value_t function_obj,
...
+
+    jerry_value_t new_timer_obj = jerry_create_object();
+    zjs_port_timer_t* timer = zjs_malloc(sizeof(zjs_port_timer_t));
+    if (!timer) {
+        ERR_PRINT("could not allocate timer object\n");
+        return ZJS_UNDEFINED;
+    }
+    zjs_port_timer_init(timer);
+    // set timer to go off every millisecond (this is the best accuracy we can get)
+    zjs_port_timer_start(timer, 1);
```
grgustaf: We shouldn't be making a timer here. Just record the start time in a structure associated with the label, and calculate the end time later by subtracting this from the then-current time. That's how the performance module is intended to be used, so let's also consider whether we really need both? Please review #325 to see how it relates to this. Offhand I don't see console.time() mentioned so maybe that's a good model for us that @pfalcon didn't consider. Let's make sure he comments before we go forward with this.

src/zjs_console.c
```diff
@@ -152,18 +160,167 @@ static jerry_value_t console_error(const jerry_value_t function_obj,
...
+{
+    if (!jerry_value_is_object(console_prototype_obj)) {
+        ERR_PRINT("console not initialized\n");
+        return ZJS_UNDEFINED;
+    }
+    jerry_value_t console = jerry_create_object();
+    jerry_set_prototype(console, console_prototype_obj);
+
+    return console;
+}
```
grgustaf: Could you explain what this buys us? This is like the constructor in node.js?

jprestwo: Yes? The node.js API I looked at had a Console() function where you could create a "new" console object... seems unnecessary though. If its something we can just not conform to I am fine removing it.

src/zjs_console.c
```diff
@@ -297,13 +299,22 @@ void zjs_console_init(void)
...
+        { console_time, "time" },
+        { console_time_end, "timeEnd" },
+        { console_assert, "assert" },
+        { NULL, NULL }
+    };
+
+    console_prototype_obj = jerry_create_object();
+    zjs_obj_add_functions(console_prototype_obj, array);
+
+    jerry_set_prototype(console, console_prototype_obj);
```
jprestwo: Good point, I was going off of probably very new Node.js documentation which had the ability to create a new Console object (`new Console()`) but it seems unnecessary and not really needed.

src/zjs_console.c
jprestwo: Yes your right, Zephyr has k_get_uptime() or something like that. Will change in my next push.

grgustaf:
@jprestwo You'll need to replace to_char_buffer calls with the new zjs_copy_jstring and/or zjs_alloc_from_jstring in utils. This will be a good test of whether they make sense to others. :)

The point of them is to force you to specify a maximum length and not get bitten by stack overflow or missing null terminators.

src/zjs_console.c
```diff
@@ -170,6 +186,10 @@ static jerry_value_t console_time(const jerry_value_t function_obj,
         return ZJS_UNDEFINED;
     }
     int sz = jerry_get_string_size(argv[0]);
+    if (sz > MAX_STR_LENGTH) {
+        ERR_PRINT("label sting too long\n");
```
grgustaf: This petty comment is going to sting. (Typo!)

You'll check return value from zjs_copy_jstring instead here.

src/zjs_console.c
```diff
@@ -205,6 +223,10 @@ static jerry_value_t console_time_end(const jerry_value_t function_obj,
...
```
grgustaf: Same typo but again you'll be updating this anyway.

grgustaf:
+1

grgustaf:
@pfalcon, could you comment on this? Apparently it gives us the same functionality as performance module but comes from Node APIs. What do you think about it?

pfalcon:
Sorry for the delay with response, just getting back from NY holidays (Happy New Year everyone!), and have a backlog of stuff.

Well, so when I proposed to add timing functionality to Z.js, I surveyed a possible ways to do that, and asked for feedback: https://github.com/01org/zephyr.js/issues/325 . "performance" module was selected because it was simple and exactly to the point, and fit well with Z.js approach of "if you don't import a module, it's compiled out".

As can be seen, Console.time() wasn't even in the list in #325, and reading .time()/.timeEnd(), it's not really a match for performance.now(). Reading https://developer.mozilla.org/en-US/docs/Web/API/Console/timeEnd, "Once stopped, the elapsed time is automatically displayed in the Web Console." I.e., it's very convenient development/debugging function, which allows to quickly instrument code to get execution duration between any 2 points of code (including multiple such pairs). But it doesn't allow to get current time/duration programmatically, like performance.now() does.

Another issue hinted above is that "performance" module, if not used, won't be compiled into firmware. But unless Z.js recently acquired ability to filter out individual functions from modules, or methods from objects, then Console.time()/timeEnd() will always be there. So actually, I'd suggest to add and #ifdef to remove them from a "production" firmware build, where they're supposedly not useful.

(Here goes the usual disclaimer is that I'm not regular-time JavaScript programmer, and don't have "intuitive" picture of its APIs, so may miss something. Hence the detailed answer, so more experienced JS programmers may see if I miss something. Thanks.)

grgustaf:
@jprestwo okay, sounds like this is a fine thing to add as it serves a little different purpose. However, I wonder if we'd be better to implement it by saving the time in JS objects instead of using our precious few malloc opportunities.

grgustaf:
Reiterating my +1, we can look into that as a refinement.

Issue #551, opened 2016-12-15
[I2C] Fail to throw error when opening an invalid bus on FRDM-K64F


xiaolongx:
Flashing [test-i2c.js](https://github.com/01org/zephyr.js/blob/master/tests/test-i2c.js) on FRDM-K64F, fail to throw error when opening an invalid bus.
```
expectThrow("I2C: open invalid bus", function () {
    i2c.open({ bus: 1, speed: 100 });
});
```

Error logs:
```
I2C bus I2C_1 is not a valid I2C bus.
FAIL - I2C: open invalid bus
```

Branch | Build ID | Target Device | Date | Result
---|---|---|---|---
master | 5c969a5 | Arduino 101 | Dec 15, 2016 | PASS
master | 5c969a5 | FRDM-K64F | Dec 15, 2016 | FAIL

grgustaf:
I'm not sure if this a regression or it never threw an error? This may come down to the expectation that JS will throw for unlikely errors. It may well be fixed just by moving to the new APIs, since they probably specify this behavior more clearly.

brianjjones:
On the Arduino 101 bus 1 is valid (at least in theory I haven't tried in practice)

I'll check this out today and if it is indeed invalid I'll change it from only printing an error to throwing

brianjjones:
This issue will be fixed by https://github.com/01org/zephyr.js/pull/554

xiaolongx:
This issue cannot reproduce on FRDM-K64F with commit 2345f96, close

PR #554, opened 2016-12-15
[bugfix] I2C will now throw an error if bus fails to open

brianjjones:
This is a fix for issue #551, now an error will be thrown if the
I2C bus fails to open.

### DIFF
--- a/src/zjs_i2c_ipm.c
+++ b/src/zjs_i2c_ipm.c
@@ -198,10 +198,12 @@ static jerry_value_t zjs_i2c_write(const jerry_value_t function_obj,
 
     uint32_t register_addr = 0;
 
-    if (argc >= 3 && !jerry_value_is_number(argv[2])) {
-        return zjs_error("zjs_i2c_read: register is not a number");
-    } else {
-        register_addr = (uint32_t)jerry_get_number_value(argv[2]);
+    if (argc >= 3) {
+        if (!jerry_value_is_number(argv[2])) {
+            return zjs_error("zjs_i2c_read: register is not a number");
+        } else {
+            register_addr = (uint32_t)jerry_get_number_value(argv[2]);
+        }
     }
 
     uint32_t bus;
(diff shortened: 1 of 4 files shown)


src/zjs_i2c_ipm.c
```diff
@@ -294,20 +294,24 @@ static jerry_value_t zjs_i2c_open(const jerry_value_t function_obj,
...
+    if (reply.error_code == 0) {
+        // create the I2C object
+        jerry_value_t i2c_obj = jerry_create_object();
+        zjs_obj_add_function(i2c_obj, zjs_i2c_read, "read");
+        zjs_obj_add_function(i2c_obj, zjs_i2c_burst_read, "burstRead");
+        zjs_obj_add_function(i2c_obj, zjs_i2c_write, "write");
+        zjs_obj_add_function(i2c_obj, zjs_i2c_abort, "abort");
+        zjs_obj_add_function(i2c_obj, zjs_i2c_close, "close");
+        zjs_obj_add_number(i2c_obj, bus, "bus");
+        zjs_obj_add_number(i2c_obj, speed, "speed");
```
grgustaf: This should use a prototype since there can be more than one of these i2c_obj's created.

See how prototype was added to grove_lcd for example with this command:
`git show 81b7fcb src/zjs_grove_lcd.c src/zjs_modules.c`

(In zjs_modules.c, just pay attention where the cleanup function was added.)

Otherwise, LGTM.

grgustaf:
@brianjjones: Oh we probably crossed the streams when you were working on this and I was doing the prototype stuff. So make sure you fix that in both the IPM and non-IPM versions if need be.

brianjjones:
@grgustaf I've added the fixes needed for both versions to use the prototype, as well as a couple other minor fixes.  It should be good to go now, i've thoroughly tested on both platforms.  It will requires a couple of merge fixes, I can't do this as I don't have write access.  Please note that in arc/src/main.c, for the ERR_PRINT msg_bus is an uint not a string, so use my version rather than master.  And in src/zjs_i2c.c, keep the return zjs_error I added to line 284.  If you have any questions let me know.

grgustaf:
+1

Issue #549, opened 2016-12-15
[sensor] Error thrown when import custom module or large segment code


cuiyanx:
When add a large segment code or import custom module [BMP280.js](https://github.com/01org/zephyr.js/blob/master/modules/BMP280.js) in samples/Accelerometer.js, throw out error.

error log:
```
zjs_sensor_ipm_send_sync: FATAL ERROR, ipm timed out
zjs_sensor_create failed to init
```

grgustaf:
@brianjjones, might be wise to git bisect this one and see where it broke. Might have been one of @jimmy-huang's recent patches or my fault again.

brianjjones:
@grgustaf This is a race condition.  By adding the BMP280.js, we now have two things trying to init at the same time.  Apparently IPM can't handle this currently, and the sensor IPM times out.  If you wrap the entire Accelerometer.js code with a setTimeout for a second, it works fine.  This may just be a limitation of the IPM system, but perhaps we can do a better job with scheduling them.  I'll dig in more and let you know.

brianjjones:
This is solved by https://github.com/01org/zephyr.js/pull/555

xiaolongx:
This issue cannot reproduce on Arduino 101 with commit 5a25755

grgustaf:
Closing, thanks.

PR #555, opened 2016-12-16
[bugfix] Extending the IPM timeout

brianjjones:
Fix for issue #549.  This was caused by having too short of a IPM
timeout.  Extending the timeout solves the issue, allowing multiple
modules to be using IPM without issue.


grgustaf:
+1

PR #556, opened 2016-12-17
[util] Add helper functions to reduce bugs with JrS strings

grgustaf:
We have to be pretty careful with JerryScript strings to avoid using
too much stack space for a big string, make sure strings are null
terminated, etc. It makes sense to handle these issues in a single
place. This patch adds two new functions for copying JrS strings to
a buffer on the stack or allocating one on the heap (which must be
freed).

### DIFF
--- a/samples/tests/Console.js
+++ b/samples/tests/Console.js
@@ -18,11 +18,15 @@ console.error("  Objects:", {}, {foo: 'bar'});
 console.error("  Strings:", '', "", 'abc', "def");  // expect: "  abc def"
 console.error("     Misc:", null, undefined);
 
-console.log("\nTesting long string...");
-var toolong =
+console.log("\nTesting a long string...");
+var longstr =
     "123456789012345678901234567890123456789012345678901234567890" +
     "123456789012345678901234567890123456789012345678901234567890" +
     "123456789012345678901234567890123456789012345678901234567890" +
     "123456789012345678901234567890123456789012345678901234567890" +
-    "12345678901234567";  // length 257
-console.log(toolong);
+    "123456789012345";  // length 255
+console.log(longstr);
+
+console.log("\nTesting a string too long...");
+longstr += '6';  // length 256
+console.log(longstr);
--- a/scripts/analyze.sh
+++ b/scripts/analyze.sh
@@ -110,6 +110,8 @@ if check_for_require ocf || check_config_file ZJS_OCF; then
     echo "CONFIG_IP_BUF_TX_SIZE=5" >> prj.conf.tmp
     echo "CONFIG_NET_MAX_CONTEXTS=9" >> prj.conf.tmp
     echo "export ZJS_OCF=y" >> zjs.conf.tmp
+    MODULES+=" -DBUILD_MODULE_EVENTS"
+    echo "export ZJS_EVENTS=y" >> zjs.conf.tmp
 fi
 
 if check_for_require gpio || check_config_file ZJS_GPIO; then
--- a/src/zjs_event.h
+++ b/src/zjs_event.h
@@ -1,3 +1,5 @@
+// Copyright (c) 2016, Intel Corporation.
+
 #ifndef __zjs_event_h__
 #define __zjs_event_h__
 
--- a/src/zjs_uart.c
+++ b/src/zjs_uart.c
@@ -236,8 +236,12 @@ static jerry_value_t uart_init(const jerry_value_t function_obj,
 
     jerry_value_t port_val = zjs_get_property(argv[0], "port");
 
-    int sz = jerry_get_string_size(port_val);
-    if (sz > 16) {
+    const int MAX_PORT_LENGTH = 16;
+    jerry_size_t size = MAX_PORT_LENGTH;
+    char port[size];
+
+    zjs_copy_jstring(port_val, port, &size);
+    if (!size) {
         DBG_PRINT("port length is too long\n");
         jerry_value_t error = make_uart_error("TypeMismatchError",
                 "port length is too long");
@@ -245,19 +249,7 @@ static jerry_value_t uart_init(const jerry_value_t function_obj,
         return promise;
     }
 
-    char port[sz + 1];
-    int len = jerry_string_to_char_buffer(port_val, (jerry_char_t *)port, sz);
-    port[len] = '\0';
-
     jerry_release_value(port_val);
-
-    if (sz != len) {
-        DBG_PRINT("size mismatch, sz=%u, len=%u\n", sz, len);
-        jerry_value_t error = make_uart_error("TypeMismatchError",
-                "size mismatch");
-        zjs_reject_promise(promise, &error, 1);
-        return promise;
-    }
 
     jerry_value_t baud_val = zjs_get_property(argv[0], "baud");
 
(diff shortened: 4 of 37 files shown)


grgustaf:
@brianjjones, @jprestwo - I'd appreciate some review of this before I get all the way done. But I'm hoping these two new functions will replace all our use of the jerry_string_to_char_buffer function. In fact, I then plan to check in Travis to make sure no one calls that "naked" again. Anyway, so far converted about 1/5 of the calls to use them.

src/zjs_util.c
```diff
@@ -180,6 +180,42 @@ bool zjs_obj_get_int32(jerry_value_t obj, const char *name, int32_t *num)
...
+    if (maxlen && size <= *maxlen) {
+        len = jerry_string_to_char_buffer(jstr, (jerry_char_t *)buffer, size);
+    }
+    buffer[len] = '\0';
+}
+
+char *zjs_alloc_from_jstring(jerry_value_t jstr, jerry_size_t *maxlen)
+{
+    jerry_size_t size = jerry_get_string_size(jstr);
+    char *buffer = zjs_malloc(size + 1);
```
brianjjones: If we have maxlen, can't we return at this point if size > maxlen?

brianjjones:
In general it LGTM, see my one comment - I think this is a good idea since its an obvious possibility for memory leaks and buffer overflows.  This will help keep us from breaking as well as save a few lines of code.

src/zjs_util.c
grgustaf: No, what this function will do is allocate the full string but then truncate it to the maxlen and still return successfully. I wrote it to replace code in grove_lcd.c and that's what that code was doing.

src/zjs_util.h
```diff
@@ -56,6 +56,35 @@ bool zjs_obj_get_double(jerry_value_t obj, const char *name, double *num);
...
+void zjs_copy_jstring(jerry_value_t jstr, char *buffer, jerry_size_t *maxlen);
+
+/**
+ * Allocate a char * buffer on the heap and copy the JerryScript string to it.
+ *
+ * @param jstr    A JerryScript string value.
+ * @param maxlen  Pointer to a maximum size for the returned string. The buffer
+ *                  allocated may be larger, but the string will be no larger
+ *                  than this (not including null terminator which is guaranteed
+ *                  to be there). If NULL or pointing to 0, there is no limit
```
grgustaf: @brianjjones, this second sentence in the description tries to explain that.

src/zjs_util.h
```diff
@@ -56,6 +56,35 @@ bool zjs_obj_get_double(jerry_value_t obj, const char *name, double *num);
 bool zjs_obj_get_uint32(jerry_value_t obj, const char *name, uint32_t *num);
 bool zjs_obj_get_int32(jerry_value_t obj, const char *name, int32_t *num);
 
+/**
+ * Copy a JerryScript string into a supplied char * buffer.
+ *
+ * @param jstr    A JerryScript string value.
+ * @param buffer  A char * buffer with at least maxlen + 1 bytes.
+ * @param maxlen  Pointer to a maximum size for the copied string. The buffer
+ *                  allocated may be larger, but the string will be no larger
```
grgustaf: I should change this 'allocated' to supplied though.

grgustaf:
There are still a few straggling locations of calls to jerry_string_to_char_buffer I will replace, plus the macro ZJS_GET_STRING which doesn't enforce a max limit on strings and needs to be changed.

grgustaf:
@brianjjones @jprestwo I think this is ready for a final review if you're up for it.

samples/tests/Console.js
```diff
@@ -18,11 +18,15 @@ console.error("  Objects:", {}, {foo: 'bar'});
...
     "123456789012345678901234567890123456789012345678901234567890" +
     "123456789012345678901234567890123456789012345678901234567890" +
-    "12345678901234567";  // length 257
-console.log(toolong);
+    "123456789012345";  // length 255
+console.log(longstr);
+
+console.log("\nTesting a string too long...");
+longstr += '6';  // length 256
+console.log(longstr);
```
grgustaf: Note I changed the semantics here so the longest string length of 256 now includes the null terminator, so the strlen is 255. I decided to test both the 255 and 256 case, where before I only tested the failing 257 case.

scripts/analyze.sh
```diff
@@ -110,6 +110,8 @@ if check_for_require ocf || check_config_file ZJS_OCF; then
     echo "CONFIG_IP_BUF_TX_SIZE=5" >> prj.conf.tmp
     echo "CONFIG_NET_MAX_CONTEXTS=9" >> prj.conf.tmp
     echo "export ZJS_OCF=y" >> zjs.conf.tmp
+    MODULES+=" -DBUILD_MODULE_EVENTS"
+    echo "export ZJS_EVENTS=y" >> zjs.conf.tmp
```
grgustaf: FYI this is a bug fix hidden in the patch, perhaps I should pull it out into its own before i merge it.

src/zjs_event.h
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) 2016, Intel Corporation.
+
```
grgustaf: missing copyrights could be separated as well and i could check to make sure i got all of them i guess

src/zjs_uart.c
```diff
@@ -236,35 +236,28 @@ static jerry_value_t uart_init(const jerry_value_t function_obj,
...
-        return promise;
-    }
-
     jerry_value_t baud_val = zjs_get_property(argv[0], "baud");
 
     if (jerry_value_is_number(baud_val)) {
         baud = (int)jerry_get_number_value(baud_val);
     }
     for (i = 0; i < (sizeof(device_map) / sizeof(device_map[0])); ++i) {
+        // FIXME: we allowed 16-char string above but are only looking at 4?
```
grgustaf: @jprestwo see this comment, not intending to address in this patch though

src/zjs_util.h
brianjjones: I personally think you should specifically call out that if jstr is longer than maxlen, it will be truncated.  I think it would help make it clearer.

src/zjs_util.c
```diff
@@ -180,6 +177,41 @@ bool zjs_obj_get_int32(jerry_value_t obj, const char *name, int32_t *num)
     return true;
 }
 
+void zjs_copy_jstring(jerry_value_t jstr, char *buffer, jerry_size_t *maxlen)
+{
+    jerry_size_t size = jerry_get_string_size(jstr);
```
grgustaf: Actually this is okay. The contract is, don't call this with NULL. That's specified in the function description. `@param buffer  A char * buffer with at least *maxlen bytes.`

This is the "stack" version so it should be called on a stack pointer that could never be NULL. So I think it's reasonable to leave this requirement on the caller.

grgustaf: (For example, strcpy doesn't check for NULL - it's the caller's fault if they do that.)

src/zjs_util.h
grgustaf: You're right, I'll make that clearer.

grgustaf:
@brianjjones - rebased and pulled out distinct fixes

brianjjones:
LGTM +1

PR #561, opened 2016-12-20
[events] Fixed and updated event module

jprestwo:
- Fixed return values of some functions to be correct and consistent
 - Changed DBG_PRINT's to ERR_PRINT's when appropriate


grgustaf:
LGTM, I haven't thought through the logic of what's being returned in each function but assume James did. It seems like a good cleanup of inconsistencies.

PR #564, opened 2016-12-20
[OCF] Updated Zephyr and iotivity-constrained to latest

jprestwo:

### DIFF
--- a/src/zjs_promise.c
+++ b/src/zjs_promise.c
@@ -46,13 +46,6 @@ static void post_promise(void* h, jerry_value_t* ret_val)
         jerry_release_value(handle->then);
         jerry_release_value(handle->catch);
         jerry_release_value(handle->this);
-    }
-}
-
-static void promise_free(const uintptr_t native)
-{
-    struct promise* handle = (struct promise*)native;
-    if (handle) {
         zjs_free(handle);
     }
 }
@@ -113,7 +106,7 @@ void zjs_make_promise(jerry_value_t obj, zjs_post_promise_func post,
 
     zjs_obj_add_function(obj, promise_then, "then");
     zjs_obj_add_function(obj, promise_catch, "catch");
-    jerry_set_object_native_handle(promise_obj, (uintptr_t)new, promise_free);
+    jerry_set_object_native_handle(promise_obj, (uintptr_t)new, NULL);
 
     new->user_handle = handle;
     new->post = post;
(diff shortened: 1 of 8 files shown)


jprestwo:
I was just informed that there seems to be a problem with QEMU Zephyr and the IP stack, specifically on Ubuntu 16.04. Apparently versions less than 16.04 work fine with QEMU and the new IP stack. I do not have an earlier version of Ubuntu, hence this patch does not run for me, but if someone does have Ubuntu 15.04 or lower and wishes to give this PR a try that would be great. It requires a few steps to run QEMU with the IP stack:

 - Read `deps/iotivity-constrained/README` to find how to get "net-tools". Its a git repo in gerrit.zephyrproject.org
 - Using 2 terminals, run `./net-tools/loop-socat.sh`, then `sudo ./net-tools/loop-slip-tap.sh`. This will create the network interface QEMU uses.
 - Once these are running, you can start QEMU with an OCF sample e.g. `make BOARD=qemu_x86 JS=samples/OcfServer.js qemu`
 - Then from another terminal you could run the other OCF sample under linux e.g. `./outdir/linux/release/jslinux samples/OcfClientUpdate.js`. Of course running `make linux` first if you haven't already.

The issue with 16.04 is or will hopefully be getting looked into by the Zephyr team as soon as possible.

grgustaf:
Yeah I should fork 0.2 before we do that. I'll hopefully do that tomorrow to keep my word. :)

grgustaf:
@jprestwo maybe you should set up virtualbox with 14.04? I do still have a real box though if you want me to try something.

jprestwo:
@grgustaf, if you have an older version its not too difficult to test this on, but more involved than just running something. You need another repo for those other tools. If we want this in 0.2 then tomorrow I can create a virtual machine and try it out to make sure it works.

src/zjs_promise.c
```diff
@@ -46,13 +46,6 @@ static void post_promise(void* h, jerry_value_t* ret_val)
...
         jerry_release_value(handle->catch);
         jerry_release_value(handle->this);
-    }
-}
-
-static void promise_free(const uintptr_t native)
-{
-    struct promise* handle = (struct promise*)native;
-    if (handle) {
         zjs_free(handle);
```
grgustaf: Seems like maybe you should also remove the object from the "promise" property of the parent object? Like set that to undefined? Also maybe that "promise" property should be "hidden", however you did that magic? Please look into those questions but I won't hold up the patch for it.

grgustaf:
+1, expect @jprestwo to update to newer Zephyr with more changes soon.

PR #567, opened 2016-12-21
[travis] Disallow new calls to jerry_string_to_char_buffer

grgustaf:
This is to force people to use the new zjs_copy_jstring and
zjs_alloc_from_jstring. These calls require a maximum size to avoid
overflowing the stack or missing a null terminator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,10 +24,6 @@ env:
     - TARGET="linux"
     - TARGET="ashell"
 
-matrix:
-  allow_failures:
-    - env: TARGET="ashell"
-
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


grgustaf:
I expect this to fail initially until I merge #556.

.travis.yml
```diff
@@ -85,6 +81,16 @@ script: >
     banner "git check" &&
     git diff --check $(git rev-list HEAD | tail -1) &&
 
+    # ensure only two uses of jerry_string_to_char_buffer in zjs_util.c
```
brianjjones: In the future we'll want to update this so we can show the trlite error message.  Otherwise it won't be clear why the build failed to users unfamiliar with our rule set.

brianjjones:
LGTM +1

PR #568, opened 2016-12-22
[general] Remove unnecessary function names from error messages

grgustaf:
ERR_PRINT and DBG_PRINT automatically display the function, file, and
line number so the function name should not be in the message string.


grgustaf:
I think let's just leave that alone for the moment, it's actually a little hard to tell what's really an important user-reportable error and what might be just a valid condition. Maybe @jimmy-huang can review them later.

brianjjones:
LGTM +1

PR #570, opened 2016-12-22
[docs] Event API documentation

jprestwo:

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,0 +1,123 @@
+ZJS API for Events
+==================
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+ZJS provides event API's which match Node.js Event's. We describe them here as there could
+potentially be minor differences.
+
+Web IDL
+-------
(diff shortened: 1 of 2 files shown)


docs/events.md
```diff
@@ -0,0 +1,118 @@
...
+ZJS provides event API's which match Node.js Event's. We describe them here as there could
+potentially be minor differences.
+
+Web IDL
+-------
+This IDL provides an overview of the interface; see below for documentation of
+specific API functions.
+
+```javascript
+this on(string event, function listener);
```
grgustaf: Using function like this breaks the javascript syntax highlighting on GitHub. Instead name your callback types something like "EventCallback" or "ListenerCallback" and then you can define it below so people know what it takes and returns.

See the [GPIO API](https://github.com/01org/zephyr.js/blob/master/docs/gpio.md)'s ChangeCallback for an example.

I think this might really be an EventEmitter interface that these functions should be enclosed inside? That might help it make 'this' less confusing because standalone here it's not clear what object we're talking about. Again take a look at GPIO maybe to see how that can be specified.

But that's kind of open-ended. I'll merge it with just the syntax highlighting fix if this is unclear.

grgustaf:
+1

PR #585, opened 2017-01-05
[tests] Add tests for Event APIs

cuiyanx:

### DIFF
--- a/tests/test-event.js
+++ b/tests/test-event.js
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2016-2017, Intel Corporation.
 
 // Testing EVENT APIs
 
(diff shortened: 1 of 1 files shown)


grgustaf:
@jprestwo, you should try this out. I get 10/18 pass on qemu and 11/18 pass on A101. "default number of listeners" is the extra one that works.

tests/test-event.js
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) 2016, Intel Corporation.
```
grgustaf: Make this 2016-2017

jprestwo:
This looks good to me. I got the same results as @grgustaf, so its seems it uncovers some bugs in events that I will have to fix.

grgustaf:
+1 based on jprestwo being content. :)

PR #599, opened 2017-01-13
[error] Implement standard error types

grgustaf:
These are defined by Zoltan's iot-js-api. It works to check them with
the .name property or instanceof, the latter due to lots of painful
experimentation so you better use it!

Also, update native_require_handler to use the new standard error
macros and test these along with error constructors in a new
test-error.js file.

### DIFF
--- a/src/zjs_error.c
+++ b/src/zjs_error.c
@@ -1,0 +1,102 @@
+// Copyright (c) 2016, Intel Corporation.
+
+// ZJS includes
+#include "zjs_util.h"
+
+typedef struct {
+    char *name;
+    jerry_value_t ctor;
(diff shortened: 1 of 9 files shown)


src/zjs_error.c
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016, Intel Corporation.
```
brianjjones: New year it's 2017 :)

brianjjones:
LGTM +1.  You could change the file dates to 2017 if you want.

grgustaf:
@brianjjones good point about copyrights, we should be watching those and edit them as we touch files again.

PR #613, opened 2017-01-17
[i2c] Reorganize code to make future maintenance easier

brianjjones:
Both ARC and non-ARC boards will now use the same base code.  The
IPM logic has been moved to its own file.  Now any changed needed
for I2C can address either the I2C code or the IPM code.  This
will remove the need to update two code bases with any future
changes.


src/zjs_i2c_handler.c
```diff
@@ -0,0 +1,106 @@
+#include "zjs_i2c_handler.h"
+#include "zjs_common.h"
+
+#define I2C_0 0
+#define I2C_1 1
+#define MAX_I2C_BUS 2
+
+static struct device *i2c_device[MAX_I2C_BUS];
+
+uint8_t zjs_i2c_handle_open (uint8_t msg_bus) {
```
grgustaf: Remove the space before the paren in function definitions for consistency.

src/zjs_i2c_handler.h
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2016, Intel Corporation.
+
+#ifndef __zjs_i2c_msg_h__
+#define __zjs_i2c_msg_h__
+
+#include <i2c.h>
+
+uint8_t zjs_i2c_handle_open (uint8_t msg_bus);
```
grgustaf: Here's another batch of those, remove the space before paren.

src/zjs_i2c_handler.c
```diff
@@ -0,0 +1,106 @@
+#include "zjs_i2c_handler.h"
```
grgustaf: Add copyright header and include 2016-2017 since you're pulling in code from the other file.

src/zjs_i2c_ipm_handler.c
```diff
@@ -0,0 +1,137 @@
+#include "zjs_i2c_handler.h"
```
grgustaf: Copyright header with 2016-2017.

src/zjs_i2c.c
```diff
@@ -9,11 +9,14 @@
 #include "zjs_i2c.h"
```
grgustaf: Update this and other files you've touched with copyright 2016-2017. Soon we'll have touched them all and be done. :)

src/zjs_i2c_ipm_handler.c
```diff
@@ -25,7 +26,9 @@ static bool zjs_i2c_ipm_send_sync(zjs_ipm_message_t* send,
     return true;
 }
 
-static void ipm_msg_receive_callback(void *context, uint32_t id, volatile void *data)
+static void ipm_msg_receive_callback(void *context,
+                                     uint32_t id,
+                                     volatile void *data)
```
grgustaf: I'd prefer you leave as many as will fit in 80 chars on one line.

src/zjs_i2c_ipm_handler.c
```diff
@@ -52,7 +55,10 @@ void zjs_i2c_ipm_init() {
     k_sem_init(&i2c_sem, 0, 1);
 }
 
-uint8_t zjs_i2c_handle_write (uint8_t msg_bus, uint8_t *data, uint32_t length, uint16_t address) {
+uint8_t zjs_i2c_handle_write (uint8_t msg_bus,
+                              uint8_t *data,
+                              uint32_t length,
+                              uint16_t address) {
```
grgustaf: And we wrap the open braces to the next line. Same below a couple times.

grgustaf:
@brianjjones, I think your problem is zjs_error moved from zjs_util.h to zjs_error.h, and it's now a macro. So the problem probably has to do with that. If you update your local master to upstream/master you'll probably reproduce the problem with trlite.

src/zjs_i2c_handler.h
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2016-2017, Intel Corporation.
+#ifndef __zjs_i2c_msg_h__
```
grgustaf: Change this to __zjs_i2c_handler_h__, it's meant to match the filename.

src/zjs_i2c_ipm_handler.c
```diff
@@ -0,0 +1,150 @@
...
+    send.data.i2c.data = data;
+    send.data.i2c.address = (uint16_t)address;
+    send.data.i2c.length = length;
+
+
+    // We have ARC send it via ipm
+    bool success = zjs_i2c_ipm_send_sync(&send, &reply);
+
+    if (!success) {
+        return zjs_error("zjs_i2c_write: ipm message failed or timed out!");
```
grgustaf: This returns a jerry_value_t so it's not the right thing in this function. Maybe just use ERR_PRINT? If this code is compiled by ARC it should not be using any JerryScript stuff, including zjs_error.h.

Same in below calls to zjs_error.

brianjjones: @grgustaf This is used by the x86 side not ARC.  The success is checking whether we got a reply from the ARC side. I'm fine with changing it to a ERR_PRINT.

brianjjones: @grgustaf Fixes uploaded

brianjjones:
@grgustaf This should be ready to go unless you see any other changes that need to be made

grgustaf:
+1

Issue #611, opened 2017-01-17
Measure and log boot time of key sample apps


poussa:
Our QA is measuring the boot time of some of the sample applications in order to catch improvements and regressions. We should use the performance API to measure and log the boot time of the following sample apps:

samples/HelloWorld.js
samples/I2C.js
samples/WebBluetoothGroveLcdDemo.js

The timestamp should be taken when all the modules (via require) have been loaded and all subsystems are initialized. The timestamp should be logged via console.log method.

grgustaf:
Should be addressed by #615,

qiaojingx:
Boot timestamps has been added to samples/I2C.js, samples/WebBluetoothGroveLcdDemo.js.
A new sample/BootTime.js is added instead of adding boot timestamps to HelloWorld.js.
Verified with commit [5352200](https://github.com/01org/zephyr.js/commit/5352200e9830021aaf670a1722535ac955f5a0dc)

PR #615, opened 2017-01-18
[perf] Add boot timestamps to key samples for QA measurement

grgustaf:
Add new BootTime.js sample that does nothing but this check, since
it didn't seem appropriate to add to HelloWorld.js.

Addresses issue #611.


brianjjones:
LGTM +1

grgustaf:
@hanyia, note I added a new BootTime.js to be used instead of HelloWorld.js as the minimal test case - because HelloWorld itself shouldn't be dependent on any other modules like performance.

PR #617, opened 2017-01-18
[buffer] Use native handle to track buffers and eliminate list

grgustaf:


brianjjones:
LGTM +1

PR #618, opened 2017-01-19
[console] Simplify time/timeEnd implementation with JS object

grgustaf:
Works without malloc or adding a JS object for each timer.
Also, add cleanup function to release the global time object.


jprestwo:
This all looks good to me. +1

Issue #623, opened 2017-01-19
[console] Throw out wrong error


cuiyanx:
Set timer name as "", throw out wrong error.

Code:
```
var Timer = "";
console.time(Timer);
console.timeEnd(Timer);
```
Error:
```
src/zjs_console.c:223 console_time_end():
[ERROR] string is too long
```

grgustaf:
This is probably fixed by my PR #618

grgustaf:
Now it will actually work with any empty string label, no error.

cuiyanx:
Yes, it is no error throwing out, but it will be hung. And when I build this, throw out a 'warn'. Please check this.

cuiyanx:
fixed and closed

Issue #622, opened 2017-01-19
[console] Not support constructor to create


cuiyanx:
Node.js v7.4.0：
`Creates a new Console by passing one or two writable stream instances.`

When I create constructor for my console class, I am failed.

grgustaf:
FYI we aim to support a subset of the LTS version of Node.js APIs (currently 6.9.4) vs. the latest and greatest (currently 7.4.0). But that doesn't change this item.

I think this bug is really just talking about some basic constructor problem? But it does point out that eventually we might want users to be able to create consoles that log to files on the flash filesystem, and/or UART or something.

jprestwo:
@grgustaf, @cuiyanx,

I think for now we don't need a console constructor because we only support printing to "stdout" (UART), but we could change this to an enhancement/feature request for allowing a constructor to be called on a stream e.g. UART1/2, FS etc.

jprestwo:
@grgustaf, @cuiyanx,

Can we close this and reopen a feature request instead.

cuiyanx:
@jprestwo OK, thanks for your attention。

Issue #621, opened 2017-01-19
[console] Not support placeholders for Console APIs


cuiyanx:
Set "%d" for number and "%s" for string, there are not work.

Code : 
```
var arg1 = 1024;
console.log("console.log: %d", arg1);
var arg2 = "My love!";
console.log("console.log: %s", arg2);
```

grgustaf:
I don't think it makes sense for us to implement support for this, unless it's already present in JerryScript or something and we can reuse it. So probably @jprestwo you can resolve this by specifying in the documentation that format specifiers are not supported. This was a surprise to me, I never knew this existed. :) But it's there in both Node and e.g. Chrome.

jprestwo:
Docs added with: https://github.com/01org/zephyr.js/pull/787

cuiyanx:
OK，thanks and closed.

Issue #620, opened 2017-01-19
[console] Not support to console.assert()


cuiyanx:
Set parameter as "true", console.assert() is not work; 
Set parameter as "false", console.assert() throw out error.

Error:
```
src/zjs_console.c:271 console_assert():
[ERROR] string is too long
```

jprestwo:
The correct usage for console.assert() is:

```
console.assert(value[, message][, ...args])
```

Specifying a string as the first parameter won't work. What is strange is that you are seeing the "string is too long" print. When I specify a string e.g. "true" as the first parameter I get:
```
[ERROR] invalid parameters
```
If you use a true boolean value as the first parameter it behaves normally. Using a false boolean value also behaves correctly by throwing an error. Are you seeing behavior different than this?

jprestwo:
Sorry I was not understanding how you produced these results. I see the problem your seeing now when you specify a message in the assert call.

cuiyanx:
fixed and closed

PR #624, opened 2017-01-19
[tests] Add test case for Console APIs

cuiyanx:
Issue: [#620](https://github.com/01org/zephyr.js/issues/620) [console] Not support to console.assert()
Issue: [#621](https://github.com/01org/zephyr.js/issues/621) [console] Not support placeholders for Console APIs
Issue: [#622](https://github.com/01org/zephyr.js/issues/622) [console] Not support constructor to create
Issue: [#623](https://github.com/01org/zephyr.js/issues/623) [console] Throw out wrong error

### DIFF
--- a/samples/tests/Console.js
+++ b/samples/tests/Console.js
@@ -3,44 +3,40 @@
 var consoleNum = 1024;
 var consoleStr = "abcd";
 console.log("Testing console.log...");
-console.log("                Arrays:", [], [1, 2, 3], ['a', 0, [1, 2, 3]]);
-console.log("              Booleans:", true, false);
-console.log("             Functions:", function () {});
-console.log("               Numbers:", 1, -1, 123.456, -123.456);
-console.log("               Objects:", {}, {foo: 'bar'});
-console.log("               Strings:", '', "", 'abc', "def");
-console.log("                  Misc:", null, undefined);
-console.log("placeholders: %d and %s", consoleNum, consoleStr);
+console.log("   Arrays:", [], [1, 2, 3], ['a', 0, [1, 2, 3]]);
+console.log(" Booleans:", true, false);
+console.log("Functions:", function () {});
+console.log("  Numbers:", 1, -1, 123.456, -123.456);
+console.log("  Objects:", {}, {foo: 'bar'});
+console.log("  Strings:", '', "", 'abc', "def");
+console.log("     Misc:", null, undefined);
 
 console.log("\nTesting console.error...");
-console.error("                Arrays:", [], [1, 2, 3], ['a', 0, [1, 2, 3]]);
-console.error("              Booleans:", true, false);
-console.error("             Functions:", function () {});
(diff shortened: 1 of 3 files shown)


samples/tests/Console.js
```diff
@@ -1,22 +1,46 @@
...
-console.log("  Strings:", '', "", 'abc', "def");  // expect: "  abc def"
-console.log("     Misc:", null, undefined);
+console.log("                Arrays:", [], [1, 2, 3], ['a', 0, [1, 2, 3]]);
+console.log("              Booleans:", true, false);
+console.log("             Functions:", function () {});
+console.log("               Numbers:", 1, -1, 123.456, -123.456);
+console.log("               Objects:", {}, {foo: 'bar'});
+console.log("               Strings:", '', "", 'abc', "def");
+console.log("                  Misc:", null, undefined);
+console.log("placeholders: %d and %s", consoleNum, consoleStr);
```
grgustaf: We don't intend to support this at this time.

samples/tests/Console.js
```diff
@@ -30,3 +54,20 @@ console.log(longstr);
 console.log("\nTesting a string too long...");
 longstr += '6';  // length 256
 console.log(longstr);
+
+console.log("\nTesting console.assert... ");
+console.assert(true);
+console.log("expected result: true");
+console.assert(false, "My lover!");
```
grgustaf: O_o let's make the sample string less interesting? :)

This should be throwing an error - with a new patch from @jprestwo it will be AssertionError like Node docs say. So you should do a try/catch.

samples/tests/Console.js
```diff
@@ -30,3 +54,20 @@ console.log(longstr);
...
+console.log("expected result: false - My lover!");
+console.assert(consoleStr);
+console.log("expected result: [error] invalid parameters");
+console.assert(true, longstr);
+console.log("expected result: [error] string is too long");
```
grgustaf: If the condition is true, the longstr is never looked at so it shouldn't fail or care that the string is too long.

cuiyanx:
sorry, I have a mistake to remove this branch.

samples/tests/Console.js
```diff
@@ -30,3 +50,30 @@ console.log(longstr);
...
+            "expected result [ERROR] string is too long");
+
+console.assert(false, consoleNum);
+console.log("assert(false, Num): " +
+            "expected result [ERROR] message parameter must be a string");
+
+console.log("\nTesting constructor to create...");
+var stdout, stderr;
+var myConsole = new Console(stdout, stderr);
+myConsole.log("create new console as 'myConsole'");
```
grgustaf: This doesn't work and I don't think we've documented the constructor yet so I don't think I would write tests for it.

In any case, we have no way for you to provide valid streams to pass the constructor at this point, and it's not going to work to give it undefined variables like this.

cuiyanx:
Close this one, i will submit new pull request instead against review comment.

PR #625, opened 2017-01-19
[console] Fix bug to support console.assert()

cuiyanx:
fix issue #620 

### DIFF
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -253,33 +253,36 @@ static jerry_value_t console_assert(const jerry_value_t function_obj,
                                     const jerry_length_t argc)
 {
     char message[MAX_STR_LENGTH];
-    uint8_t has_message = 0;
     if (!jerry_value_is_boolean(argv[0])) {
         ERR_PRINT("invalid parameters\n");
         return ZJS_UNDEFINED;
     }
-    bool b = jerry_get_boolean_value(argv[0]);
-    if (!b) {
+    if (argv[1]) {
         if (argc > 1) {
             if (!jerry_value_is_string(argv[1])) {
                 ERR_PRINT("message parameter must be a string\n");
                 return ZJS_UNDEFINED;
             } else {
                 jerry_size_t sz = MAX_STR_LENGTH;
-                zjs_copy_jstring(argv[0], message, &sz);
+                zjs_copy_jstring(argv[1], message, &sz);
                 if (!sz) {
                     ERR_PRINT("string is too long\n");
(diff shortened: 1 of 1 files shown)


src/zjs_console.c
```diff
@@ -253,33 +253,36 @@ static jerry_value_t console_assert(const jerry_value_t function_obj,
...
 {
     char message[MAX_STR_LENGTH];
-    uint8_t has_message = 0;
     if (!jerry_value_is_boolean(argv[0])) {
         ERR_PRINT("invalid parameters\n");
         return ZJS_UNDEFINED;
     }
-    bool b = jerry_get_boolean_value(argv[0]);
-    if (!b) {
+    if (argv[1]) {
```
grgustaf: You can't do this, you don't even know if there *is* a second argument yet. It's an optional argument, and even if it wasn't you still have to make sure it's there.

src/zjs_console.c
```diff
@@ -253,33 +253,36 @@ static jerry_value_t console_assert(const jerry_value_t function_obj,
...
         if (argc > 1) {
             if (!jerry_value_is_string(argv[1])) {
                 ERR_PRINT("message parameter must be a string\n");
                 return ZJS_UNDEFINED;
             } else {
                 jerry_size_t sz = MAX_STR_LENGTH;
-                zjs_copy_jstring(argv[0], message, &sz);
+                zjs_copy_jstring(argv[1], message, &sz);
```
grgustaf: OK you're right here, this was a bug, reading the wrong argument.

grgustaf: It's better to use zjs_alloc_from_jstring here so you don't have to worry about the string size as much.

src/zjs_console.c
```diff
@@ -253,33 +253,36 @@ static jerry_value_t console_assert(const jerry_value_t function_obj,
...
+    }
+    for (int i = 0; i < argc; i++) {
+        if (i) {
+            // insert spaces between arguments
+            fprintf(stdout, " ");
+        }
+        if (i == 1) {
+            fprintf(stdout, "- ");
         }
+        print_value(argv[i], stdout, true, false);
```
grgustaf: These arguments aren't supposed to be printed, they're supposed be used with format specifiers in the message string like %d. But as I mentioned elsewhere, I don't think we should attempt to support that at this time.

cuiyanx: I think it looks like the assert function on my test case, but it is not. So, Be closed. 
Please look at #630 .

PR #627, opened 2017-01-19
[analyze] Detect Buffer constructor with any arguments.

pfalcon:
Previously, detection expected a numeric argument, e.g. `new Buffer(8)`,
but it can also be called with a string, like `new Buffer("foo")`. So,
support any argument (including still supporting none at all).

### DIFF
--- a/scripts/analyze.sh
+++ b/scripts/analyze.sh
@@ -75,7 +75,7 @@ function check_config_file()
 }
 
 # Check if buffer module required in the JS or config file
-grep Buffer\([0-9]*\) $SCRIPT > /dev/null
+grep Buffer\(.*\) $SCRIPT > /dev/null
 if [ $? -eq 0 ] || check_config_file ZJS_BUFFER; then
     buffer=true;
 else


pfalcon:
This gets more and more heuristic, but it was such from the beginning (just imagine if you had code commented, it will still "detect it"). Apparently, it's better to have false positives than false negatives.

I actually wanted to ask if you guys considered doing analysis on AST level of something. If some ideas like e.g. @zolkis' from https://github.com/01org/zephyr.js/issues/368#issuecomment-259437045 come to fruition, you'd need to do it anyway. (And that's one of the most immediate surface-level drawbacks of the API in that proposal, which point I failed to communicate to @zolkis all this time, sorry about that.)

grgustaf:
+1

grgustaf:
@pfalcon I'd definitely be interested in a more accurate way to analyze the code if you know of good tools to suggest.

pfalcon:
I'm not well-versed in JS ecosystem. But such tools exist for C/C++ and other languages, so I'd assume they exist for JS too, but that would take a separate research.

PR #631, opened 2017-01-20
[build] Fix clean build to remove libjerry-core.a

jprestwo:
Added line to remove $(JERRY_BASE)/build

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,7 @@ clean:
 		cd arc/; make clean; \
 	fi
 	make -f Makefile.linux clean
+	@rm -rf $(JERRY_BASE)/build
 	@rm -f src/*.o
 	@rm -f src/zjs_script_gen.c
 	@rm -f src/zjs_snapshot_gen.c


brianjjones:
LGTM +1

Makefile
```diff
@@ -139,6 +139,7 @@ clean:
 		cd arc/; make clean; \
 	fi
 	make -f Makefile.linux clean
+	@rm -rf $(JERRY_BASE)/build
```
grgustaf: It would be better for this to be BOARD-specific, but to do this nuclear option in 'make pristine'.

However, I see that line is already in the above section... so why isn't that working now? Also why is it conditional under ZEPHYR_SDK_INSTALL_DIR? I don't think that makes sense.

grgustaf:
Oh, I see my problem. I was running "make clean" but really I needed to be doing "make BOARD=qemu_x86 clean" etc. Maybe we should store the last board used in a .last_board or something and use that as the default if one isn't specified. Then 'make clean' would work more as expected.

jprestwo:
@grgustaf, the question is now: how do you clean for Linux? Since it doesn't BOARD=. I do like having to specify BOARD if you want to clean for that target, but if you just run "make clean" what should happen?

grgustaf:
Dropping in favor of a new pull request.

PR #635, opened 2017-01-20
[general] Review and correct lifetime of jerry_value_t items

grgustaf:
The document 06.REFERENCE-COUNTING.md in JerryScript makes it clear
now how and when jerry_value_t items must be released. In particular,
you're not responsible to release values passed in to you, but any
function that returns a value is giving you a "live" reference to
it that you must release. But when you return from a native function,
you transfer ownership of the returned value back to JavaScript.

In particular, even the error value that functions return is supposed
to be released, which we rarely if ever did before this review.

Reviewed all code in src/*.c except for the OCF files which were too
foreign to me to address correctly.


brianjjones:
LGTM +1

PR #641, opened 2017-01-22
[tests] Remove abnormal test point cannot be catched

cuiyanx:

### DIFF
--- a/tests/test-event.js
+++ b/tests/test-event.js
@@ -3,59 +3,22 @@
 // Testing EVENT APIs
 
 var event = require("events");
-
-var total = 0;
-var passed = 0;
-
-function assert(actual, description) {
-    total += 1;
-
-    var label = "\033[1m\033[31mFAIL\033[0m";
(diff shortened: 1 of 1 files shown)


tests/test-event.js
```diff
@@ -3,59 +3,22 @@
...
-    eventEmitter.setMaxListeners(-1024);
-});
-
-expectThrow("event: get max of listeners on event named 'value'", function () {
-    eventEmitter.getMaxListeners("value");
-});
-
 var defaultNum = eventEmitter.getMaxListeners();
-assert(defaultNum === 0, "event: default number of listeners");
+Assert.true(defaultNum === 10, "event: default number of listeners");
```
grgustaf: If you use my new patch #650 you wouldn't have to change these lines. (Just do `var assert = require('Assert.js')` above instead.)

So if you're okay with that patch, maybe rebase this on top of it.

grgustaf:
+1

PR #644, opened 2017-01-23
Enabled buffer support on Linux

jprestwo:
- Fixed minor overflow bug that Linux compiler warned about.

### DIFF
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -31,7 +31,8 @@ include Makefile.ocf_linux
 
 LINUX_DEFINES +=	-DOC_CLIENT \
 			-DOC_SERVER \
-			-DBUILD_MODULE_OCF
+			-DBUILD_MODULE_OCF \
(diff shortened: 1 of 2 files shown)


grgustaf:
+1

pfalcon:
@jprestwo : Thanks for fixing this. But note that you added -DBUILD_MODULE_BUFFER to an "OCF" section of Makefile.linux, which is in turn conditional on Linux vs Mac stuff. As a result, Buffer support is nto available in MacOS build. I guess, instead it should be added to to "general" section: https://github.com/01org/zephyr.js/blob/master/Makefile.linux#L81

(And perhaps, to avoid such confusion in the future, it would make sense to restructure Makefile, putting general settings first, conditional - after them.)

jprestwo:
@pfalcon, good catch sorry about that. Should be fixed with this: https://github.com/01org/zephyr.js/pull/658

PR #645, opened 2017-01-23
Fixed multiple bugs in events

jprestwo:
- tests/test-events.js uncovered several bugs
 - Fixed one issue in the test which was incorrectly testing

### DIFF
--- a/src/zjs_event.c
+++ b/src/zjs_event.c
@@ -40,31 +40,32 @@ void post_event(void* h, jerry_value_t* ret_val)
     }
 }
 
-void zjs_add_event_listener(jerry_value_t obj, const char* event,
-                            jerry_value_t listener)
+static uint32_t get_num_events(jerry_value_t emitter)
 {
-    event_t *ev;
-
-    jerry_value_t event_emitter = zjs_get_property(obj, HIDDEN_PROP("event"));
-    if (!jerry_get_object_native_handle(event_emitter, (uintptr_t*)&ev)) {
-        jerry_release_value(event_emitter);
-        ERR_PRINT("native handle not found\n");
-        return;
+    jerry_value_t val = zjs_get_property(emitter, "numEvents");
+    if (!jerry_value_is_number(val)) {
+        ERR_PRINT("emitter had no numEvents property\n");
+        return 0;
     }
-    jerry_release_value(event_emitter);
+    uint32_t num = jerry_get_number_value(val);
+    jerry_release_value(val);
+    return num;
+}
 
-    if (ev->num_events >= ev->max_listeners) {
-        ERR_PRINT("max listeners reached\n");
-        return;
+static uint32_t get_max_event_listeners(jerry_value_t emitter)
+{
+    jerry_value_t val = zjs_get_property(emitter, "maxListeners");
+    if (!jerry_value_is_number(val)) {
+        ERR_PRINT("emitter had no maxListeners property\n");
+        return 0;
     }
+    uint32_t num = jerry_get_number_value(val);
+    jerry_release_value(val);
+    return num;
+}
 
-    // Event object to hold callback ID and eventually listener arguments
-    jerry_value_t event_obj = zjs_get_property(ev->map, event);
-    if (!jerry_value_is_object(event_obj)) {
-        jerry_release_value(event_obj);
-        event_obj = jerry_create_object();
-    }
-
+static int32_t get_callback_id(jerry_value_t event_obj)
+{
     int32_t callback_id = -1;
     jerry_value_t id_prop = zjs_get_property(event_obj, "callback_id");
     if (jerry_value_is_number(id_prop)) {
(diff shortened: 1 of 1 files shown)


jprestwo:
@grgustaf, once https://github.com/01org/zephyr.js/pull/664 goes in I can rebase this. Since Assert.js got added to the event tests I had to add JS module loading to the Linux target.

src/zjs_event.c
```diff
@@ -198,54 +216,50 @@ static jerry_value_t remove_all_listeners(const jerry_value_t function_obj,
...
         jerry_value_t name = jerry_create_string((const jerry_char_t*)event);
-        jerry_delete_property(ev->map, (const jerry_value_t)name);
+        jerry_delete_property(map, (const jerry_value_t)name);
         jerry_release_value(name);
     } else {
         ERR_PRINT("callback_id not found for '%s'\n", event);
     }
 
-    jerry_release_value(id_prop);
+    zjs_obj_add_number(event_emitter, 0, "numEvents");
```
grgustaf: I didn't see you decrement this in remove_listener - should you have? Is it really numListeners you're counting?

src/zjs_event.c
```diff
@@ -266,21 +280,22 @@ static jerry_value_t get_event_names(const jerry_value_t function_obj,
...
 {
-    event_t *ev;
     event_names_t names;
 
-    jerry_value_t event_emitter = zjs_get_property(this, HIDDEN_PROP("event"));
-    if (!jerry_get_object_native_handle(event_emitter, (uintptr_t*)&ev)) {
-        jerry_release_value(event_emitter);
-        ERR_PRINT("native handle not found\n");
-        return ZJS_UNDEFINED;
+    if (argc) {
```
grgustaf: For better or worse, normally we ignore extra arguments I think. Maybe that's okay?

src/zjs_event.c
```diff
@@ -290,38 +305,33 @@ static jerry_value_t get_max_listeners(const jerry_value_t function_obj,
...
                                        const jerry_length_t argc)
 {
-    event_t *ev;
-
-    jerry_value_t event_emitter = zjs_get_property(this, HIDDEN_PROP("event"));
-    if (!jerry_get_object_native_handle(event_emitter, (uintptr_t*)&ev)) {
-        jerry_release_value(event_emitter);
-        ERR_PRINT("native handle not found\n");
-        return jerry_create_number(DEFAULT_MAX_LISTENERS);
+    if (argc) {
```
grgustaf: Again, not sure it's worth checking, but feel free to ignore me. :)

src/zjs_event.c
```diff
@@ -290,38 +305,33 @@ static jerry_value_t get_max_listeners(const jerry_value_t function_obj,
...
 
-    if (!jerry_value_is_number(argv[0])) {
-        ERR_PRINT("max listeners count must be first parameter\n");
-        return jerry_acquire_value(this);
+    double num = jerry_get_number_value(argv[0]);
+    if (num < 0) {
+        return zjs_error("max listener value must be a positive integer");
     }
-    ev->max_listeners = (uint32_t)jerry_get_number_value(argv[0]);
+    zjs_obj_add_number(event_emitter, num, "maxListeners");
```
grgustaf: Should we make this readonly? If so, just replace add_number with add_readonly_number.

src/zjs_event.c
```diff
@@ -382,45 +384,33 @@ static jerry_value_t get_listeners(const jerry_value_t function_obj,
...
-    jerry_value_t event_obj = zjs_get_property(ev->map, event);
+    jerry_value_t map = zjs_get_property(event_emitter, "map");
+    jerry_value_t event_obj = zjs_get_property(map, event);
+    jerry_release_value(map);
+    jerry_release_value(event_emitter);
+
     if (!jerry_value_is_object(event_obj)) {
-        DBG_PRINT("event object not found for '%s'\n", event);
-        return ZJS_UNDEFINED;
+        return zjs_error("event object not found");
```
grgustaf: still need to free event_obj

src/zjs_event.c
```diff
@@ -445,34 +433,27 @@ bool zjs_trigger_event(jerry_value_t obj,
...
-        return false;
-    }
+    jerry_value_t map = zjs_get_property(event_emitter, "map");
+    jerry_value_t event_obj = zjs_get_property(map, event);
+    jerry_release_value(map);
     jerry_release_value(event_emitter);
 
-    jerry_value_t event_obj = zjs_get_property(ev->map, event);
     if (!jerry_value_is_object(event_obj)) {
-        jerry_release_value(event_obj);
```
grgustaf: Don't remove this. you still need to free it - what if it was a string or error, etc. If OTOH you're already 100% sure it's an object, then don't check I guess.

src/zjs_event.c
```diff
@@ -499,34 +480,27 @@ bool zjs_trigger_event_now(jerry_value_t obj,
...
```
grgustaf: don't remove

grgustaf:
Before the changes, maybe rebase against latest?

src/zjs_event.c
```diff
@@ -40,50 +40,84 @@ void post_event(void* h, jerry_value_t* ret_val)
...
+    }
+    uint32_t num = jerry_get_number_value(val);
+    jerry_release_value(val);
+    return num;
+}
+
+static int32_t get_callback_id(jerry_value_t event_obj)
+{
+    int32_t callback_id = -1;
+    jerry_value_t id_prop = zjs_get_property(event_obj, "callback_id");
```
jprestwo: This object is a property of a hidden object so there should be no way for them to access it.

src/zjs_event.c
```diff
@@ -40,50 +40,84 @@ void post_event(void* h, jerry_value_t* ret_val)
...
 
-    int32_t callback_id = -1;
-    jerry_value_t id_prop = zjs_get_property(event_obj, "callback_id");
-    if (jerry_value_is_number(id_prop)) {
-        // If there already is an event object, get the callback ID
-        zjs_obj_get_int32(event_obj, "callback_id", &callback_id);
-    }
-    jerry_release_value(id_prop);
-    callback_id = zjs_add_callback_list(listener, obj, ev, post_event,
+    int32_t callback_id = get_callback_id(event_obj);
```
jprestwo: if its -1 then add_callback_list will create a new list and return the new callback ID. If its not -1 then it will add the function to the existing callback list.

src/zjs_event.c
jprestwo: numEvents is different than the number of listeners. Events doesnt actually store the number of listeners, that is stored in a callback list, so when getNumListeners() is called we just ask the callback list how many functions there are. 

numEvents is the number of events registered to a listener, like onread, onerror, ondata etc.

src/zjs_event.c
jprestwo: I think this was a change to conform to the test-events.js test. But I agree, extra arguments shouldnt be an error.

src/zjs_event.c
jprestwo: I agree with you, removing

src/zjs_event.c
jprestwo: This can be changed by the event module though. Its also hidden behind the hidden "event" property so its not accessible from the JS script.

src/zjs_event.c
grgustaf: Obviously I was too lazy to read too carefully. :)

PR #646, opened 2017-01-23
[error] Add AssertionError to error_types array to fix bug

grgustaf:
Also, alphabetize the error lists and add a comment to prevent this kind
of problem in the future.


brianjjones:
LGTM +1

PR #647, opened 2017-01-23
[console] Improve assert to print all values and throw errors

grgustaf:
Rewrite print_value to work in terms of a new value2str so that it can
be called independently when we need a string and not direct output.


brianjjones:
LGTM +1

jprestwo:
+1, looks good

PR #648, opened 2017-01-23
[callbacks] Print errors for our own callbacks

jprestwo:
- There is still not support for handling these errors,
   but at least printing them is better than nothing

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -388,6 +388,53 @@ void print_callbacks(void)
 #define print_callbacks() do {} while (0)
 #endif
 
+#define MAX_ERROR_NAME_LENGTH       32
+#define MAX_ERROR_MESSAGE_LENGTH    128
+
+static void print_error_message(jerry_value_t error)
+{
+    uint32_t size;
+    char* message = NULL;
+    jerry_value_t name_val = zjs_get_property(error, "name");
+    if (!jerry_value_is_string(name_val)) {
+        ERR_PRINT("error did not have name\n");
+        // we should never get here.
+        return;
+    }
+    jerry_value_t msg_val = zjs_get_property(error, "message");
+    if (!jerry_value_is_string(msg_val)) {
+        ERR_PRINT("error did not have message\n");
+        // we should never get here.
+        return;
+    }
+
+    size = MAX_ERROR_NAME_LENGTH;
+    char name[size];
+
+    zjs_copy_jstring(name_val, name, &size);
+    if (!size) {
+        ERR_PRINT("name length is too long\n");
+        return;
+    }
+    jerry_release_value(name_val);
+
+    size = MAX_ERROR_MESSAGE_LENGTH;
+    char msg[size];
+
+    zjs_copy_jstring(msg_val, msg, &size);
+    if (size) {
+        message = msg;
+    }
+
+    jerry_release_value(msg_val);
+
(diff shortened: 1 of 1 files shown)


src/zjs_callbacks.c
```diff
@@ -388,6 +388,53 @@ void print_callbacks(void)
...
 #endif
 
+#define MAX_ERROR_NAME_LENGTH       32
+#define MAX_ERROR_MESSAGE_LENGTH    128
+
+static void print_error_message(jerry_value_t error)
+{
+    uint32_t size;
+    char* message = NULL;
+    jerry_value_t name_val = zjs_get_property(error, "name");
```
grgustaf: We decided not to use _val appendages, just leave them off.

src/zjs_callbacks.c
```diff
@@ -388,6 +388,53 @@ void print_callbacks(void)
...
+    if (!jerry_value_is_string(name_val)) {
+        ERR_PRINT("error did not have name\n");
```
grgustaf: You still need to release the name_val on the error path here.

src/zjs_callbacks.c
```diff
@@ -388,6 +388,53 @@ void print_callbacks(void)
...
+        // we should never get here.
+        return;
+    }
+    jerry_value_t msg_val = zjs_get_property(error, "message");
+    if (!jerry_value_is_string(msg_val)) {
+        ERR_PRINT("error did not have message\n");
```
grgustaf: And both name_val and msg_val here.

src/zjs_callbacks.c
```diff
@@ -388,6 +388,53 @@ void print_callbacks(void)
...
+        // we should never get here.
+        return;
+    }
+
+    size = MAX_ERROR_NAME_LENGTH;
+    char name[size];
+
+    zjs_copy_jstring(name_val, name, &size);
```
grgustaf: When you don't explicitly care about the string size, you're better off calling zjs_alloc_from_jstring and passing it a 0 maxlen. This will allocate from the heap as long as there's enough memory to fit it. I think that suffices in most cases and it makes the calling code simpler. Just be sure to free it when you're done.

I guess you still need to check for NULL, which I might not have done in a couple places. O_o! So tell me if I'm wrong and it doesn't end up being easier. :)

src/zjs_callbacks.c
```diff
@@ -388,6 +388,53 @@ void print_callbacks(void)
...
+    if (!size) {
+        ERR_PRINT("name length is too long\n");
```
grgustaf: Another return case where you're not freeing those other values. This also kind of encourages structuring your code with a smaller number of return points I guess. For better or worse. Wish we had smart pointers. :)

src/zjs_callbacks.c
```diff
@@ -395,41 +395,42 @@ static void print_error_message(jerry_value_t error)
...
+        jerry_release_value(err_msg);
         return;
     }
-    jerry_release_value(name_val);
+    jerry_release_value(err_name);
 
-    size = MAX_ERROR_MESSAGE_LENGTH;
-    char msg[size];
+    size = 0;
+    message = zjs_alloc_from_jstring(err_msg, &size);
```
grgustaf: Just one more fine point: If you don't need the size, you can just pass NULL.

grgustaf:
+1

PR #649, opened 2017-01-24
[docs] Alphabetize API document list

grgustaf:
Always a good idea to alphabetize: helps the users find what they're looking
for, helps us avoid duplication when the list gets longer, etc.


brianjjones:
LGTM +1

PR #650, opened 2017-01-24
[assert] Make the QA Assert module match Node APIs better

grgustaf:
This improves the basic syntax to:
var assert = require('Assert');
assert(true, "successful assertion");
assert(false, "failed assertion");

Also allow any "truthy" value to pass. If you explicitly want to check for
exact equality with true, then do:
assert(operation() === true, "perform operation");

Relax about argument types; equal can check non-object arguments. The
description argument can be a non-string or doesn't have to be given.

Swapped argument order in assert.throws() to match Node API.

### DIFF
--- a/modules/Assert.js
+++ b/modules/Assert.js
@@ -3,54 +3,27 @@
 // JavaScript library for the tests case
 
 function Assert() {
-
     // API object
-    var assert = {};
-
-    assert.total = 0;
-    assert.pass = 0;
-
-    assert.true = function (actual, description) {
-        if (typeof actual !== "boolean" ||
-            typeof description !== "string") {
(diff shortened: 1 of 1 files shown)


grgustaf:
@cuiyanx, see what you think about this. This lets you use assert() instead of assert.true(), and a few other minor changes. I tested it with this script, and it does what I think it should:

```javascript
var assert = require('Assert.js');
assert(true, "good times");
assert(false, "bad times");
assert.throws(function () { return 3; });
assert.result();
```

cuiyanx:
+1
Better to me!

brianjjones:
LGTM

PR #652, opened 2017-01-24
[util] Serialize data function

jprestwo:
- Can be used to convert any serializable jerry_value_t into a C buffer
   (Buffers, Strings, and Arrays (of bytes) supported)

### DIFF
--- a/src/zjs_uart.c
+++ b/src/zjs_uart.c
@@ -161,20 +161,22 @@ static jerry_value_t uart_write(const jerry_value_t function_obj,
 
     zjs_make_promise(promise, NULL, NULL);
 
-    if (!jerry_value_is_object(argv[0])) {
-        DBG_PRINT("first parameter must be a Buffer");
+    uint32_t len;
+    void* data = zjs_serialize_data(argv[0], NULL, &len);
+    if (!data) {
+        ERR_PRINT("invalid parameters\n");
         jerry_value_t error = make_uart_error("TypeMismatchError",
-                "first parameter must be a Buffer");
+                                              "first parameter must be a Buffer");
         zjs_reject_promise(promise, &error, 1);
         jerry_release_value(error);
         return promise;
     }
 
-    zjs_buffer_t* buffer = zjs_buffer_find(argv[0]);
-
-    write_data(uart_dev, (const char*)buffer->buffer, buffer->bufsize);
+    write_data(uart_dev, (const char*)data, len);
 
     zjs_fulfill_promise(promise, NULL, 0);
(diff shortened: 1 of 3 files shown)


src/zjs_uart.c
```diff
@@ -161,21 +161,23 @@ static jerry_value_t uart_write(const jerry_value_t function_obj,
 
     zjs_make_promise(promise, NULL, NULL);
 
-    if (!jerry_value_is_object(argv[0])) {
-        DBG_PRINT("first parameter must be a Buffer");
+    uint32_t len;
+    void* data = zjs_serialize_data(argv[0], NULL, &len);
+    if (!data) {
+        ERR_PRINT("invalid parameters\n");
         jerry_value_t error = make_uart_error("TypeMismatchError",
```
grgustaf: This could just use TYPE_ERROR now.

src/zjs_util.h
```diff
@@ -97,4 +97,18 @@ void zjs_default_convert_pin(uint32_t orig, int *dev, int *pin);
...
 
+/**
+ * Serialize a jerry_value_t into a buffer. The returned buffer must be freed
+ * after use.
+ *
+ * @param value     Value to serialize. This can be a Buffer, Array (of bytes)
+ *                  or String.
+ * @param maxLen    Maximum length of data to serialize. If the length does not
+ *                  matter you can specify 0 or NULL.
+ * @param actualLen Actual length of buffer
```
grgustaf: You really mean this is an output with the bytes written. Maybe you should also allow that to be NULL in case they don't care? E.g. "If not NULL, receives the number of bytes written to the buffer."

src/zjs_uart.c
```diff
@@ -161,21 +161,23 @@ static jerry_value_t uart_write(const jerry_value_t function_obj,
 
     zjs_make_promise(promise, NULL, NULL);
 
-    if (!jerry_value_is_object(argv[0])) {
-        DBG_PRINT("first parameter must be a Buffer");
+    uint32_t len;
+    void* data = zjs_serialize_data(argv[0], NULL, &len);
```
grgustaf: Do you intend to be changing this API so that strings and byte arrays are now valid, because you seem to be doing that. If so, I guess update the API?

jprestwo: I am not sure if this should go in. We would need to talk to Sakari because the iotjs spec does not say anything about strings/arrays. I personally think it would be nice for users, but we should talk about it.

grgustaf:
Tabling for now.

PR #653, opened 2017-01-24
[make] Made make HW support more consistent

jprestwo:
- To clean, you must specify the BOARD you want to clean for. Not specifying
   BOARD will result in the default (Arduino 101) getting cleaned.
 - To build for Linux, you now specify BOARD=linux. This matches all the other
   targets. To clean for linux you must also specify BOARD=linux.
 - Fixed the jerryscript clean issue where the library was not getting removed
   during a clean, hence JrS would use the old stale library.


grgustaf:
+1

PR #654, opened 2017-01-25
[buffer] Fix bug with uninitialized buffer in write_string

grgustaf:
This bug was hit if you tried to pass the fourth encoding parameter
to the function; it would get an error even if you passed 'utf8'.

Also, cleaned up indentation issue in I2C.js.

### DIFF
--- a/src/zjs_buffer.c
+++ b/src/zjs_buffer.c
@@ -280,15 +280,14 @@ static jerry_value_t zjs_buffer_write_string(const jerry_value_t function_obj_va
 
     // Check if the encoding string is anything other than utf8
     if (argc > 3) {
-        const char *utf8_encoding = "utf8";
-        int utf8_len = strlen(utf8_encoding);
-        jerry_size_t size = utf8_len + 1;
-        char *encoding = zjs_alloc_from_jstring(argv[3], &size);
+        char *encoding = zjs_alloc_from_jstring(argv[3], NULL);
         if (!encoding) {
             return zjs_error("zjs_buffer_write_string: allocation failure");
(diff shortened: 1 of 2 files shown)


src/zjs_buffer.c
```diff
@@ -280,20 +280,17 @@ static jerry_value_t zjs_buffer_write_string(const jerry_value_t function_obj_va
 
     // Check if the encoding string is anything other than utf8
     if (argc > 3) {
-        jerry_value_t arg4 = argv[3];
-        jerry_size_t arg4_sz = jerry_get_string_size(arg4);
-
-        if (arg4_sz > 4096) {
-            return zjs_error("zjs_buffer_write_string: encoding arg string is too long");
+        // ask for one more char than needed to make sure not just prefix match
+        jerry_size_t size = 5;
```
brianjjones: Shouldn't this be jerry_get_string_size(arg4) + 1 ?  Otherwise this is just limiting the string to a size 5 bytes.  Or am I missing something?

grgustaf: Since the string we're checking it against is "utf8", we only need the first five bytes to compare. This could be made more clear w/ comments or perhaps a standard compare function.

brianjjones: True but could we be doing utf16 or utf32 in the future?  I realize we only support 'utf8' right now, but I hate to put hard coded numbers in.

grgustaf: This code will have to change when we do that, one way or the other.

brianjjones:
LGTM +1

Issue #386, opened 2016-11-04
Node.JS UDP Socket API


baleboy:
https://nodejs.org/api/dgram.html

See also example: http://www.hacksparrow.com/node-js-udp-server-and-client-example.html

pfalcon:
As I mentioned elsewhere, I'd be interested to work on this (when more ground-level issues are tackled). So, if someone would look into this too, let's keep in touch.

poussa:
@pfalcon we don't have anybody working on this so it is yours.

pfalcon:
Thanks. With other background work (e.g. Zephyt 1.6 release and incompatibilities it brings) it may take some time before I may get to this, so leaving the above notice if anything changes.

jprestwo:
@pfalcon, have you started this yet? OCF just merged so I am open to making some headway here. @poussa did say he wanted HTTP as a priority but I think we are going to have to wait for the new IP stack (correct me if I'm wrong). Zephyr also has a simple echo client and server using UDP, so the current features for UDP do seem to work.

poussa:
@jprestwo Lets stick with the original plan. @pfalcon will take UDP (this) and you will do TCP+HTTP.

Both of you should be using the new IP stack which is in the 'net' branch of the Zephyr. The 'net' branch follows master closely and will be merged to master during this month. The current IP stack and APIs should not be used. Also, if you need any Zephyr services, please use the unified (V2) kernel APIs.

pfalcon:
@jprestwo : I didn't, I'm oscillating between Zephyr hardware drivers and being able to write desktop-style server with Z.js, and all things inbetween, so sorry that I'm making more RFCs than I can work on in a given period of time ;-). Generally, UDP support would be my pet project to keep elaborating my knowledge of Zephyr IP stack and gain some knowledge of Z.js' event/callback scheduling.

If you have nothing else in your queue and ready to work on TCP/HTTP (even though it's likely not fully ready upstream), then I may imagine you'd want to practice on UDP first, then feel free to take this. Otherwise, I agree with @poussa that with the feature list posted by @baleboy , there's doesn't seem to be lack of things to work on ;-).

jprestwo:
@poussa, sounds good, HTTP it is. I did look in the net branch and could not find any references to HTTP API's. We're we still going to go with trying to find an external library? Another option is using an already written JS module since @brianjjones is coming along with module support. Either way, we should discuss how we want to implement HTTP API's.

pfalcon:
[...]

PR #659, opened 2017-01-26
"dgram" (UDP) module

pfalcon:
This implements basic subset of "dgram" module, per https://github.com/01org/zephyr.js/issues/386. It support both IPv4 and IPv6, and implements enough functionality for a UDP echo server (examples included).

This is now ready for review, and some points to pay to:

~~1. Some things aren't yet implemented, like buffer offset/size in .send(), args to send callback, etc.~~
2. JrS API is verbose, that's why zjs_util.c exists, but I find even that verbose too. Repetitive boilerplate code just calls to be hidden behind macros, and that's what I did. Let me know if you like, and they can be made ZJS_GET_STR(), etc., and moved to zjs_util.h . Otherwise, if you don't like it, tell me what I should do.
3. Error handling in C is also verbose and cumbersome. Why the CHECK() macro. It's not supposed to be globalized, but otherwise IMHO a useful per-module pattern.
~~4. There're still printf debugging around, it's of course supposed to be removed in the final version.~~

### DIFF
--- a/docs/dgram.md
+++ b/docs/dgram.md
@@ -1,0 +1,103 @@
+ZJS API for UDP datagram sockets
+================================
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Client Requirements](#requirements)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+The `dgram` API is based on a subset of the
+[corresponding module](https://nodejs.org/api/dgram.html) in Node.js.
+It allows you to send and receive UDP datagrams.
+
+Web IDL
+-------
+This IDL provides an overview of the interface; see below for documentation of
+specific API functions.
+
+```javascript
+// require returns a socket factory object
+// var dgram = require('dgram');
+
+[NoInterfaceObject]
+interface dgram {
+    DgramSocket createSocket(string udp4_or_udp6);
+};
+
+[NoInterfaceObject]
+interface DgramSocket {
+    void on(string event, RecvCallback cb);
+    void bind(int port, string ip_addr);
+    void send(Buffer buf, unsigned long offset, unsigned long len, int port, string ip_addr, [SendCallback cb]);
+    void close();
+};
+
+callback RecvCallback = void (Buffer msg, RemoteInfo rinfo);
+callback SendCallback = void (Error err);  // or undefined if no error
+
+
+callback EventCallback = void (various);  // callback arg depends on event
+
+dictionary RemoteInfo {
+    string ip_addr;
+    string family;
+    int port;
+};
+```
+
+API Documentation
+-----------------
+### dgram.createSocket
+
+`DgramSocket createSocket(string type);`
+
+Create a datagram socket of given type, which must be `'udp4'` or `'udp6'`.
+
+### DgramSocket.on
+
+`void on(string event, RecvCallback callback);`
+
+Registers a callback. The `event` may be one of the following:
+
+* `'message'` - a datagram received. `callback` receives a Buffer
+  containing incoming datagram data and RemoteInfo dictionary with
+  information about the source address.
+* `'error'` - error occurred. `callback` receives an Error object.
+  (In the current version, this callback is never called, but this
+  will change in future versions.)
+
+### DgramSocket.bind
+
+`void bind(int port, string ip_addr);`
+
+Bind socket to a local address and port. This is required operation for
+server-side sockets, i.e. sockets which wait and receive data from other
+network nodes. `ip_addr` must be a string representing an IP address.
+This module does not support domain name resolution, so only IP addresses
+are allowed. An example of IPv4 address: `'192.0.2.1'`, IPV6: `'2001:db8::1'`.
+
+### DgramSocket.send
+
+`void send(Buffer buf, unsigned long offset, unsigned long len, int port, string ip_addr, [SendCallback cb]);`
+
+Send data contained in a buffer to remote network node. A subset of
+data in `buf` can be sent using `offset` and `len` parameters. To send
+entire buffer, using values `0` and `buf.length` respectively. See
+`bind()` method description for the format of `ip_addr`. An optional
+callback may be provided, which will be called with the result of the send
+operation: either NetworkError object in case of error, or `undefined`
+on success.
+
+### DgramSocket.close
+
+`void close();`
+
+Closes socket.
+
+Sample Apps
+-----------
+* [IPv4 UDP echo server](../samples/UDPEchoServ4.js)
+* [IPv6 UDP echo server with `offset` param to send()](../samples/UDPEchoServ6.js)
--- a/samples/UDPEchoServ4.js
+++ b/samples/UDPEchoServ4.js
@@ -1,0 +1,28 @@
+// Copyright (c) 2017, Linaro Limited.
+
+// Sample for "dgram" module implementing an IPv4 UDP echo server.
+// To run it on QEMU, you'll need additional setup for
+// networking to work, follow
+// https://www.zephyrproject.org/doc/subsystems/networking/qemu_setup.html
+// From the host, you can generate test UDP traffic using "netcat"
+// tool:
+//
+// nc -u 192.0.2.1 33333
+//
+// (Then start typing and pressing Enter).
+
+
+var PORT = 33333;
+var HOST = '192.0.2.1';
+
+var dgram = require('dgram');
+var server = dgram.createSocket('udp4');
+
+server.on('message', function (message, remote) {
+    console.log(remote.address + ':' + remote.port +' - ' + message.toString("ascii"));
+    server.send(message, 0, message.length, remote.port, remote.address);
+});
+
+server.bind(PORT, HOST);
+
+console.log("Drop me a (UDP) packet at " + HOST + ":" + PORT);
--- a/src/zjs_dgram.c
+++ b/src/zjs_dgram.c
@@ -1,0 +1,362 @@
+// Copyright (c) 2017, Linaro Limited.
+#ifdef BUILD_MODULE_DGRAM
+
+// Zephyr includes
+#include <net/net_context.h>
+#include <net/nbuf.h>
+
+// ZJS includes
+#include "zjs_util.h"
+#include "zjs_buffer.h"
+#include "zjs_callbacks.h"
+
+static jerry_value_t zjs_dgram_socket_prototype;
+
+typedef struct dgram_handle {
+    struct net_context *udp_sock;
+    zjs_callback_id message_cb_id;
+    zjs_callback_id error_cb_id;
+} dgram_handle_t;
+
+#define CHECK(x) \
+    ret = (x); if (ret < 0) { ERR_PRINT("Error in " #x ": %d\n", ret); return zjs_error(#x); }
+
+#define GET_STR(jval, buf) \
+    { \
+        jerry_size_t str_sz = sizeof(buf); \
+        zjs_copy_jstring(jval, buf, &str_sz); \
+    }
+
+#define GET_HANDLE(type, var) \
+    type *var; \
+    { \
+        uintptr_t native; \
+        if (!jerry_get_object_native_handle(this, &native)) { \
+            return zjs_error("no native handle"); \
+        } \
+        var = (type*)native; \
+    }
+
+// Parse textual address of given address family (IPv4/IPv6) and numeric
+// port and fill in sockaddr. Returns ZJS_UNDEFINED if everything is OK,
+// or error instance otherwise.
+static jerry_value_t get_addr(sa_family_t family,
+                              const jerry_value_t addr,
+                              const jerry_value_t port,
+                              struct sockaddr *sockaddr)
+{
+    int ret;
+
+    if (!jerry_value_is_number(port) || !jerry_value_is_string(addr))
+        return zjs_error("port and IP address expected");
+
+    // We employ the fact that port and address offsets are the same for IPv4&6
+    struct sockaddr_in *sockaddr_in = (struct sockaddr_in*)sockaddr;
+    sockaddr_in->sin_family = family;
+    sockaddr_in->sin_port = htons((int)jerry_get_number_value(port));
+
+    jerry_size_t str_len = 40;
+    char addr_str[str_len];
+    zjs_copy_jstring(addr, addr_str, &str_len);
+    CHECK(net_addr_pton(family, addr_str, &sockaddr_in->sin_addr));
+    return ZJS_UNDEFINED;
+}
+
+static void zjs_dgram_free_cb(const uintptr_t native)
+{
+    dgram_handle_t *handle = (dgram_handle_t*)native;
+    DBG_PRINT("zjs_dgram_free_cb: %p\n", handle);
+    if (!handle) {
+        return;
+    }
+
+    if (handle->udp_sock) {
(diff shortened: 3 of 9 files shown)


pfalcon:
This is now in much better shape: updated to use the latest goodies from net branch (and thus requires it), compiles w/o warnings, all args to .send() implemented. Perhaps the only thing I'd have in my queue is actually implement args to send's callback, and this will really require a review.

pfalcon:
Good news - this now can be built against Zephyr master. I've implemented the argument to .send() callback. On my todo list: 1) documentation; 2) remove debug statements; 3) squash.

pfalcon:
Ok, so all the points above are done, given that there were no review comments, I squashed everything to a reasonable number of commits. @grgustaf : "WIP" label can be removed and this is ready for review.

docs/dgram.md
```diff
@@ -0,0 +1,98 @@
...
+`void bind(int port, string ip_addr);`
+
+Bind socket to a local address and port. This is required operation for
+server-side sockets, i.e. sockets which wait and receive data from other
+network nodes. `ip_addr` must be a string representing an IP address.
+This module does not support domain name resolution, so only IP addresses
+are allowed. An example of IPv4 address: `'192.0.2.1'`, IPV6: `'2001:db8::1'`.
+
+### BLE.send
+
```
poussa: BLE ?

docs/dgram.md
```diff
@@ -0,0 +1,98 @@
...
+Send data contained in a buffer to remote network node. A subset of
+data in `buf` can be sent using `offset` and `len` parameters. To send
+entire buffer, using values `0` and `buf.length` respectively. See
+`bind()` method description for the format of `ip_addr`. An optional
+callback may be provided, which will be called with the result of send
+operation: either NetworkError object in case of error, or `undefined`
+on success.
+
+### BLE.close
+
```
poussa: ditto

docs/dgram.md
```diff
@@ -0,0 +1,98 @@
...
+### dgram.createSocket
+
+`DgramSocket createSocket(string type);`
+
+Create a datagram socket of given type, which must be `'udp4'` or `'udp6'`.
+
+### DgramSocket.on
+
+`void on(string event, RecvCallback callback);`
+
```
poussa: Would it be possible to support the error event also since I anticipate it will be very useful given the state of the Zephyr net stack and this new API.

src/zjs_dgram.c
```diff
@@ -0,0 +1,334 @@
...
+
+    if (user_data) {
+        jerry_value_t args[1] = {ZJS_UNDEFINED};
+        if (status != 0) {
+            char errbuf[8];
+            snprintf(errbuf, sizeof(errbuf), "%d", status);
+            args[0] = zjs_standard_error(NetworkError, errbuf);
+            // We need error object, not error value (JrS doesn't allow to
+            // pass the latter as a func argument).
+            jerry_value_clear_error_flag(&args[0]);
```
grgustaf: Interesting, with this comment and this line I think you understand something important that I don't understand. :) Please chat with me about this in IRC at some point.

src/zjs_dgram.c
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        zjs_callback_id id = zjs_add_callback_once((jerry_value_t)user_data,
+                                      ZJS_UNDEFINED,
```
grgustaf: indentation - normally we line these up after the ( above.

grgustaf: indentation - normally we line these up after the ( above.

src/zjs_dgram.c
```diff
@@ -0,0 +1,334 @@
...
+}
+
+static jerry_value_t zjs_dgram_sock_send(const jerry_value_t function_obj,
+                                         const jerry_value_t this,
+                                         const jerry_value_t argv[],
+                                         const jerry_length_t argc)
+{
+    int ret;
+
+    if (argc < 5 || argc > 6)
```
grgustaf: Again we normally only check for the right minimum, and the right types - pending discussion.

Probably should check for argv[0] being object, 1-2 being numbers, etc. I guess zjs_buffer_find on the wrong type would give you NULL like you're checking for, maybe that can be skipped.

src/zjs_dgram.c
```diff
@@ -0,0 +1,334 @@
...
+    return ZJS_UNDEFINED;
+}
+
+static jerry_value_t zjs_dgram_sock_close(const jerry_value_t function_obj,
+                                          const jerry_value_t this,
+                                          const jerry_value_t argv[],
+                                          const jerry_length_t argc)
+{
+    GET_HANDLE(dgram_handle_t, handle);
+    zjs_dgram_free_cb((uintptr_t)handle);
```
grgustaf: I think we maybe need to set handle to NULL In the object so we don't double free? Or I guess while we're doing that, we set the free cb to NULL so it won't be called again when the socket goes out of scope.

src/zjs_dgram.c
```diff
@@ -0,0 +1,334 @@
...
+
+    // create socket prototype object
+    zjs_native_func_t array[] = {
+        { zjs_dgram_sock_on, "on" },
+        { zjs_dgram_sock_send, "send" },
+        { zjs_dgram_sock_bind, "bind" },
+        { zjs_dgram_sock_close, "close" },
+        { NULL, NULL }
+    };
+    zjs_dgram_socket_prototype = jerry_create_object();
```
grgustaf: Should add a cleanup function below to free this global object.

src/zjs_dgram.h
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2017, Linaro Limited.
+
+#ifndef __zjs_dgram_h__
+#define __zjs_dgram_h__
+
+#include "jerry-api.h"
+
+jerry_value_t zjs_dgram_init();
```
grgustaf: We're trying to document new APIs (and slowly update old ones) w/ JSdoc comments here, so follow the standard comments for init/cleanup from e.g. zjs_buffer.h. Not that we actually generate these docs yet. :)

src/zjs_modules.c
```diff
@@ -92,6 +93,9 @@ module_t zjs_modules_array[] = {
     { "uart", zjs_uart_init, zjs_uart_cleanup },
 #endif
 #endif // ZJS_LINUX_BUILD
+#ifdef BUILD_MODULE_DGRAM
+    { "dgram", zjs_dgram_init },
```
grgustaf: The cleanup will need to go in here, too.

grgustaf:
And yes, now seeing your initial comments... I do like your macros. We have had such things come and go and just haven't settled on something universal. Your GET_STR is great for the case where you have a fixed buffer where sizeof works. We should just document that carefully as it moves into zjs_util.h, as I tried to do when I boiled all our existing calls down to zjs_copy_jstring and zjs_alloc_from_jstring. That can be a follow-up patch.

I'll want to try out CHECK() myself a little to get used to it, but seems like a good one too. Another place I'd like to help with macros is the argument checking that we do so much of.

docs/dgram.md
pfalcon: Sorry, used BLE module docs as a template and overlooked some references, let me fix all of that.

pfalcon: fixed

docs/dgram.md
pfalcon: fixed

docs/dgram.md
```diff
@@ -0,0 +1,98 @@
...
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Client Requirements](#requirements)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+The `dgram` API is based on the subset of the
```
pfalcon: fixed

docs/dgram.md
```diff
@@ -0,0 +1,98 @@
...
+[corresponding module](https://nodejs.org/api/dgram.html) in Node.js.
+It allows to send and receive UDP datagrams.
```
pfalcon: fixed

docs/dgram.md
```diff
@@ -0,0 +1,98 @@
...
+### BLE.send
+
+`void send(Buffer buf, unsigned long offset, unsigned long len, int port, string ip_addr, [SendCallback cb]);`
+
+Send data contained in a buffer to remote network node. A subset of
+data in `buf` can be sent using `offset` and `len` parameters. To send
+entire buffer, using values `0` and `buf.length` respectively. See
+`bind()` method description for the format of `ip_addr`. An optional
+callback may be provided, which will be called with the result of send
+operation: either NetworkError object in case of error, or `undefined`
```
pfalcon: fixed

src/zjs_dgram.c
pfalcon: fixed

samples/UDPEchoServ4.js
```diff
@@ -0,0 +1,14 @@
+var PORT = 33333;
```
pfalcon: I see that other examples have copyright... Well, I'm personally kinda against putting copyright on trivial examples, that may only complicate users' lives - either make them bother to deal with process regarding copyrights, or make them violators. But if there's a policy that every file in the project should have a copyright notice, please re-confirm, and I'll add it.

src/zjs_dgram.c
```diff
@@ -0,0 +1,334 @@
...
+    jerry_release_value(rinfo);
+}
+
+static jerry_value_t zjs_dgram_createSocket(const jerry_value_t function_obj,
+                                            const jerry_value_t this,
+                                            const jerry_value_t argv[],
+                                            const jerry_length_t argc)
+{
+    int ret;
+    if (argc != 1)
```
pfalcon: I "don't know" (ahem) what's "JS common practice" on that, but if it has practice like that, it's a dangerous practice (and JS is a dangerous language to develop programs in). In this case, the check above has very specific goal though: https://nodejs.org/api/dgram.html#dgram_dgram_createsocket_type_callback gives the function prototype:

> dgram.createSocket(type[, callback])

callback is optional and is a shortcut for calling `.on("message", callback)` later. This shortcut is not supported by a small implementation like Zephyr.js, and if user will try to use it, they will be properly warned that there's a problem.

src/zjs_dgram.c
pfalcon: Continuing the discussion from the previous case, as you can see, I check something, but not all. IMHO, checking number of args is a little bit more important than checking types, regarding the latter, if users pass wrong arg types, they know that they want trouble ;-).

Well, I'm joking of course, not checking types can be dangerous too, but checking them is quite verbose in Z.js :-I. I'll add them.

src/zjs_dgram.c
pfalcon: Yes, makes sense. I wish I knew for sure when it's safe to reset native handle though. Let me add it after the line you commented on (i.e. at the very end of close() method).

pfalcon: Btw, sorry, but I don't see any other module doing that (resetting native handle explicitly). Let's just assume that they don't have explicit "close" methods and wait to be garbage-collected...

pfalcon: I lied, zjs_aio.c is the only module which does that, ok.

src/zjs_dgram.c
pfalcon: Will do if you guys think it's important thing to do for embedded implementation.

grgustaf: In our "ashell" code, we allow you to run JS files dynamically loaded from the filesystem and such. That has to clean up the whole JS state before running the next script, so it's important for that case. You do bring up a good point that for the final production image this would be wasted space, so maybe we should have a way to build that would drop all the cleanup functions. But they're probably a few bytes each at this point.

src/zjs_dgram.c
grgustaf: Well, ignoring extra arguments is good enough for C... :)

I see, I didn't realize you were explicitly not supporting that. In that case, it's fine I guess. You might consider returning NOTSUPPORTED_ERROR() if argc >= 2.

samples/UDPEchoServ4.js
grgustaf: I guess it would help me, so we don't wrongly slap an Intel copyright on it at some point.

pfalcon: As discussed on IRC, as I'm contributing from my company, to a product of another company, my interpretations are irrelevant, and I'll just follow existing practice of the codebase. Fixed.

pfalcon:
So, as I fix issues raised in review comments, I add new commits, as was requested previously, instead of squashing my changes into existing ones right away.

But the squashing would need to happen eventually, and for the multitude of "fix" commits not sneaking in unsquashed, I purposedly don't sign-off them/use different commit message format.

src/zjs_dgram.c
pfalcon: I see, thanks for the hint. ashell is still on my todo to look at (I'm just afraid it doesn't work on anything but arduino_101, and I'll recurse into fixing that ;-) ).

pfalcon: Fixed.

src/zjs_modules.c
pfalcon: Fixed

src/zjs_dgram.h
pfalcon: Fixed

docs/dgram.md
pfalcon: Regarding state of Z net stack: the biggest issue with it (now that it basically works) exactly that many error conditions aren't detected. Instead, it can just hang (like was due to lack of net buffers, but that's now fixed in Z version to which you guys updated Z.js recently, hurrah! I updated my code for it too), or just work in unexpected ways.

Beyond that, UDP is a simple thing and you can get only so much "errors" with it. Where you can get it, e.g. with send, I implemented error callback (which gets passed to the method). So, testing is the biggest issue here (plus some interactions which aren't obvious from Node.js docs, as we're really entering "callback hell" area with various callbacks in different places).

Anyway, if you suggest it to be done, I can go for that, test by injecting a fault by patching Z (that's what I did for send error callback), and report it as done ;-).

src/zjs_dgram.c
pfalcon: > consider returning NOTSUPPORTED_ERROR() if argc >= 2.

Done.

src/zjs_dgram.c
pfalcon: > Probably should check for argv[0] being object, 1-2 being numbers, etc.

Done. Also added checking of network addr params.

pfalcon:
@grgustaf : So, I believe I addressed all of your comments. Can you please look thru the "fix" commits and let me know if I can squash them back into the main, "content" commits?

And otherwise, on TODO is just adding "error" callback event, as requested by @poussa.

docs/dgram.md
poussa: Noted. At this point, it is more important to get this merged in so folks can start testing and using it. Then we can judge if we need better error reporting or not.

pfalcon: Well, @grgustaf mentioned that you guys are busy this week preparing to OpenIoT presentation, and in few windows I have while preparing for my travel too, I'm happy to progress towards supporting "error" callback, for example I just pushed commit to make it possible to set it (but it's not called yet).

grgustaf:
@pfalcon yeah, fixes look fine, you can squash!

pfalcon:
Ok, squashed.

Issue #656, opened 2017-01-26
JS Callback argument management is cumbersome


pfalcon:
How JerryScript works, is that any jeryr_value_t must be released once its use if finished. That means that to execute any callback which accepts args one must: 1) store all args in an adhocly allocated chunk of memory; 2) have a post function which is passed a pointer to this chunk, which will go thru it to release values stored in it and free chunk itself.

At the same time, callback args are already stored in callback queue ring buffer. And if a callback is a JS callback, we know that stuff stored in the ring buffer is an array of jerry_value_t's. Then, zjs_service_callbacks(), which fetches these args from the ring buffer, is the best place to release them. For niche cases when value of an argument of a callback should not be released upon return, a caller can use jerry_acquire_value() on it.

grgustaf:
Yeah, I just went through a huge review of all our code apart from OCF stuff (which we're going to look at today) reviewing for these exact problems. See commit 8d6571a. I did discover (and fix!) the exact issue you're talking about, but it looks like I left that out of the final patch. I suspect it's sitting on my PC at home, so I'll plan to check that tonight.

grgustaf:
This should be resolved by #657 and #661.

pfalcon:
Thanks!

PR #661, opened 2017-01-27
[callbacks] Correct lifetime of jerry_value_t items in callbacks

grgustaf:
Enforce rule that ring buffer owns JrS values stored in it. They are
acquired in zjs_signal_callback and released in zjs_service_callbacks.

Various other minor cleanups.

This was inadvertently left out of my patch in commit 8d6571a.

### DIFF
--- a/src/zjs_buffer.c
+++ b/src/zjs_buffer.c
@@ -1,4 +1,5 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2016-2017, Intel Corporation.
+
 #ifdef BUILD_MODULE_BUFFER
 
 #include <string.h>
@@ -371,7 +372,7 @@ static jerry_value_t zjs_buffer(const jerry_value_t function_obj,
                                 const jerry_length_t argc)
 {
     // requires: single argument can be a numeric size in bytes, an array of
-    //             uint8, or a string
+    //             uint8s, or a string
     //  effects: constructs a new JS Buffer object, and an associated buffer
     //             tied to it through a zjs_buffer_t struct stored in a global
     //             list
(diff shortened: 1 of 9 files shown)


grgustaf:
@pfalcon, here's the patch I was talking about. There's a section like yours but I do also specifically acquire values for the ring buffer and add to the documentation for zjs_signal_callback to clarify this. Please let me know if it makes sense. The idea is follow the pattern for [JerryScript API reference counting](http://jerryscript.net/reference-counting/)

@jprestwo, please review since I messed with your callbacks.

grgustaf:
This should resolve #656 but I'll review @pfalcon's version again tomorrow to see how it differs before I merge anything.

pfalcon:
Well, as I mentioned on IRC, I would consider that ... umm, inefficiency to acquire extra reference while putting a value into ring buffer, while then immediately requiring caller to release its original reference - letting this original reference to be transferred seemed pretty natural. I actually grepped codebase again to see if you actually missed adding such jerry_release_value's around. But well, there aren't many cases where callback arguments are created on the fly, literally the only case is zjs_aio.c:146, and well, it already has jerry_release_value(). So well, looking at from this side, maybe indeed it makes sense to maintain this simple rule that if some code created/acquired value, the same code should release it (albeit it costs more code bytes and maybe more memory).

P.S. Did you intend to include zjs_buffer changes, copyright year and code style reformattings in this patch? They complicate a review.

pfalcon:
Well, I guess good simile here is how a normal function call is made - with it, all arguments still belong to a caller which must release them. Apparently, it makes good sense to follow the same pattern for callbacks. (I'd still make it transfer references if it were my project, but please don't follow a bad advice ;-) ).

grgustaf:
@pfalcon, yeah I do prefer what you're saying but I think we'll have less trouble in the long run if we minimize the exceptional cases. If JerryScript was my project, then I would do it differently across the board if possible. :)

And yeah I did intend to include that stuff, but you're right there too - normally I like to have small single-purpose patches. This all came out of one effort though, which was to review the entire codebase for jerry_value_t lifetime issues. And some of them required changes in more than one file to balance the refcounting, so I didn't think it was worth teasing apart a dozen patches out of it all. But I left this portion out of my big patch.

grgustaf:
The copyright updates are meant to be done per-file when we actually change the file in 2017. The ones I update here are files that I changed in the big patch that this was supposed to be a part of. That's why they seem so non-sensical here. I'm hoping two wrongs will make a right.

src/zjs_buffer.c
```diff
@@ -382,46 +383,48 @@ static jerry_value_t zjs_buffer(const jerry_value_t function_obj,
...
         return new_buf;
-    } else {
-        // if passed string, convert to char array and copy to buffer
+    }
+    else {
+        // treat string argument as initializer
         jerry_size_t size = 0;
         char *str = zjs_alloc_from_jstring(argv[0], &size);
         if (!str) {
             return zjs_error("zjs_buffer: could not allocate string");
```
brianjjones: Need to zjs_free(str); before returning

brianjjones:
lgtm +1

Issue #651, opened 2017-01-24
compile warnings when building samples/I2C.js


grgustaf:
After the recent reorganization to I2C there are a bunch of compile warnings when I do:

`$ make JS=samples/I2C.js`

Please fix those and make sure we're doing the right thing with error handling. I started to look at it myself and thought I'd better leave it to the expert.

brianjjones:
@grgustaf The fix can be found in PR https://github.com/01org/zephyr.js/pull/663

brianjjones:
@grgustaf this issue has been fixed and can be closed.

grgustaf:
@brianjjones, if I do `make clean; make JS=samples/I2C.js`, I still see two compile warnings:

```
  CC      src/main.o
/home/pepsi/dev/zjs/project/src/main.c: In function 'stop_js_handler':
/home/pepsi/dev/zjs/project/src/main.c:77:5: warning: implicit declaration of function 'zjs_ipm_free_callbacks' [-Wimplicit-function-declaration]
     zjs_ipm_free_callbacks();

...

  CC      src/zjs_i2c.o
/home/pepsi/dev/zjs/project/src/zjs_i2c.c: In function 'zjs_i2c_init':
/home/pepsi/dev/zjs/project/src/zjs_i2c.c:249:5: warning: implicit declaration of function 'zjs_i2c_ipm_init' [-Wimplicit-function-declaration]
     zjs_i2c_ipm_init();
     ^
```

brianjjones:
@grgustaf That's due to to my stop_js changes, nothing to do with I2C.  Fix is ready, see it here -> https://github.com/01org/zephyr.js/pull/666

grgustaf:
@brianjjones yeah, but the problem is that when I build I2C.js there are warnings. It doesn't matter if they have to do with I2C. :)

Do you see that second one about i2c init? That's still there with your new patch, and *does* have to do with I2C. :)

xiaolongx:
No warnings output after executing commands: `make clean; make JS=samples/I2C.js` with commit 21915c0.

grgustaf:
Verified, closing.

PR #663, opened 2017-01-27
[bugfix] Fixing #651 to remove compiler warnings

brianjjones:
Also fixing the return type to be int, since that's what
zephyr returns.

### DIFF
--- a/src/zjs_i2c.c
+++ b/src/zjs_i2c.c
@@ -73,17 +73,23 @@ static jerry_value_t zjs_i2c_read_base(const jerry_value_t this,
     }
 
     if (!burst) {
-        uint8_t error_msg = zjs_i2c_handle_read((uint8_t)bus,
-                                                buf->buffer,
-                                                buf->bufsize,
-                                                (uint16_t)address);
+        int error_msg = zjs_i2c_handle_read((uint8_t)bus,
+                                            buf->buffer,
+                                            buf->bufsize,
+                                            (uint16_t)address);
+        if (error_msg != 0) {
+            ERR_PRINT("i2c_read failed with error %i\n", error_msg);
+        }
     }
     else {
-        uint8_t error_msg = zjs_i2c_handle_burst_read((uint8_t)bus,
-                                                      buf->buffer,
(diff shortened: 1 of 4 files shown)


src/zjs_i2c.c
```diff
@@ -169,13 +175,13 @@ static jerry_value_t zjs_i2c_write(const jerry_value_t function_obj,
...
-    uint8_t error_msg = zjs_i2c_handle_write((uint8_t)bus,
-                                             dataBuf->buffer,
-                                             dataBuf->bufsize,
-                                             (uint16_t)address);
+    int error_msg = zjs_i2c_handle_write((uint8_t)bus,
+                                         dataBuf->buffer,
+                                         dataBuf->bufsize,
+                                         (uint16_t)address);
 
     return jerry_create_number(error_msg);
```
grgustaf: This should really return an error object back to JavaScript instead I think? The API documentation seems to say this function is void so it should return undefined in the success case.

src/zjs_i2c.c
```diff
@@ -73,17 +73,23 @@ static jerry_value_t zjs_i2c_read_base(const jerry_value_t this,
...
-        uint8_t error_msg = zjs_i2c_handle_read((uint8_t)bus,
-                                                buf->buffer,
-                                                buf->bufsize,
-                                                (uint16_t)address);
+        int error_msg = zjs_i2c_handle_read((uint8_t)bus,
+                                            buf->buffer,
+                                            buf->bufsize,
+                                            (uint16_t)address);
+        if (error_msg != 0) {
+            ERR_PRINT("i2c_read failed with error %i\n", error_msg);
```
brianjjones: I'd like to give the user the ability to recover and try again if it should i2c_read fail.  It happens (rarely) but I have seen it happen over long periods of time.  That's why I have an ERR_PRINT vs zjs_error

grgustaf:
+1, merging as is.

src/zjs_i2c.c
grgustaf: They can still catch the error with try/catch.

PR #666, opened 2017-01-31
[compiler] Fixing missing include which causes a compiler warning

brianjjones:


grgustaf:
@brianjjones, I still see one more warning when I do a clean build of that I2C.js, see #651. Can you fix that here too?

brianjjones:
@grgustaf Sure thing.  I don't see any more warnings now, hopefully you don't either!

grgustaf:
+1, the cleanest build I've seen in weeks!

PR #667, opened 2017-02-01
Fix snapshot mode

jimmy-huang:


brianjjones:
LGTM +1

jimmy-huang:
@brianjjones can you check if this patch will break ashell?  I think it might need the parser.

brianjjones:
@jimmy-huang Good call, testing now.  @grgustaf  hold off on pulling this in until I confirm it's ok for ashell.  Thanks.

Makefile
```diff
@@ -15,16 +15,19 @@ VARIANT ?= release
 # JerryScript options
 JERRY_BASE ?= $(ZJS_BASE)/deps/jerryscript
 EXT_JERRY_FLAGS ?= -DENABLE_ALL_IN_ONE=ON
+ifneq ($(DEV), ashell)
 ifeq ($(SNAPSHOT), on)
```
brianjjones: This still won't work for 'make linux'.  I think this check / Flag set need to move to after ifeq ($(TARGET), linux).

brianjjones:
LGTM thanks +1

Makefile
```diff
@@ -36,6 +39,10 @@ PRINT_FLOAT ?= off
 # MAKECMDGOALS is a Make variable that is set to the target your building for.
 TARGET = $(MAKECMDGOALS)
 
+ifeq ($(TARGET), linux)
+	SNAPSHOT = off
+endif
```
grgustaf: This block doesn't catch the "make BOARD=linux" case. (See line 58.) That will still try to build with snapshot.

jimmy-huang: updated with check for BOARD=linux

grgustaf:
+1

PR #669, opened 2017-02-02
[I2C] Fixed broken I2C and Grove LCD

jimmy-huang:
The I2C device driver name for sensor sub-system is "I2C_SS_0", not "I2C_0",
updating both I2C and Grove LCD to use the new driver name


brianjjones:
Tested and verified.  LGTM +1

grgustaf:
+1

PR #670, opened 2017-02-02
Fix sensor

jimmy-huang:


grgustaf:
+1

jimmy-huang:
@kenchris zephyr driver for the bmi160 claims to support up to 3200hz, but the max I was able to set without returning an error was 1600hz.

kenchris:
Yeah 32MHz is the max frequency of the ARC core, and it has to do lots of work so that is still pretty good

grgustaf:
@kenchris, stay away from that guy! http://goo.gl/7nG4mb

Sorry if I'm made this joke before. :)

kenchris:
Hahaha, yeah please disregard my other comment... weekend+beers you know :)

PR #671, opened 2017-02-03
[build] Adjust x86 partition size without impacting bootloader

grgustaf:
This was a big realization: Once Zephyr boots, it doesn't matter where
the bootloader thinks the ARC partition starts. Zephyr on X86 is in
charge of bringing up the ARC core. So by splicing images after they are
built, and flashing the right bits to where the bootloader thinks they
should go, we can change the partition without the bootloader being the
wiser. This removes the need for our 256KB flashpack tricks.

The magic is done with three simple dd commands after the A101 zephyr
build target completes.

This is controlled with a new SIZE argument to make, which defaults to
144 but can range from 144-296.

Also, add tests to the Makefile to sanity check the image sizes. Also,
make arc image get built and flashed along with the x86 image. The user
shouldn't really have to use "make arc" anymore, just "make" or "make
zephyr" and "make dfu" will now flash both images. The "ARC" image may
be a lot bigger because it's really the ARC image plus some portion of
the x86 image.

Update the "make help" command to reflect the removed targets and new
SIZE option.

Remove partition size code from zjs-env.h and prj.conf fragments.

Update READMEs to reflect the new state of affairs.

Also, remove jsrunner which is no longer used and referred to some of
this partition stuff.


grgustaf:
One odd note is that the UART build fails with this patch and claims to overflow ROM by 5K or so. If I build with SIZE=160, it works. But on master that didn't seem to be necessary, so it would be worth investigating why. @jprestwo may be needed, not sure yet.

scripts/jsrunner
```diff
@@ -1,85 +0,0 @@
-#!/usr/bin/env python3
```
brianjjones: Are we still using zjsmaxheap? If so you'll need to edit it as it uses jsrunner

Makefile
```diff
@@ -246,23 +267,26 @@ linux: generate
 
 .PHONY: help
 help:
+	@echo "JavaScript Runtime for Zephyr OS - Build System"
+	@echo
 	@echo "Build targets:"
-	@echo "    zephyr:    Build the main Zephyr target (default)"
-	@echo "    arc:       Build the ARC Zephyr target for Arduino 101"
-	@echo "    all:       Build the zephyr and arc targets"
+	@echo "    all:       Build for either Zephyr or Linux depending on BOARD"
```
jimmy-huang: is make all or make arc still needed?

Makefile
```diff
@@ -47,22 +58,43 @@ $(info Note: ZEPHYR_BASE is set outside the current ZJS tree ($(ZEPHYR_BASE)))
 endif
 endif
 
+ifeq ($(BOARD), arduino_101)
+ARC = arc
```
jimmy-huang: indent?

Makefile
```diff
@@ -9,6 +10,16 @@ ifndef ZJS_BASE
...
 endif
 
+# SIZE can't be less than the 144KB physically allocated for x86
+ifeq ($(shell test $(SIZE) -lt 144; echo $$?), 0)
+$(error SIZE must be at least 144)
+endif
+
+# SIZE can't be more than 296KB, the total size of x86 + arc partitions combined
+ifeq ($(shell test $(SIZE) -gt 296; echo $$?), 0)
+$(error SIZE must be no higher than 296)
```
jimmy-huang: indent?

Makefile
```diff
@@ -9,6 +10,16 @@ ifndef ZJS_BASE
 $(error ZJS_BASE not defined. You need to source zjs-env.sh)
 endif
 
+# SIZE can't be less than the 144KB physically allocated for x86
+ifeq ($(shell test $(SIZE) -lt 144; echo $$?), 0)
+$(error SIZE must be at least 144)
```
jimmy-huang: indent?

brianjjones:
ashell seems broken, more specifically I can't access the filesystem.  Everything else about ashell works fine, I can even parse JrS from the command line and see output.  But if I try and write or read from the FS I get gibberish.  I only have one 101 that is in 144 mode though, so perhaps we can test with another board Monday to see if that particular one has a faulty memory chip for the FS.

Makefile
grgustaf: We haven't been doing that at the file scope, if you look around. In Emacs at least it defaults to using Tab in a Makefile, and if you indent with Tab here it breaks it because that has special meaning. Apparently if you type four spaces manually it does work, but it doesn't seem like too great an idea to have a mixture of things indented with spaces and tabs in the same file?

Makefile
grgustaf: I removed them from the help because they're kind of an "expert" mode now. A normal user shouldn't use them, because you need your x86 and arc images to be in sync and it's better to just flash them at the same time too. Basically they've become implementation details, not a public "API".

grgustaf:
@brianjjones, yeah please look into that some more. It does work with master though? Let's discuss it in person and see if we can come up with any theories.

grgustaf:
@brianjjones BTW be sure to test something that uses ARC processor, like AIO or I2C I guess?

brianjjones:
@grgustaf It turns out this was due to hardware failure, the flash memory of one of my Arduino 101 is apparently bad.  Anyway, it works with ashell too, LGTM +1

scripts/jsrunner
grgustaf: Nope, added a patch to dump that. Thx.

jimmy-huang:
+1

grgustaf:
Note, @brianjjones did also verify that code using ARC processor still works on ashell as well.

PR #680, opened 2017-02-07
[I2C] Fixing I2C for FRDM-K64f which uses a different bus name

brianjjones:


grgustaf:
+1

PR #689, opened 2017-02-07
[timers] Fixed timers to remove timeout's properly 

jprestwo:

### DIFF
--- a/src/zjs_timers.c
+++ b/src/zjs_timers.c
@@ -67,7 +67,11 @@ static zjs_timer_t* add_timer(uint32_t interval,
     tm->repeat = repeat;
     tm->completed = false;
     tm->next = zjs_timers;
-    tm->callback_id = zjs_add_callback(callback, this, tm, NULL);
+    if (tm->repeat) {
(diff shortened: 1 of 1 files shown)


grgustaf:
A little more thought revealed this solution I think:
1. Mark the callback as "deleted" with a new flag in the struct.
2. Insert a "flush" or "remove callback" event into the ringbuf with an argument of the callback id. This might be callback #0 assigned at init time?
3. Check for the deleted flag when executing callbacks and skip it.
4. Also check when signaling this callback and skip that / fail so no new ones will be added.
5. When we hit our flush callback, actually remove the callback and now we know all pending callbacks have been skipped over so we are safe to reassign the callback number.

jprestwo:
@grgustaf, 

Per our conversation, I fixed this particular problem with timeouts, but did not implement a "flushing" mechanism. We should have that on our list of things to do but for now this fixes the underlying issue without just putting a band aid on it.

grgustaf:
+1, subtle thing going on here but the callback can be called immediately because the function that added the timer has already run to completion by the time we get into process_events.

PR #690, opened 2017-02-07
Ocf fixes

jprestwo:


src/zjs_event.c
```diff
@@ -542,8 +542,6 @@ bool zjs_trigger_event_now(jerry_value_t obj,
 
     zjs_call_callback(callback_id, argv, argc);
 
-    zjs_free(trigger);
```
grgustaf: So this was a double free?

src/zjs_ocf_ble.c
```diff
@@ -0,0 +1,109 @@
...
+    if (key == BT_STORAGE_ID_ADDR && length == sizeof(id_addr) &&
+        bt_addr_le_cmp(&id_addr, BT_ADDR_LE_ANY)) {
+        bt_addr_le_copy(data, &id_addr);
+        return sizeof(id_addr);
+    }
+
+    return -EIO;
+}
+
+static int char2hex(const char *c, uint8_t *x)
```
grgustaf: The input is hex, not the output, so maybe hex2int? Seems better to me to take char c and maybe just return uint8_t, 0xff could be failure?

Just less tricky looking dropping the pointers.

Oh... I already have a zjs_hex_to_byte in zjs_util.h that is way too similar. Don't have to do this now but let's make a note to self to consolidate these implementations. I also have something doing the reverse in zjs_buffer.c I think.

src/zjs_ocf_ble.c
```diff
@@ -0,0 +1,109 @@
...
+        *x = *c - 'A' + 10;
+    } else {
+        return -EINVAL;
+    }
+
+    return 0;
+}
+
+static int str2bt_addr_le(const char *str, const char *type, bt_addr_le_t *addr)
+{
```
grgustaf: brief comment describing the format we're parsing here might be nice.

src/zjs_ocf_ble.c
```diff
@@ -0,0 +1,109 @@
...
+        if (!(j % 3) && (*str != ':')) {
+            return -EINVAL;
+        } else if (*str == ':') {
+            i--;
+            continue;
+        }
+
+        addr->a.val[i] = addr->a.val[i] << 4;
+
+        if (char2hex(str, &tmp) < 0) {
```
grgustaf: If you follow my advice this would become hex2int(str[0]) != 0xff, I guess.

src/zjs_ocf_ble.c
```diff
@@ -0,0 +1,109 @@
...
+                                         const jerry_value_t argv[],
+                                         const jerry_length_t argc)
+{
+    if (!jerry_value_is_string(argv[0])) {
+        return zjs_error("invalid arguments");
+    }
+    jerry_size_t size = 18;
+    char addr[size];
+    zjs_copy_jstring(argv[0], addr, &size);
+    if (!size) {
```
grgustaf: size could also be non-zero but too short? maybe should check for expected size?

src/zjs_event.c
jprestwo: Yes it was. Only caused a problem on linux.

src/zjs_ocf_ble.c
```diff
@@ -20,21 +20,10 @@ static ssize_t zjs_ble_storage_read(const bt_addr_le_t *addr, uint16_t key,
...
-        *x = *c - 'A' + 10;
-    } else {
-        return -EINVAL;
-    }
-
-    return 0;
-}
-
+/*
+ * Convert a device ID (MAC) string to  Zephyr bluetooth address
```
grgustaf: extraneous space :)

src/zjs_ocf_ble.c
```diff
@@ -20,21 +20,10 @@ static ssize_t zjs_ble_storage_read(const bt_addr_le_t *addr, uint16_t key,
...
+ * e.g. input: "AA:BB:CC:DD:EE:FF"
+ */
 static int str2bt_addr_le(const char *str, const char *type, bt_addr_le_t *addr)
 {
     int i, j;
```
grgustaf: `j` is no longer used. `int i` could go inside the for, doesn't really matter though.

src/zjs_ocf_ble.c
```diff
@@ -44,21 +33,11 @@ static int str2bt_addr_le(const char *str, const char *type, bt_addr_le_t *addr)
         return -EINVAL;
     }
 
-    for (i = 5, j = 1; *str != '\0'; str++, j++) {
-        if (!(j % 3) && (*str != ':')) {
+    for (i = 5; *str != '\0', i >= 0; str+=3, i--) {
```
grgustaf: I think this comma will only check the value after the comma, you really mean &&

PR #696, opened 2017-02-08
[samples] Fix TrafficLight.js logic

grgustaf:
This was the worst traffic light controller in the world, warning you
with a yellow when the light was about to turn green, instead of red.

Note: Yellow is still represented by a green LED. ;)

### DIFF
--- a/samples/TrafficLight.js
+++ b/samples/TrafficLight.js
@@ -41,16 +41,16 @@ green.write(false);
 setInterval(function () {
     elapsed++;
     if (elapsed == 5) {
-        // switch to yellow after 5s
+        // switch to green after 5s
         red.write(false);
-        yellow.write(true);
-    } else if (elapsed == 7) {
-        // switch to green after 2s
-        yellow.write(false);
         green.write(true);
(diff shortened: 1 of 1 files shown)


brianjjones:
+1

samples/TrafficLight.js
```diff
@@ -41,16 +41,16 @@ green.write(false);
...
+        // switch to green after 5s
         red.write(false);
-        yellow.write(true);
-    } else if (elapsed == 7) {
-        // switch to green after 2s
-        yellow.write(false);
         green.write(true);
-    } else if (elapsed == 10) {
-        // switch to red after 3s
+    } else if (elapsed == 8) {
```
jimmy-huang: Do we have to use === for the sake of good practice? not that elapsed will ever turn into another type in this case

grgustaf: Refreshing myself on that, I guess yeah that is the better practice. Let's not do that here, but feel free to submit a separate PR fixing them across all samples and tests.

jimmy-huang:
+1

PR #697, opened 2017-02-08
[build] Fix debug build fail after make pristine

jimmy-huang:
The -DDEBUG_BUILD flag in the snapshot Makefile was causing
issues after make pristine, the snapshot_gen tool do not
need to a debug build anyway, so cleaning up the Makefile to
fix breakage.


grgustaf:
+1

PR #698, opened 2017-02-08
[build] Fix building of SNAPSHOT=on

jimmy-huang:
Backported patch from master that fixes building with SNAPSHOT=on
and make the default build to use the snapshot feature.


grgustaf:
+1

PR #699, opened 2017-02-08
[build] Fixed failed to require JS modules

jimmy-huang:
There's a bug in the build when snapshot is enabled,
it generates snapshot from the JS that's passed, but
it should be generating from the /tmp/zjs.js file which
contains other JS modules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,9 +54,6 @@ PRINT_FLOAT ?= off
 ifeq ($(BOARD), linux)
 	SNAPSHOT = off
 endif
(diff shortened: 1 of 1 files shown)


brianjjones:
Did you mean to now use snapshot with make linux?  If so LGTM +1

jimmy-huang:
@brianjjones no, linux still doesn't use snapshot, I think linux doesn't have to use snapshot, since there's no rom constraints.

grgustaf:
+1

PR #700, opened 2017-02-09
[build] Adjust RAM allocation boundary between x86 and arc

grgustaf:
Akin to recent patch to make ROM allocation configurable in the make
command, do the same for RAM. By default in Zephyr on Arduino 101, the
first 1K of 80K RAM is allocated for special purposes, then the next
24K is allocated for ARC and the next 55K for X86. Now we allow this
to be overridden with the RAM= command line argument. It specifies the
X86 RAM so it can range from 55-79. (It could be made lower if we see
the need for that later.)

Also, change the new SIZE argument to be ROM to distinguish it from
the new RAM option.

Add some output at the beginning of an Arduino 101 build that reports
the RAM/ROM values being used.

Revert part of the other patch that added SIZE argument to UART
builds in Travis/trlite because it doesn't seem necessary after all.

Update the README to reflect the changes, and the travis and trlite
scripts to use the renamed ROM setting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 BOARD ?= arduino_101
-SIZE ?= 144
+RAM ?= 55
+ROM ?= 144
 
 # Dump memory information: on = print allocs, full = print allocs + dump pools
 TRACE ?= off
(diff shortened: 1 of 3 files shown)


brianjjones:
Tested, works for me +1

jimmy-huang:
+1

PR #707, opened 2017-02-10
[trlite] Test each module individually; sensors, snapshot, debug

grgustaf:
Instead of test building with a random smattering of our sample apps
(at one time chosen for their spanning all of our available modules)
just write out one-line .js files to test with that will cause
individual modules to be built.

Also add cases to cover snapshot off as well as on, the a101 sensors,
print float feature, debug build, and ram size change.


jimmy-huang:
+1

PR #708, opened 2017-02-10
[sensor] Proper fix of BMI160 and update Zephyr

jimmy-huang:
The previous patch accidentally downgraded the commit of Zephyr,
and the patch the fixes the bmi160 sensor was not the correct one
since it wasn't a valid test, this is the correct fix and also
revert Zephyr commit to latest.

Also filed bug on Zephyr upstream and submitted fix ZEP-1704


grgustaf:
+1

PR #710, opened 2017-02-10
[build] Fix bug where ZEPHYR_BASE check was no longer happening

grgustaf:
New TARGET_OS variable set to either zephyr or linux, used to decide
whether to check for ZEPHYR_BASE.

Only report RAM/ROM allocation when actually building images for
arduino_101, not make clean/pristine, etc.

Detect linux target even when multiple targets are specified.

Also, only check for RAM and ROM when we're targetting arduino_101.


jimmy-huang:
+1

PR #711, opened 2017-02-10
[sensor] Fixed random hang after when frequency is set too high

jimmy-huang:
This fixes a nasty bug that in the sensor module, it was calling
a jerry_function in the IPM callback which is a interrupt callback,
and it will cause the app to hang randomly when the frequency is
set at a high rate like more than 20Hz.

The lesson here is that we should really make sure we are not calling
into JerryScript engine directly or indirectly through one of our
helper functions inside a interrupt such as IPM message handlers.

### DIFF
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -265,9 +265,11 @@ static void zjs_sensor_onchange_c_callback(void *h, void *argv)
         return;
     }
 
-    zjs_sensor_update_reading(handle->sensor_obj,
-                              handle->channel,
-                              handle->reading);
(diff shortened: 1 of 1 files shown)


grgustaf:
+1

We discussed an idea to add a check to each of our APIs for whether we're in interrupt context, that maybe would be compiled in only in debug mode. This would have caught this case, but not calling jerry_ functions directly.

Another thing might be to just label all interrupt handlers with a special comment that reminds us to be careful when reviewing or editing the code, e.g.:
`// INTERRUPT HANDLER: No JerryScript VM calls!`

PR #713, opened 2017-02-10
Fix sensor readings

jimmy-huang:


grgustaf:
It's a shame to pack in the whole union even when we're doing ambient light sensor, because it's 24 bytes I think but there we only need 8.

grgustaf:
@jimmy-huang, let's fix this to not waste extra space for ambient light sensor per my previous comment.

jimmy-huang:
@grgustaf updated patch.

src/zjs_sensor.c
```diff
@@ -264,21 +280,49 @@ static void zjs_sensor_onchange_c_callback(void *h, void *argv)
...
+        if (xyz) {
+            double xyz[3];
+            xyz[0] = data->reading.x;
+            xyz[1] = data->reading.y;
+            xyz[2] = data->reading.z;
+            zjs_signal_callback(h->id, &xyz, sizeof(xyz));
+        } else {
+            double dval = data->reading.dval;
+            zjs_signal_callback(h->id, &dval, sizeof(dval));
+        }
```
grgustaf: Shouldn't we be able to support both acclerometer and ambient light at the same time? This appears to be losing that ability? Seems like we need to check each sensor handle to see what format it is?

jimmy-huang: each sensor is only a specific type, and the IPM message received from each sensor reading is different

jimmy-huang: just FYI, I tested with both Gyroscope and ambient light sensors, and it works.

grgustaf:
+1

PR #714, opened 2017-02-10
[trlite] Fix to return error value properly

grgustaf:
The script was returning a zero value even on failure, so Travis always
succeeded, which felt warm and fuzzy, but wasn't.

Also, realized try_func is no different than try_command, so it could
be removed.


jimmy-huang:
+1

PR #715, opened 2017-02-10
[general] Wait on semaphore in main loop

jprestwo:

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -101,6 +101,9 @@ int main(int argc, char *argv[])
     jerry_value_t code_eval;
     uint32_t len;
 #endif
+#ifndef ZJS_LINUX_BUILD
+    zjs_loop_init();
+#endif
     jerry_value_t result;
 
     // print newline here to make it easier to find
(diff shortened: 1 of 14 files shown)


poussa:
What is the issue this PR is fixing?

grgustaf:
Discussed elsewhere, we should do some work to demonstrate some facts about how our system works and prove this is superior before merging. In particular, I'd like to see something that logs the timings of a faster timer running in JS, e.g. setInterval with a timeout of 10ms and log when exactly those callbacks occur. I'd like to see e.g. what the average time between callbacks is and their standard deviation. Does this improve with the patch?

We suspected maybe power usage would improve with this patch but based on our limited testing it didn't, so I think we need this kind of timing analysis to see whether we've really made things better.

I suspect that with our old sleep method, we might have slept through an entire tick now and then, causing timer events to double up in the next tick. I'm hoping we'll be able to see that by capturing some data. And then maybe that goes away or is minimized with this semaphore approach.

grgustaf:
@jimmy-huang were you able to confirm a positive impact from this patch?

jimmy-huang:
@jprestwo can you rebase this? I can try it out today

grgustaf:
Well we still can't point to this specifically improving any real-world results, but we believe it is a better approach, so no time like right at the start of a release cycle to merge it.

grgustaf:
+1

PR #716, opened 2017-02-11
Wait to print ringbuf error

grgustaf:


grgustaf:
These depend on #715 and should be rebased after it goes in.

jimmy-huang:
+1

PR #722, opened 2017-02-14
[Grove LCD] Fixed Grove LCD broken after Zephyr update

jimmy-huang:
Zephyr has updated which the previous fix the Grove LCD is no
longer needed and will not run properly if kept, thus removing
the existing config.


poussa:
LGTM. Tested and fixes the issue.

PR #730, opened 2017-02-14
[error] Better JS error printing in main loop

jprestwo:

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -9,6 +9,7 @@
 #endif // ZJS_LINUX_BUILD
 #include <string.h>
 #include "zjs_script.h"
+#include "zjs_util.h"
 
 // JerryScript includes
 #include "jerry-api.h"
@@ -204,6 +205,7 @@ int main(int argc, char *argv[])
     code_eval = jerry_parse((jerry_char_t *)script, len, false);
     if (jerry_value_has_error_flag(code_eval)) {
         ERR_PRINT("JerryScript: cannot parse javascript\n");
+        zjs_print_error_message(code_eval);
         goto error;
     }
 #endif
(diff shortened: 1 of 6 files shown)


src/main.c
```diff
@@ -204,6 +205,7 @@ int main(int argc, char *argv[])
     code_eval = jerry_parse((jerry_char_t *)script, len, false);
     if (jerry_value_has_error_flag(code_eval)) {
         ERR_PRINT("JerryScript: cannot parse javascript\n");
```
grgustaf: Change to "error parsing JavaScript"

src/main.c
```diff
@@ -224,6 +226,7 @@ int main(int argc, char *argv[])
 
     if (jerry_value_has_error_flag(result)) {
         ERR_PRINT("JerryScript: cannot run javascript\n");
```
grgustaf: Change to "Error running JavaScript". (Drop JerryScript)

grgustaf:
A few of the error messages seem interesting, Jimmy thought maybe 10 or more, so they might be worth enabling actually. We should try reproducing a few of them with bad code to see how they might help the user.

And see what the size diff is including them or not, to make the decision.

grgustaf:
OK, I think we should go ahead and take this, and prove later whether those other error messages would be of value to users.

+1

pfalcon:
Thanks for this. I didn't want to comment before I see it in action and see that it's really helpful. Whoa, now you can work with Z.js not built in debug mode (before it was simply impossible to use it otherwise). Let's see how long it takes you guys to report *where* error happened (like, module/line no) ;-).

jprestwo:
@pfalcon:

Unfortunately JerryScript does not seem to store the function/line number for runtime errors. Even in a "full" debug build you do not get these values. Parse/Syntax errors will contain the line and column though. Perhaps when the JerryScript debugging feature is live things will change.

PR #731, opened 2017-02-14
[i2c] Revert previous fixes to fix broken Grove LCD and I2C on Ashell

jimmy-huang:
Previous patch for I2C and Grove LCD were work-arounds for Zehpyr bug,
but not longer needed, the patch had a conflict where both I2C and
Grove LCD were enabled in Ashell, so reverting the patch will make
both I2C and Grove LCD work with ashell and without ashell.


brianjjones:
Works for me +1

grgustaf:
+1

PR #732, opened 2017-02-14
[samples] Make perf module optional in I2C.js

grgustaf:
The ashell build doesn't include perf module so this sample doesn't
run, but the perf line is just for QA so it should be optional.


grgustaf:
@brianjjones, this makes I2C.js work for me in ashell w/o config change, can you review at some point?

jimmy-huang:
+1

PR #734, opened 2017-02-14
[build] Added uglifyjs for catch parse errors when building snapshot

jimmy-huang:
This gives more useful debugging messages to developers to catch
syntax errors in their Javascript.


grgustaf:
+1

PR #739, opened 2017-02-15
[ocf] Device name and BLE address set on command line

jprestwo:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,8 @@ endif
 OCF_ROOT ?= deps/iotivity-constrained
 JS ?= samples/HelloWorld.js
 VARIANT ?= release
-BLE_NAME ?= "Zephyr OCF Node"
+DEVICE_NAME ?= "Zephyr OCF Node"
+BLE_ADDR ?= "none"
 # JerryScript options
 JERRY_BASE ?= $(ZJS_BASE)/deps/jerryscript
 EXT_JERRY_FLAGS ?=	-DENABLE_ALL_IN_ONE=ON \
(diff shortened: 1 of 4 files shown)


src/zjs_ocf_ble.c
```diff
@@ -81,6 +84,20 @@ static jerry_value_t ocf_set_ble_address(const jerry_value_t function_val,
...
 {
+#ifdef ZJS_CONFIG_BLE_ADDRESS
+    static const struct bt_storage storage = {
+            .read = zjs_ble_storage_read,
+            .write = NULL,
+            .clear = NULL,
+    };
+
+    if (str2bt_addr_le(default_ble, "random", &id_addr) < 0) {
+        return zjs_error("bad BLE address string");
```
jimmy-huang: why is this returning zjs_error() when the function is void()?

Makefile
```diff
@@ -43,7 +43,8 @@ endif
 OCF_ROOT ?= deps/iotivity-constrained
 JS ?= samples/HelloWorld.js
 VARIANT ?= release
-BLE_NAME ?= "Zephyr OCF Node"
+DEVICE_NAME ?= "Zephyr OCF Node"
+BLE_ADDR ?= "none"
```
jimmy-huang: should this be labeled "random" instead of none?

src/zjs_ocf_common.c
```diff
@@ -205,7 +205,7 @@ static void issue_requests(void)
 static void app_init(void)
 {
     oc_init_platform("ZJS", NULL, NULL);
-    oc_add_device("/oic/d", "oic.d.zjs", "ZJS Device", "1.0", "1.0", NULL, NULL);
+    oc_add_device("/oic/d", "oic.d.zjs", CONFIG_BLUETOOTH_DEVICE_NAME, "1.0", "1.0", NULL, NULL);
```
jimmy-huang: Would this still work if the user doesn't set the bluetooth name?

Makefile
jprestwo: "none" doesnt get into the C code, its just for the Makefile.ocf_zephyr to check. If it finds "none" it wont define the ZJS_CONFIG_BLE_ADDRESS

src/zjs_ocf_common.c
jprestwo: Yes, its a conditional set in the main Makefile. "Zephyr OCF node"

grgustaf:
+1

PR #740, opened 2017-02-15
[docs] Tweak OCF over BLE doc based on my runthrough today

grgustaf:

### DIFF
--- a/docs/ocf-ble.md
+++ b/docs/ocf-ble.md
@@ -121,7 +121,9 @@ You can then do discovery from the browser or using one of the test scripts:
 
 2. Test script method
 
-  Run the oic-get script from the iot-rest-api-server test/ directory.
+  Use the oic-get script from the iot-rest-api-server test/ directory.
+
+  Run resource discovery with the /res path:
 
   ```
   $ ./oic-get /res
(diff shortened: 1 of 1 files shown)


grgustaf:
@jimmy-huang @jprestwo please review

docs/ocf-ble.md
```diff
@@ -32,21 +32,27 @@ is advertising over BLE. The following steps will create that connection which
...
-2. Reset HCI
+3. Reset HCI
 
   ```
   hciconfig hci0
   hciconfig hci0 reset
   ```
 
-3. Look for your Arduino 101 advertisement
+4. Look for your Arduino 101 advertisement
```
jimmy-huang: If you rebuild the app reboot the device, you don't have to repeat steps 1-4, I assume, right?   The user can just start from the connect step below, so it might be good to mention that.

grgustaf: @jprestwo can you answer authoritatively?

jprestwo: @jimmy-huang, @grgustaf: Yes, you should only have to do the final echo connect command to reconnect.

grgustaf: @jimmy-huang okay - see my latest patch for an extended comment on the subject.

jimmy-huang:
+1

docs/ocf-ble.md
```diff
@@ -82,44 +97,62 @@ seconds for it to appear.
...
+  HTTP: 200
+  $ ./oic-get /a/light?di=31bf6309-8ebf-4309-5fbf-630930c06309
+  10.237.72.146:80 /a/light?di=31bf6309-8ebf-4309-5fbf-630930c06309
+  {"properties":{"state":false,"power":10}}
+  HTTP: 200
+  $ ./oic-get /a/light?di=31bf6309-8ebf-4309-5fbf-630930c06309
+  10.237.72.146:80 /a/light?di=31bf6309-8ebf-4309-5fbf-630930c06309
+  {"properties":{"state":true,"power":10}}
+  HTTP: 200
+  ```
```
poussa: Why do we have the same command 4 times? One should be enough.

pfalcon: Looks like it blinks a thing? (state: false, then state: true, repeated)

poussa: That is so true. My bad.

pfalcon: Ah, that's probably output anyway...

pfalcon: (Just trying to look at this OCF thing, which is on my todo for a while.)

poussa: if the led is blinking of its own or some other client is changing its state, this is possible output.

grgustaf: I think you're right, that could use some description. I'll give it a stab and run it by @jprestwo for correctness.

grgustaf: @poussa, @pfalcon please see if this clarifies the output?

jprestwo: OcfServer.js automatically toggles the output on every GET, so we did not need to show it 4x. This is much better.

jprestwo:
@grgustaf, looks good, just add rfkill and it will be ready.

PR #742, opened 2017-02-15
[ashell] Adding code to free cb memory and change cleanup order

brianjjones:
This adds a method that ensures all callbacks are freed.  It
also changes the order of some of the cleanup functions to
ensure stability when starting / stopping apps using ashell.

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -326,6 +326,15 @@ void zjs_remove_callback(zjs_callback_id id)
     }
 }
 
+void zjs_remove_all_callback()
+{
+   for (int i = 0; i < cb_size; i++) {
+        if (cb_map[i]) {
+            zjs_remove_callback(i);
+        }
+    }
+}
+
 void zjs_signal_callback(zjs_callback_id id, const void *args, uint32_t size)
 {
     DBG_PRINT("pushing item to ring buffer. id=%d, args=%p, size=%lu\n", id,
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -326,7 +326,7 @@ void zjs_remove_callback(zjs_callback_id id)
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/zjs_callbacks.c
```diff
@@ -326,6 +326,15 @@ void zjs_remove_callback(zjs_callback_id id)
     }
 }
 
+void zjs_remove_all_callback()
```
grgustaf: maybe callbacks?

grgustaf:
Other than that, LGTM.

src/zjs_callbacks.c
```diff
@@ -326,6 +326,15 @@ void zjs_remove_callback(zjs_callback_id id)
     }
 }
 
+void zjs_remove_all_callbacks()
+{
+   for (int i = 0; i < cb_size; i++) {
+        if (cb_map[i]) {
+            zjs_remove_callback(i);
```
jimmy-huang: Does this code clean up the ringbuffer? maybe @jprestwo can take a look here.

brianjjones: Probably not, but I don't think ring buffer was the issue I was seeing.  cb_map was the item making zjs_malloc calls.  Not saying ring buffer couldn't also be an issue however.

poussa:
I was testing this. It seems to fix the hang issue (#745) or at least improve it significantly. However, there seems to be on new issue, which I guess is in this PR.

Using the sample app from the (#745), on first run you get one callback when you press the button. When you upload the app again (2nd run), you get two callbacks when you press the button. And so on. If you upload the app 10 times, you get 10 callbacks per button press.

But it did not hang!

brianjjones:
@poussa Strange, especially since (in theory) I'm deleting all the callbacks :)

Thanks for trying it out, I'll try and reproduce and fix it so that its a complete fix.

PR #744, opened 2017-02-16
[ble] Enable BLE from ble_init function

grgustaf:
We used to enable it from outside the BLE module, but on ashell
this meant the module hadn't been initialized yet, because there is
no JS running with a require() call initially.

This fixed a spurious zjs_call_callback() error message on ashell
startup, and also allows apps like BLE.js to be run from ashell!


jimmy-huang:
+1

PR #747, opened 2017-02-16
[ocf] Fixed OCF debug flags

jprestwo:
- OCF debugging was enabled by default, now controlled by VARIANT
 - Fixed two warnings in OCF

### DIFF
--- a/src/zjs_ocf_ble.c
+++ b/src/zjs_ocf_ble.c
@@ -37,7 +37,7 @@ static int str2bt_addr_le(const char *str, const char *type, bt_addr_le_t *addr)
     }
 
     for (i = 5; i >= 0; str+=3, i--) {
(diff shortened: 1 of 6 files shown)


src/zjs_ocf_ble.c
```diff
@@ -37,7 +37,7 @@ static int str2bt_addr_le(const char *str, const char *type, bt_addr_le_t *addr)
     }
 
     for (i = 5; i >= 0; str+=3, i--) {
-        if (!zjs_hex_to_byte(str, &tmp)) {
+        if (!zjs_hex_to_byte((char*)str, &tmp)) {
```
grgustaf: The right fix to this one is really to make zjs_hex_to_byte take a const char * there.

grgustaf:
I made sure that const char * fix won't hurt any other callers, so go ahead and make that change.

After that, it's +1.

PR #748, opened 2017-02-16
[ble] Remove hard-coded Bluetooth MAC address

jimmy-huang:
Instead of a hard-coded fake Bluetooth MAC address, retrieve
the real MAC address when client connects or disconnects.


grgustaf:
+1

PR #749, opened 2017-02-16
[promise] Fixed an acquire problem in promises

jprestwo:
- Refactored promises

### DIFF
--- a/src/zjs_promise.c
+++ b/src/zjs_promise.c
@@ -7,9 +7,6 @@
 #include "zjs_callbacks.h"
 
 typedef struct zjs_promise {
-    jerry_value_t then;         // Function registered from then()
-    jerry_value_t catch;        // Function registered from catch()
-    jerry_value_t this;         // 'this' object for this promise
     zjs_callback_id then_id;    // Callback ID for then JS callback
     zjs_callback_id catch_id;   // Callback ID for catch JS callback
     void* user_handle;
@@ -22,20 +19,9 @@ zjs_promise_t *new_promise(void)
 {
     zjs_promise_t *new = zjs_malloc(sizeof(zjs_promise_t));
     memset(new, 0, sizeof(zjs_promise_t));
-    new->then = ZJS_UNDEFINED;
-    new->catch = ZJS_UNDEFINED;
     new->catch_id = -1;
     new->then_id = -1;
     return new;
-}
-
-// Dummy function for then/catch
-static jerry_value_t null_function(const jerry_value_t function_obj,
-                                   const jerry_value_t this,
-                                   const jerry_value_t argv[],
-                                   const jerry_length_t argc)
-{
-    return ZJS_UNDEFINED;
 }
 
 static void post_promise(void* h, jerry_value_t* ret_val)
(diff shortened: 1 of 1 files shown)


src/zjs_promise.c
```diff
@@ -175,27 +148,20 @@ void zjs_reject_promise(jerry_value_t obj, jerry_value_t argv[], uint32_t argc)
...
+        handle->catch_id = zjs_add_callback_once(ZJS_UNDEFINED,
+                                                 obj,
                                                  handle,
                                                  post_promise);
 
         zjs_signal_callback(handle->catch_id, argv,
                             argc * sizeof(jerry_value_t));
 
+        zjs_remove_callback(handle->then_id);
         DBG_PRINT("rejecting promise, obj=%lu, catch_id=%d, argv=%p, nargs=%lu\n",
```
grgustaf: Add a newline to match code above.

src/zjs_promise.c
```diff
@@ -138,29 +117,23 @@ void zjs_fulfill_promise(jerry_value_t obj, jerry_value_t argv[], uint32_t argc)
...
-            jerry_release_value(handle->then);
-            handle->then = jerry_create_external_function(null_function);
-        }
-
-        handle->then_id = zjs_add_callback_once(handle->then,
-                                                handle->this,
+        handle->then_id = zjs_add_callback_once(ZJS_UNDEFINED,
+                                                obj,
                                                 handle,
                                                 post_promise);
```
grgustaf: OK, I'm pretty sure this was wrong before and is still wrong now. You can't overwrite then_id in fulfill, because what if we had a true asynchronous promise and then is set and later on the promise is fulfilled.

If I'm right, this means we need some better tests written - one that truly exercises both the lame "synchronous" case and a real asynchronous case, e.g. with a timer.

src/zjs_promise.c
```diff
@@ -138,29 +117,23 @@ void zjs_fulfill_promise(jerry_value_t obj, jerry_value_t argv[], uint32_t argc)
...
 
         zjs_signal_callback(handle->then_id, argv,
                             argc * sizeof(jerry_value_t));
```
grgustaf: Since you create the callback and signal it right away, the only way it will get "edited" is if the then() happens call happens in between now and when the signal is processed. Which is just the weird synchronous case, not the "normal" case, I think.

src/zjs_promise.c
```diff
@@ -71,14 +54,10 @@ static jerry_value_t promise_resolve(const jerry_value_t function_obj,
         return zjs_error("promise handle not found");
     }
     if (fulfill) {
-        jerry_release_value(handle->then);
-        handle->then = jerry_acquire_value(argv[0]);
-        zjs_edit_js_func(handle->then_id, handle->then);
+        zjs_edit_js_func(handle->then_id, argv[0]);
         handle->then_set = 1;
```
grgustaf: I don't think then_set and catch_set are being read anymore, only written, so remove them? Also, we could probably have always done this test as "then_id != -1" instead of having a separate boolean?

jprestwo:
@grgustaf, I made the requested changes in a new branch and pushed a different PR. The new PR also has the testing code we talked about. We can go ahead and close this.

PR #752, opened 2017-02-17
[docs] API docs for OCF server

jprestwo:

### DIFF
--- a/docs/ocf.md
+++ b/docs/ocf.md
@@ -1,0 +1,107 @@
+ZJS API for OCF
+==================
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+ZJS provides OCF Server API's which allow communication using the OCF networking
+protocol.
+
+Web IDL
+-------
+This IDL provides an overview of the interface; see below for documentation of
+specific API functions.
+
+```javascript
+// require returns an OCFObject
+// var ocf = require('ocf');
+
+dictionary ResourceInit {
+    string resourcePath;      // OCF resource path
+    string[] resourceTypes;   // List of resource types
+    string[] interfaces;      // List of interfaces for resource types
+    boolean discoverable;     // Is resource discoverable
+    boolean observable;       // Is resource observable
+    boolean secure;           // Is resource security enabled
+    boolean slow;             // Is resource a slow reader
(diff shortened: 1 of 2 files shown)


poussa:
+1

docs/ocf.md
```diff
@@ -0,0 +1,107 @@
...
+`Promise<OCFResource> register(ResourceInit init);`
+
+Register a new resource with the server.
+
+The `init` contains the resource initalization information.
+
+Returns a promise which resolves to an `OCFResource`.
+
+### OCFServer.on('retrieve', RetrieveCallback listener)
+
```
zolkis: This should be a callback, and not an event. We want exactly 1 handler for a server/request, since we need exactly 1 reply on protocol level.

docs/ocf.md
```diff
@@ -0,0 +1,107 @@
...
+Register an `onretrieve` listener. This event will be emitted when a remote
+client retrieves this servers resource(s).
+
+### OCFServer.on('update', UpdateCallback listener)
+
```
zolkis: Same, should be a callback, not an event.

docs/ocf.md
```diff
@@ -0,0 +1,107 @@
...
+Register an `onupdate` listener. This event will be emitted when a remote
+client updates this servers resource(s).
+
+### OCFRequest.respond
```
zolkis: Also need respondWithError(Error error), or equivalent.

docs/ocf.md
```diff
@@ -0,0 +1,107 @@
...
+// listener for onupdate event
+callback UpdateCallback = void (OCFRequest request);
+
+```
+
+API Documentation
+-----------------
+
+### setBleAddress
+`void setBleAddress(string address);`
```
zolkis: I wonder if this is the right place to host this? It belongs more to the Board API than to the OCF API, if I get the use case.

docs/ocf.md
```diff
@@ -0,0 +1,107 @@
...
+    OCFServer server;         // OCF server object
+};
+
+// listener for onretrieve event
+callback RetrieveCallback = void (OCFRequest request, boolean observe);
+// listener for onupdate event
+callback UpdateCallback = void (OCFRequest request);
+
+```
+
```
zolkis: OCF client functionality seems to be missing.

zolkis: I wonder if you could use the definitions in 
https://github.com/01org/iot-js-api/tree/master/api/ocf
that is, aligned with iotivity-node?

PR #754, opened 2017-02-17
[gpio] Fix GPIO double signalling onchange with ashell

jprestwo:

### DIFF
--- a/src/zjs_gpio.c
+++ b/src/zjs_gpio.c
@@ -52,7 +52,10 @@ typedef struct gpio_handle {
     uint32_t last;
     uint8_t edge_both;
     bool closed;
+    struct gpio_handle* next;
 } gpio_handle_t;
+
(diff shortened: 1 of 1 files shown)


brianjjones:
While this is a totally practical option, I would prefer we use https://github.com/01org/zephyr.js/pull/755 instead.  It gets JrS to clean up the GPIO object when jerry_cleanup occurs, which means we aren't required to keep track of all the GPIO handles. JrS will take care of it by calling the zjs_gpio_free_cb we provide jerry_set_object_native_handle.

jprestwo:
@brianjjones, I agree. This PR was a quick fix that just masks the problem. If #755 does allow JrS to free up the GPIO handle then we should use that.

grgustaf:
@jprestwo, I think this is no longer needed after @brianjjones's patch? If so, please close.

Issue #745, opened 2017-02-16
[ashell] Hangs too easily when uploading the JS code multiple times


poussa:
When you upload the (same or different) JS application multiple times, the ashell hangs quite easily. First run is always golden, but then 2-5 run very often hangs it.

For example, the following app works always first time but hang on 2 or 3 time of running.

```js
var pins = require("arduino101_pins");
var gpio = require("gpio");

var button = gpio.open({ pin: pins.IO2, direction: 'in', edge: 'rising' });

button.onchange = function(event) {
    console.log("event: " + event.value);
}
```

qiaojingx:
Upload [test-a101-pins.js](https://github.com/01org/zephyr.js/blob/master/tests/test-a101-pins.js) multiple times. ashell will hang.
Commit: 7c5ff3c

qiaojingx:
Upload the code mentioned in description and [test-a101-pins.js](https://github.com/01org/zephyr.js/blob/master/tests/test-a101-pins.js) for 5 times individually. This issue is not reproducible.
Commit: 76c4663

brianjjones:
With the latest changes this should be fixed.  Let me know if you still see it.

grgustaf:
This was fixed by #742 and #755. @poussa please confirm and close.

grgustaf:
I think QA confirmed, closing.

PR #755, opened 2017-02-17
[gpio] Fix to allow JerryScript to release onchange callback

brianjjones:
This fixes the issue where GPIO was not having it's zjs_gpio_close function called because we still had a reference to the onchange function when jerry_cleanup gets called.  Now the garbage collector will clean the GPIO object like it should.


poussa:
Tested and works as expected. No more double (and more) callbacks. Fixes https://github.com/01org/zephyr.js/issues/745 

+1

grgustaf:
+1

PR #763, opened 2017-02-25
[build] Fixed invalid C string genereated when SNAPSHOT=off

jimmy-huang:
Fixed a bug in the convert.sh script where it could add a newline
to the generated C string when the input JS contains a newline at
the end of the file after running through uglifyjs.


grgustaf:
+1, newer uglify adds a newline to the end of the output.

Issue #761, opened 2017-02-23
[udp] Cannot bind other ip except '192.0.2.1' and '2001:db8::1'


cuiyanx:
**JS code:**
IPv4:
```
var HOST = "192.0.2.10";
var PORT = 33333;
server.bind(PORT, HOST);
```
IPv6:
```
var HOST = "2001:db8::10";
var PORT = 33333;
server.bind(PORT, HOST);
```

**error message:**

![selection_003](https://cloud.githubusercontent.com/assets/22387607/23247396/d9fb3c78-f9d4-11e6-9dc6-41f9c71d42f3.png)

pfalcon:
That's expected. You can bind only to an address which belongs to one of local interfaces (because you bind to a local interface(s) in the first place), and addresses you have in the title are currently hardcoded addresses for "dgram" modules (them being hardcoded is another known issue, and it's again not clear how to make them configurable, requires more discussion and then work on that).

cuiyanx:
OK, thanks to know.

haoxli:
@pfalcon if it's hardcoded, it's better to update the API description in [dgram.md](https://github.com/01org/zephyr.js/blob/master/docs/dgram.md)

> void bind(int port, string ip_addr);
>
> Bind socket to a local address and port. This is required operation for server-side sockets, i.e. sockets which wait and receive data from other network nodes. ip_addr must be a string representing an IP address. This module does not support domain name resolution, **so only IP addresses are allowed. An example of IPv4 address: '192.0.2.1', IPV6: '2001:db8::1'.**

From my understanding, **'192.0.2.1'** just is an example for IPv4, all other IP addresses should be allowed. How do you think of updating it to a notice like ` Now only '192.0.2.1' is available for IPv4 address, and '2001:db8::1' for IPV6` ?

pfalcon:
> all other IP addresses should be allowed.

Indeed, they are allowed - as long as a users follows (basic, well-known (?)) principles of how bind works. Perhaps, docs should provide hints of these principles though.

So, I submitted https://github.com/01org/zephyr.js/issues/765 regarding how to handle networking config in general, and will ponder how to unconfuse these matters on the levels of API, examples, and docs, patches will follow.

PR #767, opened 2017-02-27
[dgram] Clarify description of an address which should be used in bind().

pfalcon:
Also, tested wildcard addresses to work for bind, so included in the
docs too.

### DIFF
--- a/docs/dgram.md
+++ b/docs/dgram.md
@@ -75,9 +75,13 @@ Registers a callback. The `event` may be one of the following:
 
 Bind socket to a local address and port. This is required operation for
 server-side sockets, i.e. sockets which wait and receive data from other
-network nodes. `ip_addr` must be a string representing an IP address.
-This module does not support domain name resolution, so only IP addresses
-are allowed. An example of IPv4 address: `'192.0.2.1'`, IPV6: `'2001:db8::1'`.
+network nodes. `ip_addr` must be a string representing an IP address of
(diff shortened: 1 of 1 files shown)


pfalcon:
@haoxli : Following discussion in https://github.com/01org/zephyr.js/issues/761#issuecomment-282237291 .

My general concern with "if it's hardcoded, it's better to update the API description in docs" is that docs tend to be updated much less frequently, and then outdated adhoc information stays in docs and confuses users. I followed the suggestion though, but at least added a hint to users that they shouldn't rely that any hardcodings would live forever.

I also considered to switch samples to use wildcard bind address, but it's the typical of such networking samples: it's easy (also, dangerous, a known security risk) to bind to 0.0.0.0, but what user is interested in is the address to connect to. In the samples, the address is used to both bind and print to the user, so, I leave them as is.

haoxli:
> docs tend to be updated much less frequently, and then outdated adhoc information stays in docs and confuses users. I followed the suggestion though, but at least added a hint to users that they shouldn't rely that any hardcodings would live forever.

Yes, you are right. Consider that maybe we can put hardcode information in a separate paragraph as a note, such as Node.js [File System](https://nodejs.org/api/fs.html) doc for special case:
> Note: The behavior of fs.open() is platform specific for some flags. As such, opening a directory on OS X and Linux with the 'a+' flag - see example below - will return an error. In contrast, on Windows and FreeBSD, a file descriptor will be returned.

Just a suggestion! :)

grgustaf:
+1

PR #768, opened 2017-02-27
[ocf] Added OCF device and platform objects

jprestwo:

### DIFF
--- a/samples/OcfServer.js
+++ b/samples/OcfServer.js
@@ -1,23 +1,23 @@
 // Copyright (c) 2017, Intel Corporation.
 
-var ocf = require('ocf');
+var ocf = require("ocf");
 var server = ocf.server;
 
 ocf.device = {
 	name: "MyOCFServer",
-	coreSpecVersion: '2.0',
-	dataModels: '2.5'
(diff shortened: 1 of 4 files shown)


src/zjs_ocf_server.c
```diff
@@ -463,6 +463,8 @@ static jerry_value_t ocf_register(const jerry_value_t function_val,
 
     resource = new_server_resource(resource_path);
 
+    zjs_ocf_start();
```
jimmy-huang: Should check return value of ocf_start() if errored

samples/OcfServer.js
```diff
@@ -1,6 +1,24 @@
 // Copyright (c) 2017, Intel Corporation.
 
-var server = require('ocf').server;
+var ocf = require('ocf');
+var server = ocf.server;
+
+ocf.device = {
+	name: "MyOCFServer",
```
jimmy-huang: Did we ever standardized the use of single quote or double quotes?  Might be good idea to keep them consistent.

jimmy-huang:
+1

PR #773, opened 2017-02-28
[error] Make missing JS module a NotSupportedError in linux build

grgustaf:
This is to match behavior of a module not being built into a real
build, so that test-error.js will see the same error type.


brianjjones:
+1

jimmy-huang:
+1

Issue #735, opened 2017-02-15
[Error] No error thrown when require invalid module on Linux platform


qiaojingx:
When running [test-error.js](https://github.com/01org/zephyr.js/blob/master/tests/test-error.js) on Linux platform with commit 3f8eb85 via command `zephyr.js$ ./outdir/linux/release/jslinux tests/test-error.js` 

Expected result:  Error is thrown when require invalid module just like on Arduino 101 and FRDM-K64F.

Actually result: No error was thrown when require invalid module, only got error log.

Serial log:
```
Testing error types...

PASS - SecurityError works as expected
PASS - NotSupportedError works as expected
PASS - SyntaxError works as expected
PASS - TypeError works as expected
PASS - RangeError works as expected
PASS - TimeoutError works as expected
PASS - NetworkError works as expected
PASS - SystemError works as expected
PASS - calling require with an integer
PASS - calling require with a long name

src/zjs_script.c:17 zjs_read_script():
[ERROR] error opening file

src/zjs_modules.c:154 native_require_handler():
[ERROR] could not read module modules/easterbunny
calling require with an invalid module name: expected error object
FAIL - calling require with an invalid module name
TOTAL: 10 of 11 passed
```

Branch | Target Device | Build ID | Date | Result
---|---|---|---|---
master | Linux | 21915c0 | Feb 7, 2017 | FAIL by 677 
master | Linux | 3f8eb85 | Feb 13, 2017 | FAIL

grgustaf:
This should be fixed by #774.

qiaojingx:
Verified with commit b29dbbc. This issue is not reproducible on Linux. Closed.

PR #774, opened 2017-02-28
[assert] Demote AssertionError from standard to custom error

grgustaf:
This means instanceof AssertionError won't work but you can still test
the name; since this comes from one console function it doesn't seem
like it should be considered a standard error.

Fixes issue #735.


jimmy-huang:
+1

PR #775, opened 2017-03-01
[sensor] Use zjs_custom_error to create error events

jimmy-huang:


jimmy-huang:
submitting another patch that will include this, close for now

PR #778, opened 2017-03-02
[ashell] Removing extra window info and various cleanup

brianjjones:


poussa:
+1

src/ashell/comms-shell.c
```diff
@@ -697,4 +697,4 @@ void shell_unit_test()
     }
     printf("All tests were successful \n");
 }
-#endif
+#endif //CONFIG_SHELL_UNIT_TESTS
```
grgustaf: To follow convention, two spaces before //, one space after.

grgustaf:
Two more things I noticed running it:
- We should probably still print a short message indicating that the real shell is at ACM0 or so
- We are currently running the HelloWorld.js (or whatever is passed in with JS=) even though we're in ashell mode, that should be disabled I think.

brianjjones:
@grgustaf True no real reason (I think) to be running hello world, its just the default behaviour when JS isn't passed.  I'll look into adding the removal of helloworld in a future patch

brianjjones:
@grgustaf Warning added

grgustaf:
+1

PR #779, opened 2017-03-02
[jslinux] Improve output at termination

grgustaf:
The message looked like an error had occurred, when this was normal
operation. Mitigating that somewhat, but maybe we should do away with
the message entirely.


jprestwo:
+1

PR #781, opened 2017-03-02
Flush pending callbacks

grgustaf:


grgustaf:
Note, reordering zjs_callback_t took it from 48 bytes to 40 bytes (on Linux at least).

src/zjs_callbacks.c
```diff
@@ -323,17 +343,33 @@ void zjs_remove_callback(zjs_callback_id id)
...
         }
-        zjs_free(cb_map[id]);
-        cb_map[id] = NULL;
+        SET_CB_REMOVED(cb_map[id]->flags);
+        if (!skip_flush) {
+            int ret = zjs_port_ring_buf_put(&ring_buffer, (uint16_t)id,
+                                            CB_FLUSH_ONE, NULL, 0);
+            if (ret) {
+                // couldn't add flush command, so just free now
+                zjs_flush_callback(id);
```
jprestwo: Maybe print an error/debug message here because this likely means the ring buffer is full.

grgustaf:
OK, I added a debug print and also renamed flush function to free because "flush" really means ensuring the pending callbacks get flushed out of the queue safely, not the final freeing of the callback afterward.

jprestwo:
LGTM assuming travis is happy.

jimmy-huang:
+1

PR #788, opened 2017-03-02
[util] Fixed bug where util functions don't check for valid number value

jimmy-huang:
If the object doesn't not have the property set it will still
return true since there's no error, but the numeric values
are invalid

### DIFF
--- a/src/zjs_util.c
+++ b/src/zjs_util.c
@@ -130,8 +130,7 @@ void zjs_obj_add_readonly_number(jerry_value_t obj, double num,
 
 bool zjs_obj_get_boolean(jerry_value_t obj, const char *name, bool *flag)
 {
-    // requires: obj is an existing JS object, value name should exist as
-    //             boolean
+    // requires: obj is an existing JS object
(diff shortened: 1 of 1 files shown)


src/zjs_util.c
```diff
@@ -175,7 +175,7 @@ bool zjs_obj_get_double(jerry_value_t obj, const char *name, double *num)
     jerry_value_t value = zjs_get_property(obj, name);
     bool rval = false;
 
-    if (!jerry_value_has_error_flag(value)) {
+    if (!jerry_value_has_error_flag(value) && jerry_value_is_number(value)) {
```
grgustaf: This does seem like a good idea, since usually people probably don't know for sure they will have a numeric property. But it changes the contract... so remove the part in the "requires:" line above that says "value name should exist as number".

The idea of "requires" is that you specify what is valid usage of the function and therefore don't have to check for invalid conditions within the function. Same change both places below.

jimmy-huang: @grgustaf updated patch

grgustaf:
+1

PR #789, opened 2017-03-03
[sensor] Added default state "unconnected"

jimmy-huang:
The default state is no longer "idle" and should be "unconnected",
also it should be a readonly property, and readings should also
be readonly as well


grgustaf:
+1

PR #791, opened 2017-03-03
[sensor] Fix sensor to pass unit test

jimmy-huang:
Sensor tests are broken because because the callstack is corrupted
when sending IPM messages, and this patch rewrite the start sensor
logic to have a callback that's called from the mainloop instead
for a smaller call stack.  Also fixed an issue where stop fails
on AmbientLightSensor

### DIFF
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -30,7 +30,9 @@ enum sensor_state {
 };
 
 typedef struct sensor_handle {
-    zjs_callback_id id;
+    zjs_callback_id onchange_cb_id;
+    zjs_callback_id onstart_cb_id;
+    zjs_callback_id onstop_cb_id;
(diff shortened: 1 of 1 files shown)


src/zjs_sensor.c
```diff
@@ -450,7 +496,7 @@ static jerry_value_t zjs_sensor_create(const jerry_value_t function_obj,
                                        enum sensor_channel channel)
 {
     double frequency = DEFAULT_SAMPLING_FREQUENCY;
-    uint32_t pin;
+    uint32_t pin = -1;
```
grgustaf: Let's not assign negative numbers to unsigned ints, even if it works. :)

I guess use 0xffffffff?

jimmy-huang:
@grgustaf updated patch

grgustaf:
+1

PR #793, opened 2017-03-06
[build] Set main stack size from prj.conf instead of prj.mdef

grgustaf:


jimmy-huang:
+1

Makefile
```diff
@@ -173,7 +173,6 @@ analyze: $(JS)
 ifeq ($(DEV), ashell)
 	@cat fragments/prj.mdef.dev >> prj.mdef
```
jimmy-huang: shouldn't prj.mdef.dev and prj.mdef.heap  be removed as well? or move to prj.conf equivalent?

poussa: yes, we don't need to .mdef file at all. The heap pool can be defined in .conf file as well (CONFIG_HEAP_MEM_POOL_SIZE)

grgustaf: Patches accepted. :)

I wanted to commit the part that I had figured out so far to make some progress...

grgustaf:
Added a patch to remove heap size setting and a couple more of the files, but there is more to do to completely eliminate mdef. I think we can get these in though.

jimmy-huang:
+1

PR #800, opened 2017-03-08
[ocf] register resource correctly

jprestwo:


src/zjs_ocf_client.c
```diff
@@ -1093,4 +1093,23 @@ jerry_value_t zjs_ocf_client_init()
     return ocf_client;
 }
 
+void zjs_ocf_client_cleanup()
+{
+    if (resources) {
```
grgustaf: Hmm, using such a generic variable name isn't a great idea, for a global. But especially when it's not defined as static - can it be?

src/zjs_ocf_common.c
```diff
@@ -470,4 +479,11 @@ jerry_value_t zjs_ocf_init()
     return ocf_object;
 }
 
+void zjs_ocf_cleanup()
+{
+    zjs_ocf_server_cleanup();
+    zjs_ocf_client_cleanup();
```
grgustaf: Does this work even if one is compiled out?

src/zjs_ocf_server.c
```diff
@@ -461,47 +476,86 @@ static jerry_value_t ocf_register(const jerry_value_t function_val,
...
     }
+    new->resource = resource;
+    new->next = res_list;
+    res_list = new;
+
+    resource->flags = flags;
 
-    resource->res = oc_new_resource(resource_path, num_types, 0);
+    resource->num_types = jerry_get_array_length(res_type_array);
+    resource->resource_types = zjs_malloc(sizeof(char*) * resource->num_types);
```
grgustaf: A trick that might be worth employing here (not needed in this patch) is to move resource_types and _ifaces to a data[] portion at the end of the struct. Then allocate a buffer for the struct that is sizeof(struct) + sizeof(char *) * (num_types + num_ifaces) or whatever, write those two arrays to offsets into that data, and then access them that way.

This would allow us to use one malloc instead of three which would save overhead and wasted padding I think.

src/zjs_ocf_server.c
```diff
@@ -385,14 +397,15 @@ static jerry_value_t ocf_notify(const jerry_value_t function_val,
...
-} resource_list_t;
-
-static resource_list_t *r_list = NULL;
-static bool has_registered = false;
-*/
+ * TODO: This is a workaround for getting the resource UUID. There is no API
+ *       available to do this currently so we must get it with this external
+ *       structure.
+ */
+extern struct oc_device_info_t
```
grgustaf: The _t should be reserved for a typedef.

src/zjs_ocf_server.c
```diff
@@ -421,14 +436,14 @@ static jerry_value_t ocf_register(const jerry_value_t function_val,
...
+        jerry_release_value(res_type_array);
         ERR_PRINT("resourceTypes array not found\n");
         REJECT(promise, "TypeMismatchError", "resourceTypes array not found", h);
         return promise;
     }
-    uint32_t num_types = jerry_get_array_length(res_type_array);
 
     // Optional
-    uint32_t flags = 0;
+    uint8_t flags;
```
grgustaf: This is not good, you're ORing with an uninitialized variable below. O_o

PR #806, opened 2017-03-09
[sensor] Added work-around for an issue on Arduino 101

jimmy-huang:
The underlying Zephyr driver will return an error when
trying to unregister the trigger callback on the Arduino
101, add a second retry when this happens.


jimmy-huang:
closing this as it's replaced by #807

PR #807, opened 2017-03-10
Repeat sensor stop

grgustaf:


grgustaf:
We picked the number 50 because after creating an inifinite loop and printing out the actual number of retries it took, on my box it was usually 17-24. For Jimmy, just retrying once had seemed to be enough. So there appears to be some variance between devices and maybe phones, so we went a bit higher to be safe. But we don't want it to block everything if it never works. Luckily this is only during the 'stop' command which shouldn't be frequent in normal usage.

But @jimmy-huang will look into reproducing this in a way consumable to the Zephyr folks so they can root-cause it.

grgustaf:
This obsoletes #806, I believe.

jimmy-huang:
tested and works, +1

PR #812, opened 2017-03-10
[ocf] Add platform info to client callback

jprestwo:

### DIFF
--- a/src/zjs_ocf_client.c
+++ b/src/zjs_ocf_client.c
@@ -863,14 +863,20 @@ static void ocf_get_platform_info_handler(oc_client_response_t *data)
                 break;
             case BYTE_STRING:
             case STRING:
-                /*
-                 * TODO: try and find the names of all the platform info properties
-                 *       can set a map in C array to reference these strange names
-                 */
                 if (strcmp(oc_string(rep->name), "mnmn") == 0) {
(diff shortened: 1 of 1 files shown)


src/zjs_ocf_client.c
```diff
@@ -863,14 +863,20 @@ static void ocf_get_platform_info_handler(oc_client_response_t *data)
...
             case STRING:
-                /*
-                 * TODO: try and find the names of all the platform info properties
-                 *       can set a map in C array to reference these strange names
-                 */
                 if (strcmp(oc_string(rep->name), "mnmn") == 0) {
                     zjs_obj_add_string(platform_info, oc_string(rep->value.string), "manufacturerName");
                 } else if (strcmp(oc_string(rep->name), "pi") == 0) {
                     zjs_obj_add_string(platform_info, oc_string(rep->value.string), "id");
+                } else if (strcmp(oc_string(rep->name), "mnmo") == 0) {
```
jimmy-huang: Should these be read-only properties?

jprestwo: Changing them has no effect but perhaps they should be. Do we have an easy API for that?

grgustaf: @jprestwo, s/add_string/add_readonly_string/ - can't get much easier. :)

PR #813, opened 2017-03-10
[error] Print JS function where error was thrown

jprestwo:

### DIFF
--- a/src/zjs_error.c
+++ b/src/zjs_error.c
@@ -1,5 +1,7 @@
 // Copyright (c) 2017, Intel Corporation.
 
+#include <string.h>
+
 // ZJS includes
 #include "zjs_util.h"
 
@@ -105,16 +107,15 @@ static char* construct_message(jerry_value_t this, jerry_value_t func,
     }
     jerry_release_value(keys_array);
 
-    int nlen = strlen(name);
     int mlen = strlen(message);
-    char *msg = zjs_malloc(mlen + nlen + 4);
+    char *msg = zjs_malloc(mlen + size + 4);
 
-    memcpy(msg, name, nlen);
-    msg[nlen] = '(';
-    msg[nlen + 1] = ')';
-    msg[nlen + 2] = ':';
-    memcpy(msg + nlen + 3, message, mlen);
-    msg[mlen + nlen + 3] = '\0';
+    memcpy(msg, name, size);
+    msg[size] = '(';
+    msg[size + 1] = ')';
+    msg[size + 2] = ':';
+    memcpy(msg + size + 3, message, mlen);
+    msg[mlen + size + 3] = '\0';
 
     return msg;
 }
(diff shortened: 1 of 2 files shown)


src/zjs_error.c
```diff
@@ -80,6 +80,63 @@ void zjs_error_cleanup()
     }
 }
 
+static char* construct_message(jerry_value_t this,
+                                      jerry_value_t func,
+                                      const char *message)
```
grgustaf: Indent looks wrong, move func up to first line, tab message to the right place.

src/zjs_error.c
```diff
@@ -80,6 +80,63 @@ void zjs_error_cleanup()
...
+{
+    jerry_size_t size = 64;
+    char name[size];
+    jerry_value_t keys_array = jerry_get_object_keys(this);
```
grgustaf: Should probably check for an error.

src/zjs_error.c
```diff
@@ -80,6 +80,63 @@ void zjs_error_cleanup()
...
+    uint32_t arr_length = jerry_get_array_length(keys_array);
+    int i;
+    for (i = 0; i < arr_length; ++i) {
+        jerry_value_t val = jerry_get_property_by_index(keys_array, i);
+        zjs_copy_jstring(val, name, &size);
+        jerry_value_t func_val = zjs_get_property(this, name);
```
grgustaf: I think you can probably call jerry_get_property_by_index(this, i) to get func_val, and then you don't need to copy the string every time. If it matches, then copy the string and save the name.

src/zjs_error.c
```diff
@@ -80,6 +80,63 @@ void zjs_error_cleanup()
...
+        if (func_val == func) {
+            break;
+        }
+        jerry_release_value(val);
+        jerry_release_value(func_val);
+    }
+    jerry_release_value(keys_array);
+
+    int nlen = strlen(name);
```
grgustaf: You already have size returned above so you should be able to skip this strlen.

src/zjs_error.c
```diff
@@ -80,6 +80,63 @@ void zjs_error_cleanup()
...
+    char *msg = zjs_malloc(strlen(message) + nlen + 4);
+
+    memcpy(msg, name, nlen);
+    msg[nlen] = '(';
+    msg[nlen + 1] = ')';
+    msg[nlen + 2] = ':';
+    memcpy(msg + nlen + 3, message, strlen(message));
+    msg[strlen(message) + nlen + 3] = '\0';
```
grgustaf: Save the strlen(message) above rather than calling it 3x.

grgustaf:
Otherwise, looks good. I guess we're not using it yet, but before we do we should make this macro do this only on a certain debug level. A lower level would leave the function names out and call the original function.

src/zjs_error.c
jprestwo: I tried this and the function named ended up beings "X". Not sure why but it didn't seem to work.

src/zjs_error.c
grgustaf: You can still make this change I think.

grgustaf:
+1

PR #815, opened 2017-03-10
[callbacks] Back off ring buf reporting over time

grgustaf:
This makes sure we don't cascade into bigger problems with our
ringbuf error reporting.


jimmy-huang:
lgtm, this really helps with debugging sensor thoughput

PR #816, opened 2017-03-10
[demo] Increase sensor frequency to 100

jimmy-huang:
It looks like without any prints, we can achieve a maximum of 100hz
for sending sensor data over BLE, seeing only a few dropped ring buffer
errors.


grgustaf:
+1, works for me too

PR #817, opened 2017-03-11
[util] Add macros/functions to standardize argument validation

grgustaf:
The new ZJS_VALIDATE_ARGS macro and related macros are to be used at
the top of any JerryScript C binding to verify the quantity and types
of arguments, and return standard exceptions if there is a problem.

The comments above the macros in zjs_util.h explain the details of how
they are to be used.

### DIFF
--- a/src/zjs_console.c
+++ b/src/zjs_console.c
@@ -182,9 +182,8 @@ static jerry_value_t console_time(const jerry_value_t function_obj,
                                   const jerry_value_t argv[],
                                   const jerry_length_t argc)
 {
-    if (argc < 1 || !jerry_value_is_string(argv[0])) {
-        return TYPE_ERROR("invalid argument");
-    }
+    // args: label
+    ZJS_VALIDATE_ARGS(Z_STRING);
 
     uint32_t start = zjs_port_timer_get_uptime();
 
@@ -199,9 +198,8 @@ static jerry_value_t console_time_end(const jerry_value_t function_obj,
                                       const jerry_value_t argv[],
                                       const jerry_length_t argc)
 {
-    if (argc < 1 || !jerry_value_is_string(argv[0])) {
-        return TYPE_ERROR("invalid argument");
-    }
+    // args: label
+    ZJS_VALIDATE_ARGS(Z_STRING);
 
     jerry_value_t num = jerry_get_property(gbl_time_obj, argv[0]);
     jerry_delete_property(gbl_time_obj, argv[0]);
@@ -231,10 +229,10 @@ static jerry_value_t console_assert(const jerry_value_t function_obj,
                                     const jerry_value_t argv[],
                                     const jerry_length_t argc)
 {
+    // args: validity[, output]
+    ZJS_VALIDATE_ARGS(Z_BOOL, Z_OPTIONAL Z_ANY);
+
     char message[MAX_STR_LENGTH];
-    if (!jerry_value_is_boolean(argv[0])) {
-        return TYPE_ERROR("invalid parameter");
-    }
     bool b = jerry_get_boolean_value(argv[0]);
     if (!b) {
         if (argc > 1) {
--- a/src/zjs_dgram.c
+++ b/src/zjs_dgram.c
@@ -1,4 +1,5 @@
 // Copyright (c) 2017, Linaro Limited.
+
 #ifdef BUILD_MODULE_DGRAM
 
 // Zephyr includes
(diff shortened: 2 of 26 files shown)


grgustaf:
I was expecting this patch to save code size, because it sometimes replaces many lines of error checking in API functions with one line. However, that line expands to a 6 line macro, which is sometimes more than what was there originally.

I compiled ashell without my patch, and the image was 242,828 bytes. With my patch it was 243,148 bytes (+320). So that was disappointing. :) But in dozens of call sites this *improved* the argument checking, not just replacing it. Many of those spots were actual bugs like using argv[0] when there might be no arguments. So I still think this is a big win.

But another thing I added was a line to the macro that returns an "insufficient arguments" error if there are not enough arguments, as opposed to "invalid argument" when one is the wrong type. If I remove those lines, and rebuild ashell I get 242,380 bytes (-448). So that's more like it. So I guess I should ask whether it's worth 768 bytes to get that extra information about bad arguments passed to our functions, and I think the answer is no. So I'll remove that now, and both conditions will return "invalid arguments".

grgustaf:
Since I do love my statistics, this patch only saved a net of 18 lines of code from our tree... but did save 3137 characters (0.68% of our code in src/). :)

poussa:
LGTM despite the code size increase.

src/zjs_fs.c
```diff
@@ -702,42 +648,19 @@ static jerry_value_t zjs_mkdir(const jerry_value_t function_obj,
...
-        } else {
-            size = 4;
-            char mode_str[size];
+#endif
 
-            zjs_copy_jstring(argv[1], mode_str, &size);
-            if (!size) {
-                return zjs_error("size mismatch");
-            }
+    uint32_t mode = MODE_A_PLUS;
```
jprestwo: Can remove mode, its unused. It was not actually being used for anything before but it was technically "used" so there was no compiler warning before.

grgustaf:
Later I realized I added a ton of // args: comments, so actual lines of code decreased a lot more than 18.

src/zjs_fs.c
grgustaf: Thanks.

brianjjones:
LGTM for my portion

src/zjs_console.c
```diff
@@ -231,10 +229,10 @@ static jerry_value_t console_assert(const jerry_value_t function_obj,
                                     const jerry_value_t argv[],
                                     const jerry_length_t argc)
 {
+    // args: validity[, output]
+    ZJS_VALIDATE_ARGS(Z_BOOL, Z_OPTIONAL Z_ANY);
+
     char message[MAX_STR_LENGTH];
-    if (!jerry_value_is_boolean(argv[0])) {
```
grgustaf: Here's one example of where this patch fixed a minor bug - before we weren't checking to make sure there *was* an argv[0] so if a JS user passed no args we would have made an illegal access.

src/zjs_fs.c
```diff
@@ -488,59 +467,23 @@ static jerry_value_t zjs_write(const jerry_value_t function_obj,
                                const jerry_length_t argc,
                                uint8_t async)
 {
+    // args: file descriptor, buffer, offset, length[, position]
+    ZJS_VALIDATE_ARGS_OPTCOUNT(optcount, Z_OBJECT, Z_OBJECT, Z_NUMBER,
+                               Z_NUMBER, Z_OPTIONAL Z_NUMBER);
```
jprestwo: The Node prototype for this is: `write(fd, buffer[, offset[, length[, position]]])`. Can we put Z_OPTIONAL in front of all the needed parameters? I tried it locally and it "worked" i.e. test-fs.js passed. Whether or not it actually validated I am not sure but if the validation can't do multiple optional arguments then we need to change the docs/test for FS because it explicitly tests different ways of calling fs.writeSync().

jimmy-huang:
+1

grgustaf:
I've tested about two thirds of the tests in tests/ with this patch and they still work. I think we're in pretty good shape. I'm looking at maybe adding C unit tests though.

jprestwo:
LGTM +1

src/zjs_dgram.c
```diff
@@ -295,10 +293,13 @@ static jerry_value_t zjs_dgram_sock_bind(const jerry_value_t function_obj,
                                          const jerry_value_t argv[],
                                          const jerry_length_t argc)
 {
-    int ret;
+    // NOTE: really argv[0] and argv[1] are supposed to be optional, but
+    //   leaving them as they were here for now
```
pfalcon: Current googleable Node.js docs show them as optional, but Node.js I have on my Ubuntu 14.04 requires them and doesn't allow to omit. And then googleable docs don't say anything from which version of Node.js they became optional. In such situation, I consider a real-world executable to be definitive, whereas randomly googled-up docs can be for too old a version (was optional, but no longer), or too new (will be become optional in a version which will get world-wide deployment in a year). Generally, I find Node.js docs lacking many important details, that's why I asked you guys if you have a habit of poking into Node.js source when unclear cases arise (it's in my TODO to acquire such a habit ;-) ).

PR #822, opened 2017-03-13
Make sensor demo work with negative numbers.

kenchris:
We now encode the JS float value as an unsigned 32 bit integer
by multiplying with 2^(32-14) due to 14 bit precision and shifting
one to retain the sign bit.


src/zjs_buffer.c
```diff
@@ -89,7 +92,9 @@ static jerry_value_t zjs_buffer_write_bytes(const jerry_value_t this,
         return zjs_error("zjs_buffer_write_bytes: invalid argument");
     }
 
-    uint32_t value = jerry_get_number_value(argv[0]);
+    // Cast a jerry_value_t to an uint32_t.
+    // info: jerry-core/ecma/base/ecma-helpers-value.c
+    uint32_t value = (uint32_t) (((int32_t) argv[0]) >> ECMA_DIRECT_SHIFT);
```
grgustaf: So was this hurting signed values for you?

Can you try this:
uint32_t value = (uint32_t)(int32_t)jerry_get_number_value(argv[0]);

Then we don't need to expose internal details...

grgustaf: Also @kenchris could you let me know some sample values that were breaking for you, so however we fix it right now we can add tests for that in the future?

kenchris: I will try. I added a test, it was breaking before

grgustaf: Sorry, should have kept reviewing instead of jumping out to write my own sample code. :)

kenchris:
OK, worked, so updated patch. Could you explain to me why the first cast to int32_t makes a difference?

jimmy-huang:
@kenchris with this patch, does the chart int the UI need to be adjusted for negative values as well?  because now the accel x y z axis have a range of -1 to 50  and they gyro has a -1 to 20, which will need to be adjusted, correct?

kenchris:
the test case is https://sensor-compass.appspot.com

Open the menu and select Use Zephyr Device, then choose the Complementary filter (or accelerometer or gyroscope) and move the Arduino 101 around

kenchris:
@jimmy-huang yes, it needs to be adjusted. Both how I get the values (simple fix) and probably the ranges.

I guess you want -50 to 50 and -20 to 20?

I can quickly fix that after this lands

jimmy-huang:
yeah, think it will work

jimmy-huang:
@kenchris i tried your site, I can see the Gyroscope moving, how do you see Accelerometer data?

kenchris:
Use the menu (swipe in from left or click on hamburger menu). Then select accelerometer. Or add the following to the URL: ?filter=a

grgustaf:
@kenchris I wasn't sure if it would solve it, just going off a gut math intuition. :)

If I can find a good example it will probably be more clear. But basically:

Casting from double to uint32_t is casting into the range 0 to 2^32-1.
Casting from double to in32_t is casting into the range -2^31 to 2^31-1, I think.

I guess I just suspect that the rules for how those casts are performed are different. So for example, if you have a large negative number, and you cast it to a uint32_t you might make it 0. But if you cast it to a int32_t you might make it the lower possible negative number. Then when you cast that to uint32_t you'd get a different result.

Here's a tiny sample app that demonstrates it:

#include <stdio.h>

void main() {
    unsigned long ul = -1L << 40;
    printf("big int is %lu\n", ul);
    double d = ul;
    printf("double is %g\n", d);
    printf("%u, %d, %u\n", (unsigned int)d, (int)d, (unsigned int)(int)d);
    return;
}

Now what I *don't* understand is if you change -1L to 1L up there, the double still shows up as a negative number. This stuff is really messy. :)

jimmy-huang:
@kenchris yeah, i already tried that, but i only saw 1 number displayed above the Accelerometer text, but i never seen a negative value in there

kenchris:
@jimmy-huang the number above is just the heading... It you can move the compass in all directions and it works, it is using negative numbers. Like lay device on table - there are 4 sides. Put a finger under a side and lift it...That only worked for 2 of the 4 sides before

kenchris:
Thanks for the explanation Geoffrey :) I wish I knew these things by heart

tests/test-buffer.js
```diff
@@ -84,4 +84,13 @@ assert.throws(function () {
...
 }, "Error thrown when 'hex' is not given to toString()");
 
+var buff = new Buffer(4);
+var writeValue = -0.232;
+// write negative float value (14 bit precision) as uint32 by disabling noAssert.
+buff.writeUInt32BE((writeValue * Math.pow(2, 32-14)) << 1, 0, true);
+var readValue = (buff.readUInt32BE(0) >> 1) / Math.pow(2, 32-14);
+assert(writeValue.toPrecision(3) === readValue.toPrecision(3),
+       "The value of readUInt32BE() with precision 3 expected: "
+       + writeValue + " got: " + readValue.toPrecision(3));
```
grgustaf: Hmm this doesn't work on A101 because we drop support for Math libraries and such to fit. But I guess it works in jslinux. We can probably leave it in and work on this in the future.

kenchris: Or we compare substrings, though it is a bit dirty. Ie find index of . and do splice 3 chars after

jimmy-huang:
@kenchris ok, i guess i'll have to play with it more, i just can't see the compass moving at all, it's only rotating/tilting no matter how i moved the Arduino 101.  But anway, I am okay with merging this if it works for you.

grgustaf:
+1 from code review perspective, haven't really tested yet.

kenchris:
Yes it will rotate and tilt, like this https://youtu.be/K0efnqWDrkY

Is that what you see?

More info: https://w3c.github.io/motion-sensors/#relative-orientation

jimmy-huang:
@kenchris yeah that's what i see, i was assuming when you move it, the compass will move accelerated to the side or to the top, maybe i am overthinking it,

kenchris:
Ah :) No this is a basic complementary filter as that has quite some used already, instead of looking fancy ;) We could of course experiment with such things later

jimmy-huang:
@kenchris look good though, @poussa want us to hold off merging it until the demo, so if you can submit a patch too to my repo to the charts, I'll test it out for you.

kenchris:
OK, I will do that then

kenchris:
@jimmy-huang I pushed a PR to your repo, please test.

kenchris:
@grgustaf I still then wonder why they do the (ecma) shift of 4 and how that affects the casts. If that is the proper way to do it, it would be nice if JerryScript had a method for it

kenchris:
screenshot_20170313-203217

Jimmy this is how it looks with this patch applied and the PR on your repo

kenchris:
Could we land this now? Then I can push the changes to the demos sites before I pack my laptop away (traveling tomorrow)

jimmy-huang:
+1

samples/SensorBLEDemo.js
```diff
@@ -31,15 +31,18 @@ SensorCharacteristic.onUnsubscribe = function() {
...
 
     var data = new Buffer(13);
     data.writeUInt8(isAccel, 0);
-    data.writeUInt32BE(xval, 1);
-    data.writeUInt32BE(yval, 5);
-    data.writeUInt32BE(zval, 9);
+
+    data.writeUInt32BE(xval, 1, true);
+    data.writeUInt32BE(yval, 5, true);
+    data.writeUInt32BE(zval, 9, true);
```
grgustaf: @kenchris, Hmm, we don't support the noAssert argument. Did you add these because you needed them when you ran this under Node?

src/zjs_buffer.c
```diff
@@ -89,7 +90,7 @@ static jerry_value_t zjs_buffer_write_bytes(const jerry_value_t this,
         return zjs_error("zjs_buffer_write_bytes: invalid argument");
     }
 
-    uint32_t value = jerry_get_number_value(argv[0]);
+    uint32_t value = (uint32_t)(int32_t)jerry_get_number_value(argv[0]);
```
grgustaf: @kenchris, this broke our big endian functions when testing with unsigned int over 2bil like 0xdeadbeef as we do in tests/test-buffer-rw.js. Then I noticed in the Node docs that behavior is undefined if you pass them anything other than an unsigned number (i.e. a negative). So the right solution is to fix your demo probably. But I think I do have a way to make it continue to work. I'll cc you on the PR shortly if you want to try it.

PR #827, opened 2017-03-14
[sensor] Fix broken sensor in ashell

jimmy-huang:
The problem is that there's a conflict of the SPI bus in ashell
and sensor, so we need to explicitly disable SPI_0 and SPI_1 for
ARC when running in ashell.  We also need to increase the callstack
size for ashell, also have to decrease frequency to 20hz since Ashell is
so slow that even 100hz will be flooding the ringbuffer.


jimmy-huang:
@poussa please test if this works for you

scripts/analyze.sh
```diff
@@ -304,7 +304,23 @@ if [ $? -eq 0 ] || check_config_file ZJS_SENSOR; then
     >&2 echo Using module: Sensor
     MODULES+=" -DBUILD_MODULE_SENSOR"
     echo "export ZJS_SENSOR=y" >> zjs.conf.tmp
-    if [ $BOARD = "arduino_101" ]; then
+    if [ $DEV = "ashell" ]; then
```
brianjjones: Is there a reason this configuration wouldn't work for non-ashell builds on arduino 101? i.e. could we just combine this into if [ $DEV = "ashell" ] || [ $BOARD = "arduino_101" ]

grgustaf: This should go into prj.conf.dev or something, shouldn't it?

@brianjjones, Most/all of these are set by analyze.sh for the non-ashell case, when appropriate.

jimmy-huang: @brianjjones, I believe these configurations can be used for non-ashell case, but we are not 100% sure why we have to set these for ARC, that's why when I submitted the patch, I didn't want to break the existing configuration.  Maybe when you heard anything back from the Zephyr on the bug you filed, we could adapt these changes for non-ashell.

jimmy-huang: @grgustaf you still want to easily toggle these through zjs.conf, right? so if you put these in the prj.conf.dev, it will always get included.

grgustaf: I didn't realize this code was in analyze.sh. :) Yeah you're right.

samples/SensorBLEDemo.js
```diff
@@ -90,11 +90,11 @@ ble.on('disconnect', function(clientAddress) {
 });
 
 var accel = new Accelerometer({
-    frequency: 100
+    frequency: 20
 });
 
 var gyro = new Gyroscope({
-    frequency: 100
+    frequency: 20
```
grgustaf: Maybe define this as an updateFrequency variable at the top of the file? And maybe add a note there that it seems to work with 100 for production mode, but we need 20 for ashell?

brianjjones:
lgtm +1

Issue #794, opened 2017-03-07
[REG] Not used parameter in AIO changed to required parameter


qiaojingx:
### Description

 Not used parameter in AIO changed to required parameter

### Test Code

[test-aio.js](https://github.com/01org/zephyr.js/blob/master/tests/test-aio.js)
 
### Steps to Reproduction

 1. `$ make JS=tests/test-aio.js`
 2. `$ make dfu`
 3. Wire IO4 to A0
 4. Reset device.
### Actual Result

![test-aio](https://cloud.githubusercontent.com/assets/19813546/23643241/6950e7b4-033b-11e7-942a-98b6295555e6.png)
 
### Expected Result

 All cases of test-aio.js will pass without error.

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 76c4663| Arduino 101 | Mar 1, 2017 | Pass
master | b29dbbc| Arduino 101 | Mar 7, 2017 | Fail
 
### Additional Information

Modify `var pinA = aio.open({pin: pins.A0 });` to `var pinA = aio.open({ device: 0, pin: pins.A0 });` in test-aio.js.
All cases of test-aio.js will pass.
Is this by desgined? If so, please update [aio.md](https://github.com/01org/zephyr.js/blob/master/docs/aio.md)

qiaojingx:
Verified with commit 8a2964e on Arduino 101. This issue has been fixed.
Closed

PR #830, opened 2017-03-14
[aio] Remove unused parameters in open

jimmy-huang:
The device, name and raw properties are all unused, so
removing these for fixing #794.


grgustaf:
+1

Issue #826, opened 2017-03-13
prj.conf construction is confusing and error-prone


pfalcon:
I'm investigating https://github.com/01org/zephyr.js/issues/819 (which is mis-named by now).

My mental model of how prj.conf is generated is: prj.conf.base is merged with a board-specific fragment, then optionally with feature-specific one like ashell. That's why https://github.com/01org/zephyr.js/commit/b15bf43c0f1041436d73719eea0ba6b4390ac4cb#diff-51461e961b292b547a2fb47817be9f26 is expected to cover all boards. Alas, it's not how it works. Turns out, qemu_x86 doesn't use prj.conf.base, ashell config is actually arduino_101 specific (so I now can be sure ashell won't work on anything else), etc.

Can this be cleaned please? Just making prj.conf.base be used for everything would be a good start.

brianjjones:
This should be addressed by https://github.com/01org/zephyr.js/pull/835 
Now .base is used by all and the other prj.conf files are for specific cases only.

PR #835, opened 2017-03-15
[build] Clean up prj.conf and other config files in build

brianjjones:
This is also a fix for #826


Makefile
```diff
@@ -171,7 +171,7 @@ analyze: $(JS)
 		echo "ccflags-y += -DZJS_SNAPSHOT_BUILD" >> src/Makefile; \
 	fi
 ifeq ($(DEV), ashell)
-	@cat fragments/prj.mdef.dev >> prj.mdef
+	@cat fragments/prj.mdef.ashell >> prj.mdef
```
jimmy-huang: couldn't this be moved into the bottom section where you append the prj.conf? so you don't need a duplicate ifeq

fragments/prj.conf.ashell
```diff
@@ -1,14 +1,10 @@
-# HEAP
-CONFIG_HEAP_MEM_POOL_SIZE=4096
-
 # SHELL
 CONFIG_EARLY_CONSOLE=y
```
jimmy-huang: if you are already get these configs from prj.conf.base, then there's a lot of duplicate here like STDOUT_CONSOLE, FLOAT, NEWLIB

grgustaf:
+1

Makefile
grgustaf: They're in different sections, I don't want to risk that right now. The mdef stuff is in one place, the prj.conf is in one place.

PR #837, opened 2017-03-15
[general] Make pointer declarations consistent across the project

grgustaf:

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -36,10 +36,10 @@
 
 static struct k_sem arc_sem;
 static struct zjs_ipm_message msg_queue[QUEUE_SIZE];
-static struct zjs_ipm_message* end_of_queue_ptr = msg_queue + QUEUE_SIZE;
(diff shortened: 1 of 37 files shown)


jimmy-huang:
lgtm

pfalcon:
Did you guys consider to start a codestyle doc (wiki page)? It doesn't have to be comprehensive for starters, or many-word'y, just: "do it like this [example1], not like this: [example2], [example3]".

In this case, I definitely relate that in a declaration, the pointer should be like:

~~~
void *p;
~~~

However, when you need a raw type (without variable identifier), my preferred form is to avoid verbose spacing:

~~~
(void*)p
~~~

When I wrote zjs_dgram.c, I checked codebase that there's existing usage like the above, so used it myself. If I didn't do `git log -p` while looking for regression, I'd probably still think that using `(void*)` is right (and keep fingers crossed I'll remember it from this single incident).

PR #838, opened 2017-03-15
[promise] Implement JS promises

jprestwo:


src/zjs_promise.c
```diff
@@ -5,155 +5,596 @@
...
 
-static jerry_value_t promise_catch(const jerry_value_t function_obj,
-                                   const jerry_value_t this,
-                                   const jerry_value_t argv[],
-                                   const jerry_length_t argc)
+// 25.4.1.3
+static jerry_value_t create_resolving_functions(jerry_value_t promise)
 {
-    return promise_resolve(function_obj, this, argv, argc, 0);
+    ZVAL already_resolved = jerry_create_boolean(false);
```
jiangzidong: i think the already_resolved should be an boolean object instead of a boolean value.
both resolve.[[AlreadyResolved]] and reject.[[AlreadyResolved]] will ref to same object

src/zjs_promise.c
```diff
@@ -5,155 +5,596 @@
...
 
-    new->user_handle = handle;
-    new->post = post;
+    jerry_value_t argv[] = {res, rej};
 
-    new->then_id = zjs_add_callback_once(ZJS_UNDEFINED,
-                                         obj,
-                                         new,
-                                         post_promise);
+    ZVAL ret_val = jerry_call_function(executor, ZJS_UNDEFINED, argv, 2);
```
jiangzidong: should we free the ret_val later?

jiangzidong: ignore this. I found ZVAL will cleanup automatically

grgustaf:
#1058 now implements promises using support in JerryScript, so this one will not be necessary.

PR #842, opened 2017-03-17
[dgram] Fix bug in passing argument count

grgustaf:
Avoiding magic numbers is good, but here the magic is explained in the
line directly above, nicer than sizeof(args) / sizeof(jerry_value_t).

### DIFF
--- a/src/zjs_dgram.c
+++ b/src/zjs_dgram.c
@@ -134,7 +134,7 @@ static void udp_received(struct net_context *context,
     net_nbuf_unref(net_buf);
 
     jerry_value_t args[2] = {buf_js, rinfo};
-    zjs_signal_callback(handle->message_cb_id, &args, sizeof(args));
+    zjs_signal_callback(handle->message_cb_id, args, 2);
     jerry_release_value(buf_js);
     jerry_release_value(rinfo);
 }


jimmy-huang:
+1

pfalcon:
Bingo! Here's the incorrect change, and it's a standalone, single-purpose patch which dubs to fix bug, and actually introduces it :-I. https://github.com/01org/zephyr.js/blob/master/src/zjs_callbacks.h#L171

pfalcon:
I looked how it's done in other places and see things like:

~~~
        jerry_value_t args[3];
        args[0] = jerry_create_number(0);
        args[1] = jerry_create_number(written);
        args[2] = argv[1];
...
        zjs_signal_callback(id, args, sizeof(jerry_value_t) * 3);
~~~

This is rather error-prone. If it needs an array and its size in bytes, it should be pass that (i.e. array and its sizeof), not magic expressions like `sizeof(jerry_value_t) * 3`. YMMV

pfalcon:
Suggestion: add an assertion to zjs_signal_callback() that argument size is in full words to catch errors like this early.

grgustaf:
zjs_signal_callback doesn't have to take full words, or we *would* have used argc instead of size. :)

grgustaf:
You're right that sizeof(args) is better in that case; my PR #903 makes those changes where available throughout the project.

pfalcon:
> zjs_signal_callback doesn't have to take full words, or we would have used argc instead of size. :)

Right, it doesn't have. But are there *really* cases when it doesn't take full words? If so, ok (curious, where?) Otherwise, it's a chance to catch errors like those... (Note that zjs_signal_callback rounds up size to a full word anyway).

PR #843, opened 2017-03-17
[sensor] Update sensor API for readings

jimmy-huang:
The readings no longer are passed to the event object
in the onchange() callback, but instead stored
as a internal property in the object itself in the
latest W3C spec.

### DIFF
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -407,12 +407,12 @@ static void zjs_sensor_onstop_c_callback(void *h, void *argv)
     jerry_value_t null_val = jerry_create_null();
     switch(handle->channel) {
     case SENSOR_CHAN_ACCEL_XYZ:
-    case SENSOR_CHAN_GYRO_XYZ: ;
+    case SENSOR_CHAN_GYRO_XYZ:
         zjs_set_readonly_property(obj, "x", null_val);
         zjs_set_readonly_property(obj, "y", null_val);
         zjs_set_readonly_property(obj, "z", null_val);
         break;
-    case SENSOR_CHAN_LIGHT: ;
+    case SENSOR_CHAN_LIGHT:
         zjs_set_readonly_property(obj, "illuminance", null_val);
         break;
 
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -216,20 +216,21 @@ static void zjs_sensor_update_reading(jerry_value_t obj,
                                       void *reading)
 {
     // update object's internal properties and trigger onchange event
+    double x, y, z, d;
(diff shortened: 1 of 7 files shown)


src/zjs_sensor.c
```diff
@@ -406,7 +403,22 @@ static void zjs_sensor_onstop_c_callback(void *h, void *argv)
     send.data.sensor.channel = handle->channel;
 
     ZVAL reading_val = jerry_create_null();
-    zjs_set_property(obj, "reading", reading_val);
+    jerry_value_t null_val = jerry_create_null();
+    switch(handle->channel) {
+    case SENSOR_CHAN_ACCEL_XYZ:
+    case SENSOR_CHAN_GYRO_XYZ: ;
```
grgustaf: stray semi-colon here

src/zjs_sensor.c
```diff
@@ -515,35 +537,32 @@ static jerry_value_t zjs_sensor_create(const jerry_value_t function_obj,
...
+                                             "includeGravity");
             }
         }
 
-        if (channel == SENSOR_CHAN_LIGHT) {
-            hasPin = zjs_obj_get_uint32(options, "pin", &pin);
+        // initialize default sensor readings to null
+        jerry_value_t null_val = jerry_create_null();
+        if (channel == SENSOR_CHAN_ACCEL_XYZ ||
+            channel == SENSOR_CHAN_GYRO_XYZ) {
```
grgustaf: This should fit on one line? Not that it's hurting anything. :)

grgustaf:
Otherwise LGTM

src/zjs_sensor.c
jimmy-huang: this is actually intentional, if you don't put a semi-colon there, you get a compilation warning
error: a label can only be part of a statement and a declaration is not a statement
         double x = ((double *)reading)[0];

so i'll probably just declare it at the top

src/zjs_sensor.c
jimmy-huang: yeah, it's 82 chars so better to keep it 2 lines

jimmy-huang:
@grgustaf updated patch

src/zjs_sensor.c
grgustaf: I don't see any compile warning. What are you talking about? :D

grgustaf: Oh maybe you couldn't declare a new variable inside? But yeah you could just do that above the switch, or possibly before the cases in the switch. Or open a new {} context.

Anyway there are still two more of these in the file I think. Search for ": ;" - they can be removed.

PR #841, opened 2017-03-16
[callbacks] Remove unused return value argument to post functions

grgustaf:
Signed-off-by: Geoff Gustafson <geoff@linux.intel.com>

PR #844, opened 2017-03-17
Jerry value cleanup

grgustaf:


grgustaf:
Note: This includes the patch from PR #841 so it should be rebased when that goes in.

src/zjs_util.h
```diff
@@ -107,6 +107,30 @@ uint32_t zjs_uncompress_16_to_32(uint16_t num);
...
+ * NOTE: If you reuse the variable in your function, you will still need to
+ *   release the old value before you lose it.
+ *
+ * NOTE: If you want to return a value created with ZVAL, you need to call
+ *   jerry_acquire_value() on it first, typically in the return line.
+ *
+ * NOTE: Be careful to always initialize this value, or else an uninitialized
+ *   value will get freed. (Or will compiler catch that?)
+ */
+#define ZVAL(name) jerry_value_t name __attribute__ ((__cleanup__(zjs_free_value)))
```
martijnthe: Personally, I'd much rather be able to write:

```
ZVAL xyz = jerry_create_object();
```

So the `#define` would become:

```
#define ZVAL const jerry_value_t __attribute__ ((__cleanup__(zjs_free_value)))
```

I changed it to `const`, to prevent you from accidentally doing:

```
ZVAL xyz = jerry_create_object();
xyz = jerry_create_object(); // whoops, leaking!
```

Lastly, you probably also want a macro for "unused" `jerry_value_t` return values, so something along these lines:

```
#define CONCAT_INNER(x, y) x##y
#define CONCAT(x, y) CONCAT_INNER(x, y)
#define ZVAL_UNUSED ZVAL CONCAT(zval_unused, __COUNTER__)

ZVAL_UNUSED = jerry_call_function(...); // say I don't care about the return value
```

src/zjs_gpio.c
```diff
@@ -421,11 +414,10 @@ static jerry_value_t zjs_gpio_open(const jerry_value_t function_obj,
             zjs_reject_promise(promise_ret, &handle->open_rval, 1);
         }
 
-        jerry_release_value(pinobj);
         return promise_ret;
     }
 
-    return pinobj;
+    return jerry_acquire_value(pinobj);
```
grgustaf: Yeah, @martijnthe, that's actually what I intend them to do. Here's an example where I did it because of the error path above. It does result in an extra function call, but worth it for the safety in this kind of case I think.

src/zjs_util.h
grgustaf: Didn't notice the unused bit before, I don't think I ran into a need for that so far, but I'll keep that in mind, thanks. And the __COUNTER__ is another good thing to know about which I didn't before. :)

I don't get what the CONCAT vs. CONCAT_INNER thing is solving though. Is that needed to get __COUNTER__ to resolve instead of being appended literally (zval_unused__COUNTER__ instead of zval_unused1)?

brianjjones:
LGTM for ashell, I2C, and modules.  This will really be helpful, thanks to all who came up with it.

src/zjs_util.h
martijnthe: > I don't get what the CONCAT vs. CONCAT_INNER thing is solving though. Is that needed to get __COUNTER__ to resolve instead of being appended literally (zval_unused__COUNTER__ instead of zval_unused1)?

Right, I'm not sure if I typed it 100% correctly, there might be some syntax issues etc. (I haven't tried running it through the preprocessor.)

src/zjs_ocf_server.c
```diff
@@ -271,17 +268,21 @@ static jerry_value_t ocf_respond(const jerry_value_t function_val,
...
 
-static jerry_value_t create_request(struct server_resource *resource, oc_method_t method, struct ocf_handler *handler)
+static jerry_value_t create_request(struct server_resource *resource,
+                                    oc_method_t method,
+                                    struct ocf_handler *handler)
 {
-
     handler->resp = create_response(resource, method);
+
     jerry_value_t object = jerry_create_object();
```
jimmy-huang: ZVAL here and return acquire(object)?

grgustaf: This was intentional... if you do that in this particular case then you end up calling both jerry_release_value and jerry_acquire_object. So two function calls where none are needed.

So I think it makes sense to leave it off in this kind of case. But admittedly, there's some chance someone will come along later and return with an error and not free it.

src/main.c
```diff
@@ -230,6 +230,8 @@ int main(int argc, char *argv[])
         goto error;
     }
 
+    // NOTE: don't use ZVAL on these because we want to release
+    //   them early, here
```
grgustaf: Sure

src/zjs_ocf_server.c
jimmy-huang: yeah, it make sense, i thought it would be nice to keep it consistent

jimmy-huang:
lgtm, +1

src/zjs_dgram.c
```diff
@@ -217,22 +215,19 @@ static void udp_sent(struct net_context *context, int status, void *token,
...
         }
 
         zjs_callback_id id = zjs_add_callback_once((jerry_value_t)user_data,
-                                                   ZJS_UNDEFINED,
-                                                   NULL,
-                                                   NULL);
-        zjs_signal_callback(id, args, sizeof(args));
-        jerry_release_value(args[0]);
+                                                   ZJS_UNDEFINED, NULL, NULL);
+        zjs_signal_callback(id, &rval, 1);
```
grgustaf: Sorry for screwing up your code, @pfalcon. Here's where I went astray... in this massive patch I had a lot of room to screw something up, and I did. I had people review their areas but you weren't around that week. :/

So first I made *this* bug. Then later when I came back to the dgram.c file I think I saw this line and thought "oh, we're passing the number of jerry_values, not bytes", and thought the other one was wrong. So I screwed up a second one!

I'll get a PR in shortly to fix these.

pfalcon: np, this happens, my thinking is "how can we avoid this issue, as it may happen again for other folks", hence ideas https://github.com/01org/zephyr.js/pull/842#issuecomment-289207453 (which might not work, at least we try to think about it).

PR #845, opened 2017-03-18
Ram panacea

grgustaf:


jimmy-huang:
this is awesome, so why did we use k_malloc at the first place?  Because Zephyr wasn't intended to be used with newlib originally?

grgustaf:
Probably in our very first tests, there was no malloc because we weren't using newlib, so we just didn't realize that became available when we added it. I guess there should be some kind of code size (ROM) penalty for bringing in that good memory management, maybe I should check that.

grgustaf:
Digging into that, actually we save 1300 bytes, primarily from dropping the pool memory allocator. So this actually turns out to be a win on both RAM and ROM fronts.

jimmy-huang:
+1

PR #846, opened 2017-03-20
[build] Creating new make targets for ide and ashell

brianjjones:
This will allow faster building for ide and ashell (command line).
It also cleans the makefile up a bit


Makefile
```diff
@@ -199,8 +206,15 @@ ifeq ($(BOARD), qemu_x86)
...
 ifeq ($(DEV), ashell)
 	@cat fragments/prj.conf.ashell >> prj.conf
+	@cat fragments/prj.mdef.ashell >> prj.mdef
+ifeq ($(filter ide,$(MAKECMDGOALS)),ide)
+	@echo CONFIG_USB_CDC_ACM=n >> prj.conf
+else
+	@echo CONFIG_USB_CDC_ACM=y >> prj.conf
+endif
+else
+	@cat fragments/prj.conf.base >> prj.conf
```
jimmy-huang: ashell and ide do not use prj.conf.base any more?

Makefile
```diff
@@ -51,11 +51,21 @@ EXT_JERRY_FLAGS ?=	-DENABLE_ALL_IN_ONE=ON \
...
 					-DFEATURE_ERROR_MESSAGES=OFF \
 					-DJERRY_LIBM=OFF
-ifneq ($(DEV), ashell)
+# Settings for ashell builds
+ifneq (,$(filter $(MAKECMDGOALS),ide ashell))
+SNAPSHOT=off
+# if the user passes in SNAPSHOT=on give an error
 ifeq ($(SNAPSHOT), on)
-EXT_JERRY_FLAGS += -DFEATURE_JS_PARSER=OFF
+$(error IDE does not support SNAPSHOT $(SNAPSHOT))
```
jimmy-huang: maybe just print IDE doesnot support snapshot mode, since $(SNAPSHOT) is always on here, no need to print it

brianjjones: It needs to be an error, if the user passes in SNAPSHOT=on I can't turn it off.  Which means it won't include the parser, which means ashell will fail.

Makefile
brianjjones: They do, but they get it from a few lines above this.  However this does point out that this else isn't necessary, I'll remove it.  Thanks.

jimmy-huang:
+1

Makefile
```diff
@@ -51,11 +51,21 @@ EXT_JERRY_FLAGS ?=	-DENABLE_ALL_IN_ONE=ON \
 					-DFEATURE_PROFILE=$(ZJS_BASE)/jerry_feature.profile \
 					-DFEATURE_ERROR_MESSAGES=OFF \
 					-DJERRY_LIBM=OFF
-ifneq ($(DEV), ashell)
+# Settings for ashell builds
+ifneq (,$(filter $(MAKECMDGOALS),ide ashell))
+SNAPSHOT=off
```
grgustaf: I don't think this is needed, right? It's already off. If it's on, you're about to give an error.

brianjjones: Line 8 sets SNAPSHOT=on by default, if you take that line out it fails.  So it's necessary, unless we don't make SNAPSHOT=on by default.

grgustaf: Oh I see, so now you're taking advatage of what you learned, that SNAPSHOT=off doesn't actually work if the user has forced SNAPSHOT=on?

Makefile
```diff
@@ -51,11 +51,21 @@ EXT_JERRY_FLAGS ?=	-DENABLE_ALL_IN_ONE=ON \
...
+# if the user passes in SNAPSHOT=on give an error
 ifeq ($(SNAPSHOT), on)
-EXT_JERRY_FLAGS += -DFEATURE_JS_PARSER=OFF
+$(error IDE does not support SNAPSHOT=$(SNAPSHOT))
```
grgustaf: Please consolidate the setting of SNAPSHOT and this error message, along with the other two places SNAPSHOT is set (at the top, and for linux case) all in one place.

Otherwise, LGTM +1.

Makefile
```diff
@@ -4,8 +4,19 @@ ROM ?= 256
...
+ifneq (,$(filter $(MAKECMDGOALS),ide ashell linux))
+SNAPSHOT=off
+# if the user passes in SNAPSHOT=on for ide, ashell, or linux give an error
+ifeq ($(SNAPSHOT), on)
+$(error ide, ashell, and linux do not support SNAPSHOT=$(SNAPSHOT))
+endif
+else
+# snapshot is enabled by default
 SNAPSHOT ?= on
+EXT_JERRY_FLAGS += -DFEATURE_JS_PARSER=OFF
```
grgustaf: Putting this here looks like a bug. if the user set it to SNAPSHOT=off, it will be off even though you attempted to turn it on here. So you still need to tests actual SNAPSHOT value as before.

PR #848, opened 2017-03-21
[build] Re-enable warnings as errors

grgustaf:
Add one exception for format errors to allow iotivity-constrained build
to succeed, but we should follow up with a patch to that project.

Also, add the flags to linux and ashell builds, for the first time.


jimmy-huang:
lgtm, +1

Issue #757, opened 2017-02-21
[REG] Unable to allocate buffer on Arduino 101


qiaojingx:
### Issue Description
Unable to allocate buffer on Arduino 101

### Test Code
[HeartRateDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/HeartRateDemo.js)
[WebBluetoothDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/WebBluetoothDemo.js)
[WebBluetoothGroveLcdDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/WebBluetoothGroveLcdDemo.js)

### Actual Result
"[ERROR]Unable to allocate buffer" will be outputted after running these samples for a while.
Detail info please refer to serial logs.

### Expected Result
No error message come out.

### Test Builds
Branch |Commit Id |Target Device |Test Date |Result
---|---|---|---|---
master|3f8eb85| Arduino 101 | Feb 13, 2017 | Pass
master|7c5ff3c| Arduino 101 | Feb 13, 2017 | Fail

### Additional Information
Serial log for: [WebBluetoothGroveLcdDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/WebBluetoothGroveLcdDemo.js)
```
Boot timestamp: 60 

ipm_console0: 'Sensor core running ZJS ARC support image'
WebBluetooth Demo with Grove LCD...
Advertising as Physical Web device
Client connected: cc:fa:00:73:4f:dc (public)
Temperature change: 21C
Temperature change: 23C
Temperature change: 24C
Temperature change: 23C
Temperature change: 24C
Temperature change: 23C

src/zjs_buffer.c:347 zjs_buffer_create():
[ERROR] unable to allocate buffer

src/zjs_util.c:368 zjs_print_error_message():
[ERROR] Uncaught exception: Error: 
Temperature change: 24C

src/zjs_buffer.c:347 zjs_buffer_create():
[ERROR] unable to allocate buffer

src/zjs_util.c:368 zjs_print_error_message():
[ERROR] Uncaught exception: Error: 
Temperature change: 23C
Temperature change: 24C
```
Serial log for: [WebBluetoothDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/WebBluetoothDemo.js)
```
ipm_console0: 'Sensor core running ZJS ARC support image'

WebBluetooth Demo with LED Bulb...
Advertising as Physical Web device
[...]

PR #849, opened 2017-03-21
[util] Run JrS garbage collection and retry upon malloc failure

grgustaf:
If we get a malloc failure, it may be because there are JavaScript
objects that hold references to C memory through handles that can't
be freed until they are garbage collected.

This fixes issues #757.

### DIFF
--- a/src/zjs_util.c
+++ b/src/zjs_util.c
@@ -5,15 +5,12 @@
 // ZJS includes
 #include "zjs_util.h"
 
-static int zjs_malloc_error_count = 0;
-
 void *zjs_malloc_with_retry(size_t size)
 {
     void *ptr = malloc(size);
     if (!ptr) {
         // see if stale JerryScript objects are holding memory
         jerry_gc();
-        ++zjs_malloc_error_count;
         ptr = malloc(size);
     }
     return ptr;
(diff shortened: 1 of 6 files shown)


src/zjs_util.c
```diff
@@ -5,6 +5,20 @@
...
 
+static int zjs_malloc_error_count = 0;
+
+void *zjs_malloc_with_retry(size_t size)
+{
+    void *ptr = malloc(size);
+    if (!ptr) {
+        // see if stale JerryScript objects are holding memory
+        jerry_gc();
+        ++zjs_malloc_error_count;
```
jimmy-huang: where is zjs_malloc_error_count used?

grgustaf:
It's not, that's kinda a TODO. I'm not sure where to put it, have any good ideas? Maybe I add a new function and check from the event loop.

I didn't want to slow down at the failure point with a print though, sorta like the ringbuf push error.

grgustaf:
OK, rethinking that I don't think counting those is necessary. I thought it might be useful to the user to know they were almost running out of RAM, but it's really not important. It's fundamentally more correct to run the GC in this situation.

jimmy-huang:
+1

PR #856, opened 2017-03-21
[general] Fix format warnings that appear with SDK 0.9

grgustaf:
uint32_t is now (properly) defined as unsigned int, instead of unsigned
long, which we'd worked around with incorrect format specifiers for the
last year.

Fixed the warnings from building HelloWorld and WebBT demo so far.


jimmy-huang:
+1

PR #857, opened 2017-03-21
[dgram] Fix error introduced by ZVAL patch, add dgram to trlite

grgustaf:
Now dgram module will build in trlite to catch future bugs earlier.


jimmy-huang:
+1

PR #858, opened 2017-03-21
[i2c] Disallow invalid bus #1 on Arduino 101

brianjjones:
I2C will now throw an error if you try and access the
I2C_SS_1 bus on Arduino 101, since it's not connected
to anything.

### DIFF
--- a/src/zjs_i2c_handler.c
+++ b/src/zjs_i2c_handler.c
@@ -2,9 +2,6 @@
 #include "zjs_i2c_handler.h"
 #include "zjs_common.h"
 
-#define I2C_0 0
-#define I2C_1 1
-
 #ifdef CONFIG_BOARD_ARDUINO_101_SSS
 #define MAX_I2C_BUS 1
 #else
--- a/src/zjs_i2c_handler.c
+++ b/src/zjs_i2c_handler.c
@@ -2,6 +2,8 @@
 #include "zjs_i2c_handler.h"
 #include "zjs_common.h"
 
+// The Arduino 101 has two I2C busses, but I2C_SS_1 isn't connected to anything.
+// So we have disabled the ability to connect to it, to avoid user confusion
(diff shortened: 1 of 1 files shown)


src/zjs_i2c_handler.c
```diff
@@ -4,7 +4,12 @@
 
 #define I2C_0 0
 #define I2C_1 1
+
+#ifdef CONFIG_BOARD_ARDUINO_101_SSS
+#define MAX_I2C_BUS 1
+#else
 #define MAX_I2C_BUS 2
+#endif
```
grgustaf: Per IRC discussion we'll need to work on pulling this value from Zephyr config flags or something somehow, more generically. But good enough for now.

src/zjs_i2c_handler.c
```diff
@@ -4,7 +4,12 @@
 
 #define I2C_0 0
 #define I2C_1 1
```
grgustaf: Please remove these unused #define's while we're here.

Otherwise, +1.

src/zjs_i2c_handler.c
jimmy-huang: I am curious, if I2C_1 is a valid bus on Arduino101, eventhough it's not connected to anything, shouldn't we still allow app to open the bus? So instead of fixing the module to limit to only 1 bus, the test case was invalid, since the Arduino 101 and frdm is different

brianjjones: @jimmy-huang I talked with @grgustaf about this, we determined it was better for the user experience to explicitly not allow it since it could lead to a headache on their end.  It literally can't do anything, so no point in letting them connect to it.  Should someone really want to, we can revisit that later :)

grgustaf:
In IRC, I think @jimmy-huang is okay with this. Ship it!

+1

PR #860, opened 2017-03-22
[general] Fix build breakage: use non-const script str in Linux build

grgustaf:


jimmy-huang:
+1

PR #866, opened 2017-03-22
[docs] Added QEMU networking docs and created README index

jprestwo:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,18 +11,10 @@ everything else.
 
 ### Index
 * [Getting Started](#getting-started)
-* [Prerequisites](#prerequisites)
-* [Initial Setup](#initial-setup)
-* [Shell Setup](#shell-setup)
-* [Build and Flash](#build-and-flash)
 * [Next steps](#next-steps)
 * [Contributing](#contributing)
-* [File Descriptions](#file-descriptions)
-* [Subdirectories](#subdirectories)
+* [Repo Organization](#repo-organization)
 * [Increase space on A101](#getting-more-space-on-your-arduino-101)
-* [Building system images](#building-system-images)
-* [Building the Hello World sample](#building-the-hello-world-sample)
-* [Build other samples](#build-other-samples)
 * [JS Minifier](#js-minifier)
 * [FRDM-K64F Platform](#frdm-k64f-platform)
 * [Building and running on Linux/Mac OSX](#building-and-running-on-linux-and-mac-osx)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,27 @@ This code requires a local copy of JerryScript and Zephyr OS source.  We
 will upstream patches to those projects as appropriate, but this repo is for
 everything else.
 
+### Index
```
grgustaf: This index is tl;dr. :)

README.md
```diff
@@ -318,7 +339,7 @@ install this earlier, you can do so with the command:
 sudo apt-get install node-uglify
 ```
 
-## FRDM-K64F Platform
+## FRDM K64F Platform
```
grgustaf: The dash is [how they identify it](https://developer.mbed.org/platforms/FRDM-K64F/).

README.md
```diff
@@ -376,7 +397,7 @@ Using the same procedure as above, once you hit Reset you should see
 
 Zephyr is a trademark of the Linux Foundation. *Other names and brands may be claimed as the property of others.
 
-## Building and running on Linux/Mac OSX
+## Building and running on Linux and Mac OSX
 In addition to Zephyr there is a "linux" target which does not use Zephyr at all
 and instead uses the host OS. This can be build on Linux or Mac OSX using the
```
grgustaf: built, while you're here :)

pfalcon:
Having such section is definitely a good idea, thanks. I won't argue that *instead* of duplicating instructions from https://www.zephyrproject.org/doc/subsystems/networking/qemu_setup.html, it should link there, but I'd say it definitely should link there *in addition* to those quick inline instructions, in case something changes on Zephyr side, or just for additional info.

README.md
jprestwo: I couldn't link the section with the dash. The (#section-name) uses dashes where there are spaces. I am not very proficient at markdown but if you know how to escape the dashes so it will work in the link I will do that.

README.md
jprestwo: Yeah I didn't want to open up that can of worms. Really, I started by just adding the QEMU docs, but then realized how terrible and long the readme was. I can though, we just may want to talk about what our main bullets should be, and then sort the existing sections into those categories.

README.md
grgustaf: If you look, the link already works, like this:
https://github.com/01org/zephyr.js#frdm-k64f-platform

So I think it would just work w/ the same one you used, with dashes.

grgustaf:
+1

PR #867, opened 2017-03-23
[sensor] support for different controllers in constructor

jimmy-huang:
Allow app to pass sensor driver name by setting controller
string in constructor, for Accelerometer and Gyroscope, the
default controller is "bmi160"


grgustaf:
+1

rwaldron:
Interesting... Can you share an example program that utilizes this feature?

PR #868, opened 2017-03-23
[trlite] Check that copyrights have been updated in modified files

grgustaf:
Only works for .c, .h, and .js files currently. Checks files modified in the
current commit. Checks that the very first line contains a copyright and that
it contains the current year in it.


jimmy-huang:
lgtm, +1

PR #869, opened 2017-03-23
[general] Add some copyrights to some makefiles

grgustaf:
These have reached code-like significance and effort.

Also, clean up a few things.


jimmy-huang:
+1

PR #871, opened 2017-03-23
[snapshot] Fixed broken snapshot build due to updated JerryScript

jimmy-huang:
JerryScript has changed the snapshot bytecode from uint8_t
to uint32_t, therefore we need to update the generated array
to store the increased byte code size

### DIFF
--- a/src/snapshot.c
+++ b/src/snapshot.c
@@ -26,14 +26,12 @@ static int generate_snapshot(const char *file_name, uint32_t *buf, int buf_size)
     fputs("#include \"zjs_common.h\"\n\n", f);
     fputs("const uint32_t snapshot_bytecode[] = {\n", f);
 
-    char byte[11];
     for (int i = 0; i < buf_size; i++)
     {
         if (i > 0) {
             fputs(",", f);
         }
         fprintf(f, "0x%08lx", (unsigned long)buf[i]);
-        DBG_PRINT("%s,", byte);
     }
 
     fputs("\n};\n\n", f);
--- a/src/snapshot.c
+++ b/src/snapshot.c
@@ -22,25 +22,23 @@ static int generate_snapshot(const char *file_name, uint32_t *buf, int buf_size)
         return 1;
     }
 
-    size_t bs = sizeof(char);
-    fwrite("/* This file was auto-generated */\n\n", bs, 36, f);
-    fwrite("#include \"zjs_common.h\"\n\n", bs, 25, f);
-    fwrite("const uint32_t snapshot_bytecode[] = {\n", bs, 38, f);
+    fputs("/* This file was auto-generated */\n\n", f);
+    fputs("#include \"zjs_common.h\"\n\n", f);
+    fputs("const uint32_t snapshot_bytecode[] = {\n", f);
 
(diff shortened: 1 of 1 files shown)


src/snapshot.c
```diff
@@ -22,24 +22,25 @@ static int generate_snapshot(const char *file_name, uint8_t *buf, int buf_size)
...
     for (int i = 0; i < buf_size; i++)
     {
         if (i > 0) {
-            fwrite(",", 1, 1, f);
+            fwrite(",", bs, 1, f);
         }
-        char byte[5];
-        snprintf(byte, 5, "0x%02x", snapshot_buf[i]);
-        fwrite(byte, 1, 4, f);
+        snprintf(byte, 11, "0x%08lx", (unsigned long)buf[i]);
```
grgustaf: Remove the l and make this (unsigned int) just for consistency with what I've been doing elsewhere.

And yeah I'd prefer to just leave the 'bs' var out, it's confusing I think.

src/snapshot.c
```diff
@@ -22,24 +22,25 @@ static int generate_snapshot(const char *file_name, uint8_t *buf, int buf_size)
         return 1;
     }
 
-    fwrite("/* This file was auto-generated */\n\n", 1, 36, f);
-    fwrite("#include \"zjs_common.h\"\n\n", 1, 25, f);
-    fwrite("const uint32_t snapshot_bytecode[] = {\n", 1, 38, f);
+    size_t bs = sizeof(char);
+    fwrite("/* This file was auto-generated */\n\n", bs, 36, f);
```
grgustaf: Oh, this is Linux... use fputs here instead of fwrite and manually counting the bytes, which is easy to break.

src/snapshot.c
```diff
@@ -22,24 +22,25 @@ static int generate_snapshot(const char *file_name, uint8_t *buf, int buf_size)
...
     {
         if (i > 0) {
-            fwrite(",", 1, 1, f);
+            fwrite(",", bs, 1, f);
         }
-        char byte[5];
-        snprintf(byte, 5, "0x%02x", snapshot_buf[i]);
-        fwrite(byte, 1, 4, f);
+        snprintf(byte, 11, "0x%08lx", (unsigned long)buf[i]);
+        fwrite(byte, bs, 10, f);
```
grgustaf: And here just use fprintf instead of snprintf + fwrite which makes it way simpler.

src/snapshot.c
```diff
@@ -22,24 +22,23 @@ static int generate_snapshot(const char *file_name, uint8_t *buf, int buf_size)
...
     {
         if (i > 0) {
-            fwrite(",", 1, 1, f);
+            fputs(",", f);
         }
-        char byte[5];
-        snprintf(byte, 5, "0x%02x", snapshot_buf[i]);
-        fwrite(byte, 1, 4, f);
+        fprintf(f, "0x%08lx", (unsigned long)buf[i]);
         DBG_PRINT("%s,", byte);
```
grgustaf: You're not really using byte anymore, right?

grgustaf:
+1

PR #877, opened 2017-03-23
[Buffer] Remove invalid case

haoxli:
- writeUInt8(-1) is not a valid check point following Buffer API

### DIFF
--- a/tests/test-buffer.js
+++ b/tests/test-buffer.js
@@ -27,8 +27,7 @@ assert(buff.length === original, "The length of Buffer is readonly");
 
 // Function: writeUInt8(unsigned char value, unsigned long offset)
 //           unsigned char readUInt8(unsigned long offset)
-var uints = [[-1, 0],
-             [0, 0],
+var uints = [[0, 0],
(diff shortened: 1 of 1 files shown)


grgustaf:
@haoxli, you're right: the test is invalid. I haven't really stated this intention explicitly before, but it's good for you to know. I mean for us to be aiming for subset compatibility with the LTS version of Node.js (which changes over time). Currently this is https://nodejs.org/dist/latest-v6.x/docs/api/. The way I find it is go to http://node.js.org, click on DOCS at the top, then on the left, the version of the API with small LTS next to it.

When we started the LTS version was 4.x so it's changed a lot and one example where we haven't kept up is the Buffer constructor. So feel free to file bugs when we're not following the LTS APIs.

So for this particular question, here's the ruling documentation:
https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_buf_writeuint8_value_offset_noassert

"Behavior is undefined when value is anything other than an unsigned 8-bit integer."

+1

PR #878, opened 2017-03-23
[trlite] Update to build ashell the new way

grgustaf:
Instead of DEV=ashell you now pass an ashell or ide target. Currently,
then only differ by one config line but I build both in Travis to catch
possible future changes.

Also, fixed a build error and a few format warnings w/ 0.9 SDK.


jprestwo:
+1

PR #879, opened 2017-03-23
[sensor] Update AmbientLight.js sample

jimmy-huang:
Update the sample that was still using the old Sensor api


grgustaf:
+1

PR #881, opened 2017-03-23
[trlite] Split out a fourth VM to reduce Travis latency

grgustaf:
VM1 is taking ~14min so this should cut us down to ~7min.


jimmy-huang:
+1

PR #882, opened 2017-03-23
[readme] Update with new python-yaml dependency in Zephyr

grgustaf:

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,11 @@ default with Ubuntu:
 
 ```bash
 $ sudo apt-get update
-$ sudo apt-get install cmake dfu-util git screen sysvbanner uglifyjs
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,9 +33,12 @@ default with Ubuntu:
 
 ```bash
 $ sudo apt-get update
-$ sudo apt-get install cmake dfu-util git screen sysvbanner uglifyjs
+$ sudo apt-get install cmake dfu-util git python-yaml screen sysvbanner uglifyjs
 ```
 
+Note: python-yaml is a recent requirement for the frdm-k64f build due to some
+change in Zephyr, so it could be left out currently if you don't use k64f.
```
jimmy-huang: changes

jimmy-huang:
+1

PR #890, opened 2017-03-24
[ocf] Fix ocf.setBleAddress() and BLE_ADDR

jprestwo:
- Setting the BLE address was broken due to these two
   configurations conflicting

### DIFF
--- a/src/zjs_ocf_ble.c
+++ b/src/zjs_ocf_ble.c
@@ -59,6 +59,7 @@ jerry_value_t zjs_ocf_set_ble_address(const jerry_value_t function_val,
                                       const jerry_value_t argv[],
                                       const jerry_length_t argc)
 {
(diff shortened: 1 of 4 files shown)


src/zjs_ocf_ble.c
```diff
@@ -59,6 +59,7 @@ jerry_value_t zjs_ocf_set_ble_address(const jerry_value_t function_val,
                                       const jerry_value_t argv[],
                                       const jerry_length_t argc)
 {
+#ifndef ZJS_CONFIG_BLE_ADDRESS
```
grgustaf: I'm still not sure whether this makes sense or the JS should override command line. But document this behavior in the setBleAddress doc I guess, if it isn't already.

jprestwo: Documented with latest commit

grgustaf:
+1

PR #892, opened 2017-03-24
[snapshot] Clean up snapshot tool

jimmy-huang:
Move the source and Makefile to tools directory, update snapshot tool
and convert.sh to generate just the needed bytes into header file, and
clean up the Makefile to build using the generated header in the outdir,
this should clean up all the generated src files in the root directory


grgustaf:
+1

PR #894, opened 2017-03-24
[build] Changing DEV to ASHELL to make things clearer

brianjjones:
This will also cause DEV=ashell to no longer be supported. This
helps remove confusion about building ashell the proper way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,10 @@ EXT_JERRY_FLAGS += -DFEATURE_JS_PARSER=OFF
 endif
 endif
 
+ifneq (,$(DEV))
+$(error DEV= is no longer supported, please use make ide or make ashell)
+endif
+
 ifndef ZJS_BASE
 $(error ZJS_BASE not defined. You need to source zjs-env.sh)
 endif
--- a/Makefile
+++ b/Makefile
@@ -360,6 +360,8 @@ help:
 	@echo "Build targets:"
 	@echo "    all:       Build for either Zephyr or Linux depending on BOARD"
 	@echo "    zephyr:    Build Zephyr for the given BOARD (A101 is default)"
+	@echo "    ide        Build Zephyr in development mode for the IDE"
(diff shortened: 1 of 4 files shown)


grgustaf:
So just to clarify, I don't think we intend the user to pass DEV or ASHELL variable on the command line anymore; instead they should just make the 'ashell' or 'ide' target. Right?

So this is fine, it's just an internal variable. But please also add a check for the user setting DEV and tell them to run "make ashell" or "make ide" instead.

brianjjones:
Yes, DEV=ashell should never be used now.  However I'm hesitant to add a warning for obsolete build options, it seems like that adds unnecessary bloat to our already complex Makefile.  However, I'm all for making a generic "This is not a supported build option" message that then prints make help.  This informs the user what they are doing is wrong, and provides them with the list of correct options.  Would that be acceptable?

grgustaf:
That's not really going to work here because you can't check for every possible environment variable they have set. You have to specifically check for DEV for this to work. This is basically just a matter of having a clean deprecation path. It is mainly to help us, so that for the next few months if we type DEV=ashell out of habit, we will get an error message. After a while we can remove it.

grgustaf:
You could also throw a fix for #920 into this patch. (And maybe grep for any other references to DEV=)

jimmy-huang:
I think it would be nice to standardize some of the make commands, right now there's VARIANT=debug/release, SNAPSHOT=on/off, TRACE=on/off ASHELL=y/n, most if not all of these are true/false variants, why not just use the same for all of them, I am not sure which one is better.

grgustaf:
Well ASHELL is really an internal make variable the user shouldn't know about.

I think on/off are decent for the others. VARIANT is odd, yeah, but that can wait for another day.

brianjjones:
@grgustaf error added for DEV

grgustaf:
+1

Issue #797, opened 2017-03-07
AsyncGPIOPin.onchange and uart.on didn't work


qiaojingx:
### Description

AsyncGPIOPin.onchange and uart.on didn't work

### Test Code

[ButtonLEDsAsync.js](https://github.com/01org/zephyr.js/blob/master/samples/ButtonLEDsAsync.js); [UART.js](https://github.com/01org/zephyr.js/blob/master/samples/UART.js); [test-gpio.js](https://github.com/01org/zephyr.js/blob/master/tests/test-gpio.js); [test-uart-read-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-uart-read-manual.js)
 
### Steps to Reproduction

1. `$ make JS=samples/UART.js`
2. `$ make dfu`
3. Reset device. 

### Actual Result

[ButtonLEDsAsync.js](https://github.com/01org/zephyr.js/blob/master/samples/ButtonLEDsAsync.js)
 Only one on board LED is always bright, not blinking.

[UART.js](https://github.com/01org/zephyr.js/blob/master/samples/UART.js)
![uart](https://cloud.githubusercontent.com/assets/19813546/23649995/773f7fdc-035b-11e7-99b1-e76e89d1ff7c.png)

[test-uart-read-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-uart-read-manual.js)
![test-uart-manual](https://cloud.githubusercontent.com/assets/19813546/23650018/9735e1c8-035b-11e7-9cd0-1b8bf1b91336.png)

[test-gpio.js](https://github.com/01org/zephyr.js/blob/master/tests/test-gpio.js)
Test stuck at the tenth case.
![test-gpio](https://cloud.githubusercontent.com/assets/19813546/23650004/8194129a-035b-11e7-88ed-a3b0054d4b02.png)

### Expected Result

[ButtonLEDsAsync.js](https://github.com/01org/zephyr.js/blob/master/samples/ButtonLEDsAsync.js)
Blink the two LEDs on and off together

[UART.js](https://github.com/01org/zephyr.js/blob/master/samples/UART.js) and [test-uart-read-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-uart-read-manual.js)
No error thrown

[test-gpio.js](https://github.com/01org/zephyr.js/blob/master/tests/test-gpio.js)
All test cases are executed successfully.
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
[...]

PR #906, opened 2017-03-27
[callbacks] Add check if JS function is undefined (#797)

jprestwo:
- Promises add undefined functions for then/catch, and if
   then/catch are never called, callbacks was signaling on
   an undefined function.


grgustaf:
+1

PR #910, opened 2017-03-28
[buffer] Return a 0-length buffer if given negative size

grgustaf:
This follows Node.js docs for v6.10.1, although the constructor like
this is deprecated and really should change to Buffer.alloc().

### DIFF
--- a/src/zjs_buffer.c
+++ b/src/zjs_buffer.c
@@ -379,9 +379,13 @@ static jerry_value_t zjs_buffer(const jerry_value_t function_obj,
     ZJS_VALIDATE_ARGS(Z_NUMBER Z_ARRAY Z_STRING);
 
     if (jerry_value_is_number(argv[0])) {
+        double dnum = jerry_get_number_value(argv[0]);
+        if (dnum < 0) {
+            dnum = 0;
+        }
+
         // treat a number argument as a length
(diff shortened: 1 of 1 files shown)


src/zjs_buffer.c
```diff
@@ -379,9 +379,13 @@ static jerry_value_t zjs_buffer(const jerry_value_t function_obj,
...
     if (jerry_value_is_number(argv[0])) {
+        double dnum = jerry_get_number_value(argv[0]);
+        if (dnum < 0) {
+            dnum = 0;
+        }
+
         // treat a number argument as a length
-        uint32_t size = (uint32_t)jerry_get_number_value(argv[0]);
-        return zjs_buffer_create(size);
+        return zjs_buffer_create((uint32_t)dnum);
```
jimmy-huang: @grgustaf so what happens when you call underlying malloc with 0? it returns NULL pointer? then the api will return an error? or a actuall jerry_value_t?

grgustaf: malloc(0) works and returns you a pointer that you're not supposed to write to I suppose. :)

So it all works fine. This is the same path that would happen if you call zjs_buffer_create(0) which has always been tested by QA and works.

jimmy-huang:
+1

PR #911, opened 2017-03-28
Fix dgram malloc bug

grgustaf:


pfalcon:
Oops, that's a real thinko-typo, thanks for spotting it. Weird that I never seem to have seen any effect of it in my testing %).

+1

PR #912, opened 2017-03-28
[sensor] Fixed bug where controller name can't get set

jimmy-huang:
Change controller from a char* to array so that it can be overwritten,
akso make default controller for AmbientLight to be "ADC_0" and
fix a leak

### DIFF
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -526,8 +526,7 @@ static jerry_value_t zjs_sensor_create(const jerry_value_t function_obj,
         jerry_value_t options = argv[0];
         ZVAL controller_val = zjs_get_property(options, "controller");
 
-        if (!jerry_value_is_undefined(controller_val) &&
-            jerry_value_is_string(controller_val)) {
+        if (jerry_value_is_string(controller_val)) {
             zjs_copy_jstring(controller_val, controller, &size);
         } else {
             switch(channel) {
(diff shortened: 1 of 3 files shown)


src/zjs_sensor.c
```diff
@@ -510,30 +515,37 @@ static jerry_value_t zjs_sensor_create(const jerry_value_t function_obj,
...
 
     if (argc >= 1) {
         jerry_value_t options = argv[0];
-        const int BUFLEN = 20;
-        char buffer[BUFLEN];
+        ZVAL controller_val = zjs_get_property(options, "controller");
 
-        if (!zjs_obj_get_string(options, "controller", buffer, BUFLEN)) {
+        if (!jerry_value_is_undefined(controller_val) &&
+            jerry_value_is_string(controller_val)) {
```
grgustaf: If it is string than it's already not undefined. :)

src/zjs_sensor.c
```diff
@@ -17,7 +17,10 @@
 
 #define ZJS_SENSOR_TIMEOUT_TICKS 5000
 #define DEFAULT_SAMPLING_FREQUENCY 20
+
+#define SENSOR_MAX_CONTROLLER_LEN 16
 #define BMI160_CONTROLLER "bmi160"
+#define ANALOG_CONTROLLER "ADC_0"
```
grgustaf: How about make this common with ADC_DEVICE_NAME elsewhere.

jimmy-huang:
@grgustaf updated patch

grgustaf:
+1

PR #922, opened 2017-03-28
[ocf] Fixed null termination bug

jprestwo:

### DIFF
--- a/src/zjs_ocf_server.c
+++ b/src/zjs_ocf_server.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2016-2017, Intel Corporation.
 
 #ifdef BUILD_MODULE_OCF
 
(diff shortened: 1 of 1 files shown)


src/zjs_ocf_server.c
```diff
@@ -150,7 +150,7 @@ struct server_resource *new_server_resource(char *path)
     struct server_resource *resource = zjs_malloc(sizeof(struct server_resource));
     memset(resource, 0, sizeof(struct server_resource));
 
-    resource->resource_path = zjs_malloc(strlen(path));
+    resource->resource_path = zjs_malloc(strlen(path) + 1);
     memcpy(resource->resource_path, path, strlen(path));
     resource->resource_path[strlen(path)] = '\0';
```
grgustaf: I think they call this strcpy. :) But we want to use strncpy. Also I don't like the calling strlen 3x in a row on the same thing. And we're actually recommended to use strnlen too.

So go make your life a little more difficult first. :)

src/zjs_ocf_server.c
```diff
@@ -150,9 +150,10 @@ struct server_resource *new_server_resource(char *path)
     struct server_resource *resource = zjs_malloc(sizeof(struct server_resource));
     memset(resource, 0, sizeof(struct server_resource));
 
-    resource->resource_path = zjs_malloc(strlen(path));
-    memcpy(resource->resource_path, path, strlen(path));
-    resource->resource_path[strlen(path)] = '\0';
+    uint32_t len = strlen(path);
+    resource->resource_path = zjs_malloc(len + 1);
+    strncpy(resource->resource_path, path, len);
+    resource->resource_path[len] = '\0';
```
grgustaf: Really you should check results of malloc twice in this function.

And if you do strncpy on len + 1 you apparently wouldn't need to manually set the '\0'. (Verify, but that seems to be what the manpage says.)

grgustaf:
+1

PR #923, opened 2017-03-28
[snapshot] Removed dbg print and redireted error to stderr

jimmy-huang:


grgustaf:
+1

PR #928, opened 2017-03-28
[pwm] Require PWM period/pulseWidth to be set at the same time

### LINKED ISSUE #359: [PWM] Error thrown when set pulsewidth before period
```
var pinA = pwm.open({ channel: pins.IO3 });
pinA.setPeriod(1000);
pinA.setPulseWidth(300);
```

It work well.

```
pinA.setPulseWidth(300);
pinA.setPeriod(1000);
```

**error message:**

`zjs_pwm_set: pulseWidth was greater than period`

grgustaf:
This is a wise API choice because you can go through invalid states
by setting the wrong one first, where the pulse width is bigger than
the period. I convinced Zephyr itself to make this change, but we
hadn't gotten around to reflecting this in our own API.

Fixes #359.

### DIFF
--- a/docs/pwm.md
+++ b/docs/pwm.md
@@ -54,10 +54,8 @@ dictionary PWMInit {
 
 [NoInterfaceObject]
 interface PWMPin {
-    void setPeriod(double ms);
-    void setPeriodCycles(unsigned long cycles);
-    void setPulseWidth(double ms);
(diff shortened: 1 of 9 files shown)


grgustaf:
I updated and tested all five of the affects samples/tests in our tree, so I'm confident in the patch.

jimmy-huang:
+1

PR #930, opened 2017-03-29
[docs] Document caveat about conflict between SPI and A101 LED

### LINKED ISSUE #693: ashell: LED0 does not work on Arduino 101
The LED0 (onboard) does not work with the ashell (master, d6f1ce185212e892c73dbacd0b865717600b09c8)

I tried with following samples 

samples/arduino/basics/Blink.js
samples/arduino/digital/Button.js
samples/TrafficLight.js

They do not work with shell but do work without it.

grgustaf:
Also, rename the LEDs so the "bad" LED that breaks when SPI is enabled
is LED2 instead of LED0. Update samples to use the new names or switch
to using LED0 as the default.

Fixes #693.

### DIFF
--- a/docs/a101_pins.md
+++ b/docs/a101_pins.md
@@ -74,13 +74,15 @@ inputs or outputs.
 The Arduino 101 has three onboard LEDs that you can access as additional GPIO
 outputs.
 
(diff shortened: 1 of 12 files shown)


jimmy-huang:
+1

PR #931, opened 2017-03-29
[fs] Enable FS async API's automatically if script uses them

jprestwo:

### DIFF
--- a/scripts/analyze.sh
+++ b/scripts/analyze.sh
@@ -248,6 +248,12 @@ fi
 
 if check_for_require fs || check_config_file ZJS_FS; then
     >&2 echo Using module: FS
+    OBJ_NAME=$(grep "require *( *['\"]fs['\"] *)*;" $SCRIPT | cut -d'=' -f1 | cut -d' ' -f2)
+
+    if grep "$OBJ_NAME\..*(" tests/test-fs-async.js | grep -q -v "Sync"; then
+        MODULES+=" -DZJS_FS_ASYNC_APIS"
+    fi
+
     MODULES+=" -DBUILD_MODULE_FS -DBUILD_MODULE_BUFFER"
     if [ $BOARD = "arduino_101" ]; then
         echo "CONFIG_FS_FAT_FLASH_DISK_W25QXXDV=y" >> $PRJFILE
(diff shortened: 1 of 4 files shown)


scripts/analyze.sh
```diff
@@ -248,6 +248,12 @@ fi
 
 if check_for_require fs || check_config_file ZJS_FS; then
     >&2 echo Using module: FS
+    OBJ_NAME=$(grep "require *( *['\"]fs['\"] *)*;" $SCRIPT | cut -d'=' -f1 | cut -d' ' -f2)
+
+    if grep "$OBJ_NAME\..*(" tests/test-fs-async.js | grep -q -v "Sync"; then
```
grgustaf: This refers to a test that's maybe only on your machine? I suggest you do submit a simple sample at least that uses this because there doesn't seem to be one.

I haven't figured out what this does but hardcoding that test here probably isn't right?

src/zjs_fs.c
```diff
@@ -261,13 +261,14 @@ static jerry_value_t zjs_open(const jerry_value_t function_obj,
...
+
+        jerry_value_t args[] = {err, fd_val};
 
-        jerry_value_t args[2];
-        args[0] = jerry_create_number(handle->error);
-        args[1] = handle->fd_val;
+        zjs_callback_id id = zjs_add_callback_once(argv[2], this, handle, NULL);
 
-        zjs_signal_callback(id, args, sizeof(args))
+        zjs_signal_callback(id, args, sizeof(args));
```
grgustaf: Funny what can pass muster when it's ifdef'd out. :)

src/zjs_fs.c
```diff
@@ -992,9 +992,10 @@ static jerry_value_t zjs_write_file(const jerry_value_t function_obj,
 Finished:
 #ifdef ZJS_FS_ASYNC_APIS
     if (async) {
+        ZVAL err = jerry_create_number(error);
+
```
grgustaf: Kind of excessive vertical whitespace throughout here. Every line doesn't need to be its own paragraph. :) I'd remove some.

scripts/analyze.sh
jprestwo: Yes, I don't know how I didn't catch this. The test part is supposed to be $SCRIPT!

grgustaf:
+1

PR #933, opened 2017-03-29
[tests] Fix exception thrown after grove lcd manual tests

jimmy-huang:
The index counter has been incremented passed the length of
the array and therefore accessing the array throws an undefined
object exception.  The fix here to simply return the function
to abort the rest of the tests after tests are complete.


tests/test-grovelcd-ds-manual.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2017, Intel Corporation.
```
grgustaf: Use 2016-2017 to preserve original creation date.

PR #938, opened 2017-03-29
[trlite] Only pester Intel folks about updating copyright year

grgustaf:


jimmy-huang:
lgtm, at least only Intel people gets burned :)

PR #939, opened 2017-03-29
[samples] Added missing colon and renamed samples

jimmy-huang:
Renaming both Accelerometer and Gyroscope samples to
indicate they are BMI160 specific


grgustaf:
You need to change links in sensor.md. That's all I see grepping, but pause for a second to think if anything else would be affected.

Otherwise, LGTM.

jimmy-huang:
Yeah, updated doc

grgustaf:
+1

PR #941, opened 2017-03-29
[arc] Remove strnlen and use strlen

jimmy-huang:
strnlen is supported in NEWLIB and if we build ARC with
NEWLIB we no longer need our own strnlen()

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -72,18 +72,6 @@ static double accel_last_value[3];
 static double gyro_last_value[3];
 #endif
 
-// add strnlen() support for security since it is missing
-// in Zephyr's minimal libc implementation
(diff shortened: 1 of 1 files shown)


grgustaf:
OK, newlib is making ARC bigger and breaking the build, so better to leave it out.

But here's a better solution: drop strnlen and just use strlen() there. The snprintf on the line before it guarantees to write a '\0' to the buffer, so it's safe. (I just tested that and found it in the man page.)

I know we previously said not to use strlen, but we were stupid. See the [note here](https://msdn.microsoft.com/en-us/library/z50ty2zh.aspx#Remarks).

jimmy-huang:
@grgustaf updated patch and commit logs

grgustaf:
+1

PR #942, opened 2017-03-29
[sensor] Add missing timestamp

jimmy-huang:
Also fixed bug where default NULL values are not set when no
arguments are passed in the constructor

### DIFF
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -245,7 +245,7 @@ static void zjs_sensor_update_reading(jerry_value_t obj,
         return;
     }
 
-    uint32_t timestamp = k_uptime_get();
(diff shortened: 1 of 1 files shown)


src/zjs_sensor.c
```diff
@@ -245,6 +245,9 @@ static void zjs_sensor_update_reading(jerry_value_t obj,
         return;
     }
 
+    uint32_t timestamp = k_uptime_get();
```
grgustaf: k_uptime_get returns a uint64_t - if you only need the bottom part use k_uptime_get_32() - it's faster at least.

Otherwise LGTM.

jimmy-huang: updated with uint64_t

grgustaf:
+1

PR #946, opened 2017-03-30
[analyze] Fix OCF client or server being allowed but not both

grgustaf:


jprestwo:
+1

PR #949, opened 2017-03-31
[mac] fix build issues for BOARD=linux build

poussa:
On Mac with clang we need these fixes to build: make BOARD=linux

Signed-off-by: Sakari Poussa <sakari.poussa@intel.com>

### DIFF
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -56,8 +56,7 @@ LINUX_FLAGS += 	-fno-asynchronous-unwind-tables \
 		-fno-stack-protector \
 		-ffunction-sections \
 		-fdata-sections \
-		-fno-defer-pop \
-		-Wno-pointer-sign \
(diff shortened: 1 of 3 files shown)


jimmy-huang:
Lgtm, other than the facat that this patch only fixes some of the issues, but doesn't completely fix the build issue, at least for me, there are a bunch of other warnings, I had to update a bunch of packages like Xcode and Python and too, so we can merge this for now and I am creating a PR to submit a fix the rest of the issues.

jimmy-huang:
@poussa here's the patch I am working on which will including some of your changes https://github.com/jimmy-huang/zephyr.js/tree/mac-build-travis

poussa:
@jimmy-huang your patch is more complete and covers more issues. I'll drop this and let's work on yours.

Issue #916, opened 2017-03-28
[buffer] malloc failed in buffer


qiaojingx:
### Description

malloc failed in buffer
 
### Test Code

[test-buffer.js](https://github.com/01org/zephyr.js/blob/master/tests/test-buffer.js)
 
### Steps to Reproduction

1. `make JS=tests/test-buffer.js`
2. `make dfu`
 
### Actual Result

![selection_005](https://cloud.githubusercontent.com/assets/19813546/24389475/72e707d4-13b4-11e7-8675-36154e322cba.png)
 
### Expected Result

No error thrown
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 02d9471| Arduino 101 | Mar 23, 2017 | Pass
master | 02d9471| FRDM-K64F | Mar 23, 2017 | Pass
master | bb11854| Arduino 101 | Mar 28, 2017 | Fail
master | bb11854| FRDM-K64F | Mar 28, 2017 | Fail

grgustaf:
This is actually not a bug. Obviously we can't actually create a 4GB buffer on an A101. So it should have always failed like this.

The [Node docs](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_new_buffer_size) say that buffers over [kMaxLength](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_buffer_kmaxlength) should throw a RangeError. On a 32-bit platform like this that should be basically 1GB. However, we only have 80K of RAM on this platform so obviously that's still ridiculous. So it doesn't really make sense to follow this part of the API. Instead I think for now the right thing to do is just throw an error when malloc fails, like we're doing. The max you can really allocate will depend on memory conditions but in the best case it would be something like 20K.

grgustaf:
We should update the test case; there's not really going to be a good "maximum" value to use, but you could check to make sure an error is thrown when you use a very large size like 1GB that is obviously outside the scope of ZJS target devices. Linux can run on systems as small as 8MB RAM, and Node.js can run on systems with as little as 64MB, so anything that big would always be out of scope for ZJS.

grgustaf:
I just submitted #921 to update the docs to be more clear on intended constructor behavior.

cuiyanx:
@grgustaf  Please throw out an error when too long or not enough available memory.
I test this, just console log, not error.

cuiyanx:
Will be fix by #944 .

[...]

PR #955, opened 2017-03-31
[buffer] Return an error when out of memory in zjs_buffer_create

grgustaf:
Also, return the handle in a new out param to simplify detecting the
error case from the caller.

Addresses a request in #916.

### DIFF
--- a/src/zjs_buffer.c
+++ b/src/zjs_buffer.c
@@ -333,21 +333,23 @@ static jerry_value_t zjs_buffer_write_string(const jerry_value_t function_obj,
     return jerry_create_number(length);
 }
 
-jerry_value_t zjs_buffer_create(uint32_t size)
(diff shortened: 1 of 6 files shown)


src/zjs_buffer.c
```diff
@@ -414,12 +416,12 @@ static jerry_value_t zjs_buffer(const jerry_value_t function_obj,
         jerry_size_t size = 0;
         char *str = zjs_alloc_from_jstring(argv[0], &size);
         if (!str) {
-            return zjs_error("zjs_buffer: could not allocate string");
+            return zjs_error("could not allocate string");
```
jimmy-huang: possible leak here, shouldn't you release the new_buf and also free the *buf created above?

grgustaf: This one you've got to look at the actual code, not just the diff... :) It's a separate else section... but thanks for thinking about the mallocs.

jimmy-huang:
+1

PR #957, opened 2017-03-31
[trlite/analyze] Changed trlite back to ROM=250

jprestwo:
- Fixed analyze so that it wont compile in FS async with
   ashell builds


grgustaf:
+1

PR #958, opened 2017-03-31
[trlite] Build OCF client/server code, not just common

grgustaf:
We were mistakenly leaving out most of the OCF code in our test builds.


jimmy-huang:
+1

PR #959, opened 2017-03-31
[ashell] Adding javascript module functionality

brianjjones:
Now javascript files that contain require("file.js") will load
that javascript module before running the javascript.  It is
recursive as well, so files that get included will have their
requires checked as well.

### DIFF
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2016-2017, Intel Corporation.
 
 /**
  * @file
@@ -72,7 +72,7 @@ const char *system_get_prompt()
     return system_prompt;
 }
 
-#define MAX_LINE_LEN 80
+#define MAX_LINE_LEN 90
 #define FIFO_CACHE 2
 
 /* Configuration of the callbacks to be called */
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -382,8 +382,7 @@ int javascript_parse_code(const char *file_name, bool show_lines)
         }
 
         // Find and load all required js modules
-        if (load_require_modules(buf))
-        {
+        if (load_require_modules(buf)) {
             /* Setup Global scope code */
             parsed_code = jerry_parse((const jerry_char_t *) buf, size, false);
             if (jerry_value_has_error_flag(parsed_code)) {
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2016-2017, Intel Corporation.
 
 /**
  * @file
@@ -14,7 +14,7 @@
 
 /* JerryScript includes */
 #include "jerry-port.h"
-
+#include "jerry-code.h"
 #include "file-utils.h"
 
 /* Zephyr.js init everything */
@@ -32,7 +32,53 @@ void jerry_port_default_set_log_level(jerry_log_level_t level); /** Inside jerry
 
 static jerry_value_t parsed_code = 0;
 
-#define MAX_BUFFER_SIZE 4096
+typedef struct requires_list {
+    char *file_name;
+    struct requires_list *next;
+} requires_list_t;
+
+static requires_list_t *req_list = NULL;
+static requires_list_t *loaded_list = NULL;
+
+static bool is_loaded(const char *file_name)
+{
+    if(loaded_list) {
+        requires_list_t *cur = loaded_list;
+        while (cur) {
+            if (strcmp(file_name, cur->file_name) != 0) {
+                cur = cur->next;
+            }
+            else {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+static void push_to_list(requires_list_t **list, const char *item)
+{
+    requires_list_t *new = zjs_malloc(sizeof(requires_list_t));
+    size_t len = strlen(item) + 1;
+    new->file_name = (char*)zjs_malloc(len);
+    strncpy(new->file_name, item, len);
+    new->file_name[len] = '\0';
+    new->next = *list;
+    *list = new;
+}
+
+static void free_list(requires_list_t *list)
+{
+    if (list) {
+        requires_list_t *cur = list;
+        while (cur) {
+            zjs_free(cur->file_name);
+            list = list->next;
+            zjs_free(cur);
+            cur = list;
+        }
+    }
+}
 
 static void javascript_print_value(const jerry_value_t value)
 {
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
     jerry_port_console("\n");
 }
 
+static int read_file(const char *file_name, char **file_buf)
+{
+    fs_file_t *fp = fs_open_alloc(file_name, "r");
+
+    if (fp == NULL)
+        return 0;
+
+    ssize_t size = fs_size(fp);
+    if (size == 0) {
+        comms_printf("[ERR] Empty file (%s)\n", file_name);
+        fs_close_alloc(fp);
+        return 0;
+    }
+
+    *file_buf = (char *)zjs_malloc(size);
+    if (*file_buf == NULL) {
+        comms_printf("[ERR] Not enough memory for (%s)\n", file_name);
+        fs_close_alloc(fp);
+        return 0;
+    }
+
+    ssize_t brw = fs_read(fp, *file_buf, size);
+
+    if (brw != size) {
+        comms_printf("[ERR] Failed loading code %s\n", file_name);
+        fs_close_alloc(fp);
+        return 0;
+    }
+
+    return size;
+}
+
+static void find_requires(char *filebuf)
+{
+    char *ptr1 = filebuf;
+    char *ptr2 = NULL;
+    char *filestr = NULL;
+    char *ext_check = NULL;
+    while(ptr1 != NULL){
+
+        // Find next instance of "require"
+        ptr1 = strstr(ptr1, "require");
+
+        if (ptr1 != NULL) {
+            // Find the text between the two " " after require
+            ptr1 = strchr(ptr1, '"')+1;
+            ptr2 = strchr(ptr1, '"');
+            size_t len = ptr2-ptr1;
+
+            // Allocate the memory for the string
+            filestr = (char*)zjs_malloc(len+1);
+            strncpy(filestr, ptr1, len);
+            filestr[len] = '\0';
+
+            // Check that this is a *.js require and not a zephyr require
+            // Also check that the file hasn't already been loaded
+            ext_check = &filestr[len-3];
+
+            if (strcmp(ext_check, ".js") == 0 && !is_loaded(filestr))
+            {
+                push_to_list(&req_list, filestr);
+            }
+            zjs_free(filestr);
+            filestr = NULL;
+        }
+    }
+}
+
+static void load_require_modules(char *file_buffer)
+{
+    // Check if the file we've been asked to run requires any JS modules
+    find_requires(file_buffer);
+
+    // If it does, check if they have requires as well, and then eval the code
+    while (req_list) {
+        char *req_file_buffer = NULL;
+        uint file_size = read_file(req_list->file_name, &req_file_buffer);
+
+        if (file_size > 0)
+        {
+            find_requires(req_file_buffer);
+            javascript_eval_code(req_file_buffer, file_size);
+            zjs_free(req_file_buffer);
+            // Add file to the loaded list
+            push_to_list(&loaded_list, req_list->file_name);
+        }
+        // Go to the next item in the require list
+        requires_list_t *cur = req_list;
+        zjs_free(cur->file_name);
+        req_list = req_list->next;
+        zjs_free(cur);
+    }
+}
+
 static void javascript_print_error(jerry_value_t error_value)
 {
     if (!jerry_value_has_error_flag(error_value))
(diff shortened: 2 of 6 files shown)


src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+
+    ssize_t brw = fs_read(fp, *file_buf, size);
+
+    if (brw != size) {
+        comms_printf("[ERR] Failed loading code %s\n", file_name);
+        fs_close_alloc(fp);
+        return 0;
+    }
+
+    return size;
```
grgustaf: You're leaking the fp descriptor on the success path, need to fs_close_alloc. This would be a great time to use ZVAL's cleanup trick:

```c
fs_file_t __attribute__ ((__cleanup__(fs_close_alloc)) *fp = fs_open_alloc(...
```

(Or it might pass a `fs_file_t **` to the function in which case you'd have to write a wrapper for fs_close_alloc.) Anyway, that would let you remove all four calls to fs_close_alloc (including the missing one).

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+}
+
+static void find_requires(char *filebuf)
+{
+    char *ptr1 = filebuf;
+    char *ptr2 = NULL;
+    char *filestr = NULL;
+    char *ext_check = NULL;
+    while(ptr1 != NULL){
+
```
grgustaf: 1. add spaces before `(` and after `)`
2. remove blank line

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+        // Find next instance of "require"
+        ptr1 = strstr(ptr1, "require");
+
+        if (ptr1 != NULL) {
+            // Find the text between the two " " after require
+            ptr1 = strchr(ptr1, '"')+1;
```
grgustaf: Spaces on either side of the `+`.

Again, you could skip the 'require' with `ptr1 + 7`. But maybe adding another magic number is worse than saving 7 loop iterations in strchr.

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+            ptr2 = strchr(ptr1, '"');
+            size_t len = ptr2-ptr1;
```
grgustaf: Spaces around the `-`; also `len+1` and `len-3` below.

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+            filestr = (char*)zjs_malloc(len+1);
+            strncpy(filestr, ptr1, len);
+            filestr[len] = '\0';
+
+            // Check that this is a *.js require and not a zephyr require
+            // Also check that the file hasn't already been loaded
+            ext_check = &filestr[len-3];
+
+            if (strcmp(ext_check, ".js") == 0 && !is_loaded(filestr))
+            {
```
grgustaf: Move brace to previous line.

Personally, I use !strcmp() for "equal" or just strcmp() for "not equal" rather than comparing with zero. I find that easier to read; when I see this I have to sit and think about it. But maybe that's just habit.

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+    // Check if the file we've been asked to run requires any JS modules
+    find_requires(file_buffer);
+
+    // If it does, check if they have requires as well, and then eval the code
+    while (req_list) {
+        char *req_file_buffer = NULL;
+        uint file_size = read_file(req_list->file_name, &req_file_buffer);
+
+        if (file_size > 0)
+        {
```
grgustaf: move brace up

src/ashell/jerry-code.c
```diff
@@ -32,7 +32,53 @@ void jerry_port_default_set_log_level(jerry_log_level_t level); /** Inside jerry
...
+            }
+            else {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+static void push_to_list(requires_list_t **list, const char *item)
```
grgustaf: Something like `list_front_insert` might be more clear.

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
     jerry_port_console("\n");
 }
 
+static int read_file(const char *file_name, char **file_buf)
```
grgustaf: I think it would be safer for maintenance to return file_buf and take int *size as an argument instead. That makes it more clear that you're doing a malloc inside, whereas the int is not something you would leak. Also then you can optionally set the size only if they give you a real pointer.

`read_file_to_new_buffer` might make it really clear. :)

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+    find_requires(file_buffer);
+
+    // If it does, check if they have requires as well, and then eval the code
+    while (req_list) {
+        char *req_file_buffer = NULL;
+        uint file_size = read_file(req_list->file_name, &req_file_buffer);
+
+        if (file_size > 0)
+        {
+            find_requires(req_file_buffer);
```
grgustaf: I don't think this is going to be safe? You're in the middle over iterating through req_list but this function can add more stuff to (if I got it, the front of) req_list.

In particular, below when you set `cur = req_list`, that might be a *NEW* item that was just added by `find_requires`. Then you're going to free that list item instead of the one you just worked on.

Also, for dependencies to work, don't you need to eval these *new* requires before the one you're looking at now, or it's going to fail?

I think the solution is to do one pass just finding requires all the way down the rabbit hole. If you ever see a require that already *is* loaded you actually need to rip it out from its current place in the list and move it up to the front so it will be in front of the current item I think. Then somehow make sure you can't get into circular dependencies, or else we ignore it, document it well, and call it user error. And then once you know the right order to read all the files in simply go eval them all in a row in a second pass.

src/ashell/jerry-code.c
```diff
@@ -74,6 +120,100 @@ static void javascript_print_value(const jerry_value_t value)
...
+            javascript_eval_code(req_file_buffer, file_size);
+            zjs_free(req_file_buffer);
+            // Add file to the loaded list
+            push_to_list(&loaded_list, req_list->file_name);
```
grgustaf: If you push to the loaded list *before* the `find_requires` above you might protect yourself from circular dependencies - although maybe what those *should* do is fail horribly. :) Except you might get in an infinite loop here if A requires B and B requires A.

src/ashell/jerry-code.c
```diff
@@ -93,11 +233,11 @@ static void javascript_print_error(jerry_value_t error_value)
     zjs_free(msg);
 }
 
-void javascript_eval_code(const char *source_buffer)
+void javascript_eval_code(const char *source_buffer, size_t size)
 {
     jerry_port_default_set_log_level(JERRY_LOG_LEVEL_TRACE);
     ZVAL ret_val = jerry_eval((jerry_char_t *) source_buffer,
-                              strnlen(source_buffer, MAX_BUFFER_SIZE),
+                              size,
```
grgustaf: These args can move up to the first line I think.

src/ashell/comms-uart.c
```diff
@@ -72,7 +72,7 @@ const char *system_get_prompt()
     return system_prompt;
 }
 
-#define MAX_LINE_LEN 80
+#define MAX_LINE_LEN 90
```
jimmy-huang: Maybe increase it to 120? I find even 90 characters per line is going to likely to run into issues down the road.

src/ashell/jerry-code.c
brianjjones: Hmm this is true.  I was really hoping to avoid having to open files multiple times, but it may be a necessary evil.  Thanks for the feedback.

rwaldron:
Is the intention here to match Node.js module semantics?

src/ashell/comms-uart.c
brianjjones: @jimmy-huang Can't do that currently, 90 seems to be the max before it hangs.

brianjjones:
Perhaps in the long run, for now only require() is supported.

rwaldron:
> Perhaps in the long run, for now only require() is supported.

I don't mean to be a pest, but could you elaborate?

grgustaf:
@rwaldron, we don't really have concrete plans here. We are just enabling module features as we see the need for them or they are requested. Are there are specific aspects of Node modules you think make sense for us or you're expecting to see?

We do actually support module.exports as well. I haven't really looked into how well we match what Node does, myself.

src/ashell/jerry-code.c
brianjjones: Yeah I assume  require      (   "test.js");  is valid in javascript, so +7 won't work in that case.

grgustaf:
I don't imagine this needs to go in for 0.3, right?

brianjjones:
@grgustaf Not required no, its an extra feature not a bug fix.

rwaldron:
1. This doesn't appear to wrap the JS module code in a function, which is used provide a private "module" scope at execution time: https://github.com/nodejs/node/blob/a94a5da78c6fdecf1aef541db237a1c96a9801a3/lib/internal/bootstrap_node.js#L536-L563 . Without that, the module code would be executed in the same scope as the program code—allowing the program code to pave overwrite bindings defined in the module code. 
2. The module resolution rules: 
```
require(X) from module at path Y
1. If X is a core module,
   a. return the core module
   b. STOP
2. If X begins with '/'
   a. set Y to be the filesystem root
3. If X begins with './' or '/' or '../'
   a. LOAD_AS_FILE(Y + X)
   b. LOAD_AS_DIRECTORY(Y + X)
4. LOAD_NODE_MODULES(X, dirname(Y))
5. THROW "not found"

LOAD_AS_FILE(X)
1. If X is a file, load X as JavaScript text.  STOP
2. If X.js is a file, load X.js as JavaScript text.  STOP
3. If X.json is a file, parse X.json to a JavaScript Object.  STOP
4. If X.node is a file, load X.node as binary addon.  STOP

LOAD_INDEX(X)
1. If X/index.js is a file, load X/index.js as JavaScript text.  STOP
2. If X/index.json is a file, parse X/index.json to a JavaScript object. STOP
3. If X/index.node is a file, load X/index.node as binary addon.  STOP

LOAD_AS_DIRECTORY(X)
1. If X/package.json is a file,
   a. Parse X/package.json, and look for "main" field.
   b. let M = X + (json main field)
   c. LOAD_AS_FILE(M)
   d. LOAD_INDEX(M)
2. LOAD_INDEX(X)

LOAD_NODE_MODULES(X, START)
1. let DIRS=NODE_MODULES_PATHS(START)
2. for each DIR in DIRS:
   a. LOAD_AS_FILE(DIR/X)
   b. LOAD_AS_DIRECTORY(DIR/X)

NODE_MODULES_PATHS(START)
1. let PARTS = path split(START)
2. let I = count of PARTS - 1
3. let DIRS = []
4. while I >= 0,
   a. if PARTS[I] = "node_modules" CONTINUE
   b. DIR = path join(PARTS[0 .. I] + "node_modules")
   c. DIRS = DIRS + DIR
   d. let I = I - 1
5. return DIRS
```

Obviously the `.node` rules do not apply, but that still leaves plenty of expected behavior:

```
require(X) from module at path Y
1. If X is a core module,
   a. return the core module
   b. STOP
2. If X begins with '/'
   a. set Y to be the filesystem root
3. If X begins with './' or '/' or '../'
   a. LOAD_AS_FILE(Y + X)
   b. LOAD_AS_DIRECTORY(Y + X)
4. LOAD_NODE_MODULES(X, dirname(Y))
5. THROW "not found"

LOAD_AS_FILE(X)
1. If X is a file, load X as JavaScript text.  STOP
2. If X.js is a file, load X.js as JavaScript text.  STOP
3. If X.json is a file, parse X.json to a JavaScript Object.  STOP

LOAD_INDEX(X)
1. If X/index.js is a file, load X/index.js as JavaScript text.  STOP
2. If X/index.json is a file, parse X/index.json to a JavaScript object. STOP

LOAD_AS_DIRECTORY(X)
1. If X/package.json is a file,
   a. Parse X/package.json, and look for "main" field.
   b. let M = X + (json main field)
   c. LOAD_AS_FILE(M)
   d. LOAD_INDEX(M)
2. LOAD_INDEX(X)

LOAD_NODE_MODULES(X, START)
1. let DIRS=NODE_MODULES_PATHS(START)
2. for each DIR in DIRS:
   a. LOAD_AS_FILE(DIR/X)
   b. LOAD_AS_DIRECTORY(DIR/X)

NODE_MODULES_PATHS(START)
1. let PARTS = path split(START)
2. let I = count of PARTS - 1
3. let DIRS = []
4. while I >= 0,
   a. if PARTS[I] = "node_modules" CONTINUE
   b. DIR = path join(PARTS[0 .. I] + "node_modules")
   c. DIRS = DIRS + DIR
   d. let I = I - 1
5. return DIRS
```

brianjjones:
@rwaldron Correct, at the moment it is assumed the module code is wrapped by the author.  I will certainly look into the feasibility of adding the ability to wrap it for the user, it would improve the user experience.

brianjjones:
@grgustaf were there any other changes I need to make on this?

grgustaf:
Please add a GH issue for doing the module wrapping like @rwaldron suggested, and another one to review and discuss the full Node module rules he pointed to.

I'll take another look now.

src/ashell/jerry-code.c
```diff
@@ -74,6 +134,123 @@ static void javascript_print_value(const jerry_value_t value)
...
+static void add_requires(requires_list_t **list, char *filebuf)
+{
+    // Scan the file and if it contains requires, append them before *list
+    char *ptr1 = filebuf;
+    char *ptr2 = NULL;
+    char *filestr = NULL;
+    char *ext_check = NULL;
+
+    while (ptr1 != NULL) {
+
```
grgustaf: remove blank line

src/ashell/jerry-code.c
```diff
@@ -74,6 +134,123 @@ static void javascript_print_value(const jerry_value_t value)
     jerry_port_console("\n");
 }
 
+static char *read_file(const char *file_name, ssize_t *size)
```
grgustaf: Add alloc to the name or something, to make it clear that it's returning a new buffer. Otherwise, it's a memory leak waiting to happen when Mr. Maintenance comes along.

E.g. `read_alloc_file`

brianjjones: Sure, changed to read_file_alloc

brianjjones:
@grgustaf  Thanks, I've updated with changes and added the tasks to our GH issues list.

grgustaf:
+1

PR #964, opened 2017-04-04
[buffer] Return buffer as out parameter to zjs_buffer_create()

jprestwo:


grgustaf:
+1

PR #966, opened 2017-04-04
[buffer] Update Buffer test to allow out of memory errors

grgustaf:

### DIFF
--- a/tests/test-buffer.js
+++ b/tests/test-buffer.js
@@ -9,14 +9,65 @@ var buff;
 var lens = [[-1, 0],
             [0, 0],
             [8, 8],
-            [8.5, 8],
-            [4294967295, 4294967295],
-            [4294967296, 4294967295]];
-for(var i = 0; i < lens.length; i++) {
+            [8.49, 8],  // now it does rounding
+            [8.5, 9]];
+for (var i = 0; i < lens.length; i++) {
     buff = new Buffer(lens[i][0]);
     assert(buff.length === lens[i][1],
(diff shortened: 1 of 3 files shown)


jimmy-huang:
+1

tests/test-buffer.js
```diff
@@ -9,14 +9,28 @@ var buff;
...
     assert(buff.length === lens[i][1],
-           "The length of Buffer(" + lens[i][0] + ") expected:" + lens[i][1] +
-           " got:" + buff.length);
+           "Creating buffer of length " + lens[i][0] + " expected:" +
+           lens[i][1] + " got:" + buff.length);
+}
+
+// thrown errors are allowed on these lengths
+lens = [[4294967295, 4294967295],
+        [4294967296, 4294967295]];
```
haoxli: I think we'd better to use 1073741823(2^30-1) following Node.js 6.x [buffer.kMaxLength
](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_buffer_kmaxlength), and it's expected to be a valid length for large memory(>1G) or throw error when devices have no enough available memory.

1073741824(2^30) should always throw error for out of memory.

grgustaf: This wouldn't be bad, but it's kind of irrelevant for these small devices we're really targeting. They will always run out of RAM way before this "max length". I guess I can add it under a Linux ifdef so it doesn't add more code to the small device case.

grgustaf:
@jimmy-huang there are non-trivial updates to review. But I've tested with all the buffer tests.

PR #968, opened 2017-04-04
[mac] Fix build issues on Mac

jimmy-huang:
This adds a BOARD=linux job on Travis for Mac OSX builds, it
includes fixes to all the warnings on mac as well as updated
scripts that will work on OSX.   It will take too much time
to build Zephyr targets on Mac OSX as it needs to fetch crosstool-ng
which is a cross compiler and then build it and fetches the arduino
ARC compiler to build the arc.

Also updating the Mac OSX build instructions to the latest.


.travis.yml
```diff
@@ -19,19 +23,33 @@ env:
...
+      env: TARGET="zephyr"
+    - os: linux
+      env: TARGET="extra"
+    - os: linux
+      env: TARGET="linux"
+    - os: linux
+      env: TARGET="ashell"
+    - os: osx
+      osx_image: xcode8.3
+      env: TARGET="linux"
```
grgustaf: I'd like something different here to distinguish which VM is which in the Travis UI. So maybe add OS="osx" to the last one? The variable won't be used by we can at least read that then.

jimmy-huang: tthe os: osx is in the last one already, so the top 4 is os: linux, the 5th one is os: osx, and if you look at the build status page, the osx one has a apple logo instead of penguin

grgustaf:
-1 for now until we see whether his Mac backlog on open source builds is a problem all the time. We don't want to be waiting for hours to approve code.

Makefile.linux
```diff
@@ -56,8 +60,6 @@ LINUX_FLAGS += 	-fno-asynchronous-unwind-tables \
 		-fno-stack-protector \
 		-ffunction-sections \
 		-fdata-sections \
-		-fno-defer-pop \
-		-Wno-pointer-sign \
```
grgustaf: Why these changes?

jimmy-huang: they are not supported in the clang on mac, i don't know if there are equivalents we can replace with, if we do wanna keep these, i'll put a ifdef there

README.md
```diff
@@ -412,6 +415,170 @@ on it, specifically the hardware modules (AIO, I2C, GPIO etc.). There are some
...
+
+Note. There's currently a bug that you'll run into a linker issue using the latest Zephyr (1.7.0),
+we filed a bug upstream to track this, we'll update this document once that is fixed.
+
+### Other targets like FRDM-K64F or possibly other ARM boards on Mac
+This also have limited support currently, the requiremenet here is that you'll
+need to install the GCC ARM Embedded cross compiler [here](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads)
+
+After you download it, set these environment variables:
+
```
grgustaf: Drop blank line to match elsewhere

README.md
```diff
@@ -412,6 +415,170 @@ on it, specifically the hardware modules (AIO, I2C, GPIO etc.). There are some
...
+```bash
+$ export GCCARMEMB_TOOLCHAIN_PATH="~/Downloads/gcc-arm-none-eabi-6-2017-q1-update/"
+$ export ZEPHYR_GCC_VARIANT=gccarmemb
+```
+
+Then you can build it like:
```
grgustaf: like this:

fragments/prj.conf.qemu_x86_osx
```diff
@@ -0,0 +1,2 @@
+CONFIG_RAM_SIZE=192
+CONFIG_ROM_SIZE=256
```
grgustaf: Uh, what?

scripts/trlite
```diff
@@ -298,7 +298,7 @@ function check_jstring_util() {
     CALLCOUNT=$(grep jerry_string_to_char_buffer src/zjs_util.c | wc -l)
 
     # make sure there are exactly two calls to the function
-    if [ "$CALLCOUNT" != "2" ]; then
+    if [ $CALLCOUNT != "2" ]; then
```
grgustaf: Are you sure this is needed? Usually the quotes are sort of optional but kind of safer in case it ever got a result with a space in it, so BKM AFAIK.

src/main.c
```diff
@@ -216,7 +216,7 @@ int main(int argc, char *argv[])
     zjs_obj_add_function(global_obj, native_print_handler, "print");
     zjs_obj_add_function(global_obj, stop_js_handler, "stopJS");
 #ifndef ZJS_SNAPSHOT_BUILD
-    code_eval = jerry_parse(script, len, false);
+    code_eval = jerry_parse((jerry_char_t *)script, len, false);
```
grgustaf: Was this giving a warning?

src/zjs_error.c
```diff
@@ -161,7 +161,7 @@ jerry_value_t zjs_custom_error(const char *name, const char *message)
 jerry_value_t zjs_standard_error(zjs_error_type_t type, const char *message)
 {
     int count = sizeof(error_types) / sizeof(zjs_error_t);
-    if (type >= count || type < 0) {
+    if (type >= count) {
```
grgustaf: Why? This seems safer, even if the enum type is unsigned. (I'm not sure... googling a little it appears gcc makes it unsigned by ISO standard says signed, so I'd stand by this being safer. We don't want to use a negative array index below.)

src/zjs_unit_tests.c
```diff
@@ -101,9 +101,7 @@ static void test_validate_args()
 {
     // create some dummy objects
     ZVAL obj1 = jerry_create_object();
-    ZVAL obj2 = jerry_create_object();
-    ZVAL str1 = jerry_create_string("arthur");
-    ZVAL str2 = jerry_create_string("ford");
+    ZVAL str1 = jerry_create_string((jerry_char_t *)"arthur");
```
grgustaf: More warnings?

src/zjs_util.c
```diff
@@ -309,7 +309,20 @@ void zjs_default_convert_pin(uint32_t orig, int *dev, int *pin) {
...
+        // GCC states that can't use __builtin_clz
+        // if x is 0, because the result is undefined
+        return 0;
+    }
+
+#ifdef __MAC__
+    // on Mac OSX, uint32_t is unsigned int
+    int zeroes = __builtin_clz(num);
+#else
+    // on Linux, uint32_t is unsigned long
```
grgustaf: I think this is only in the broken SDK. So here's where just casting to (unsigned int) probably suffices? (And I guess switch to the _clz instead of _clzl.

I just wrote a sample app and showed that uint32_t is unsigned int on normal Linux, command-line default gcc.

fragments/prj.conf.qemu_x86_osx
jimmy-huang: In QEMU, we set CONFIG_XIP=Y, so I have to drop that, and increase the RAM_SIZE from 80 to 192, or I will run into the ram not fit error.

scripts/trlite
jimmy-huang: This is needed on Mac, because returned value after running wc -l contains some blank spaces before it, so if you double quote it, it became like "      2", and then comparison will fail

src/main.c
jimmy-huang: yeah, on Mac they were giving warnings, clang it's much more strict

src/zjs_error.c
jimmy-huang: again, clang gives a warning that type is never < 0 since it's unsigned, so I had to drop this.

src/zjs_unit_tests.c
jimmy-huang: yep, clang

src/zjs_util.c
jimmy-huang: maybe you can submit another patch for it?

grgustaf:
OK, this is almost ready, as long as it doesn't turn out that Mac has a long backlog every day. So let's wait at least one more day. I guess we could always roll it back if it does.

grgustaf:
@jimmy-huang this is three days in a row that Mac has had a backlog of like 2/3 of the active jobs on Travis, i.e. they're overbooked by ~66%. On Linux, the peak I see is about ~3%. This means we're going to be waiting around for Travis builds to complete on Mac. So I'd rather leave it out of Travis for now. Hopefully this is something they will address.

Can you split apart the build fixes patch and the Travis patch?

scripts/analyze.sh
```diff
@@ -364,7 +364,7 @@ if [ $? -eq 0 ] || check_config_file ZJS_SENSOR; then
             echo "CONFIG_ADC=y" >> $ARCPRJFILE
             # Workaround for the Zephyr issue ZEP-1882: ADC doesn't work with
             # SDK 0.9 due to some compiler optimization, so enable debug mode.
-            if [ $SDK_VERSION = "0.9" ]; then
+            if [[ $SDK_VERSION = "0.9" ]]; then
```
grgustaf: Explain this one again?

jimmy-huang: I am getting one of those unary errors without adding this, seems like it either needs to wrap $SDK_VERSION with "", or add this extra bracket on mac

jimmy-huang:
I reverted the travis changes, and submitted as a separate patch #981

Makefile.linux
jprestwo: They can be removed. The no-defer-pop gets disabled with any optimizations (-O_) and the no-pointer-sign disables warnings for pointer conversions.

grgustaf:
+1

PR #972, opened 2017-04-05
[ashell] All ashell commands check if the file exists

### LINKED ISSUE #799: ashell creates empty files when you make a typo
If you type the wrong file name in a run, parse, or du command at the ashell command line, it will create an empty file. I think this is because the underlying Zephyr FS API creates on open. So instead we should be checking for existence first and giving an error if it doesn't exist.

The 'cat' command *does* check, however its File not found error is missing a newline so it doesn't print nicely, so that could be fixed too.

brianjjones:
Also added newline to all the error strings to keep the terminal
output clean.

Fixes #799

### DIFF
--- a/src/ashell/shell-state.c
+++ b/src/ashell/shell-state.c
@@ -39,17 +39,17 @@ static struct shell_state_config shell = {
     .state_flags = kShellTransferRaw
 };
 
-const char ERROR_NOT_RECOGNIZED[] = "Unknown command\n";
-const char ERROR_NOT_ENOUGH_ARGUMENTS[] = "Not enough arguments\n";
-const char ERROR_FILE_NOT_FOUND[] = "File not found\n";
-const char ERROR_EXCEDEED_SIZE[] = "String too long\n";
+const char ERROR_NOT_RECOGNIZED[] = "Unknown command";
+const char ERROR_NOT_ENOUGH_ARGUMENTS[] = "Not enough arguments";
+const char ERROR_FILE_NOT_FOUND[] = "File not found";
+const char ERROR_EXCEDEED_SIZE[] = "String too long";
 
 const char MSG_FILE_SAVED[] =
      ANSI_FG_GREEN "Saving file. " ANSI_FG_RESTORE
-     "run the 'run' command to see the result\n";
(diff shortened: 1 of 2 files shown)


src/ashell/shell-state.c
```diff
@@ -39,17 +39,17 @@ static struct shell_state_config shell = {
     .state_flags = kShellTransferRaw
 };
 
-const char ERROR_NOT_RECOGNIZED[] = "Unknown command";
-const char ERROR_NOT_ENOUGH_ARGUMENTS[] = "Not enough arguments";
-const char ERROR_FILE_NOT_FOUND[] = "File not found";
-const char ERROR_EXCEDEED_SIZE[] = "String too long";
+const char ERROR_NOT_RECOGNIZED[] = "Unknown command\n";
```
jimmy-huang: you should add the \n to the printf, not in the string here

grgustaf:
Don't put the bug number in the summary line. Put "Fixes #799." in its own paragraph, just before your Signed-off-by.

brianjjones:
@grgustaf updated

src/ashell/shell-state.c
```diff
@@ -88,13 +88,26 @@ int32_t ashell_get_filename_buffer(const char *buf, char *destination)
...
         *destination = '\0';
-        comms_printf(ERROR_NOT_ENOUGH_ARGUMENTS);
+        comms_println(ERROR_NOT_ENOUGH_ARGUMENTS);
         return RET_ERROR;
     }
 
     return arg_len;
 }
 
+static int8_t file_exists(const char *buf, char *filename)
```
grgustaf: This function name isn't great because it also gets the filename. That got me looking into it more.

I think a better way to fix this is to change fs_open_alloc() in file_utils.c. It takes a `mode` argument but mostly ignores it. If it starts with 'r' it should not be creating a new file. That would solve the problem I think.

brianjjones: I foolishly saw fs_open_alloc and assumed it was a zephyr call.  Yeah makes more sense I'll take care of it there.

grgustaf:
+1

PR #974, opened 2017-04-05
[snapshot] Fix possible cause of parser disabled in ashell

### LINKED ISSUE #936: ashell has parser disabled now
Probably some recent Makefile change messed this up and left SNAPSHOT=on or something?

jimmy-huang:
We should not delete the .snapshot.last_build file since make clean
do not rebuild snapshot tool.

Fixes #936

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,6 @@ cleanlocal:
 	@rm -f prj.mdef
 	@rm -f zjs.conf.tmp
 	@rm -f js.tmp
-	@rm -f .snapshot.last_build
 
 # Explicit clean
 .PHONY: clean


grgustaf:
Cleaning up non-source files is what `make clean` should do. So it's right for it to remove that snapshot file. That's why I don't like this solution. Instead, `make clean` should probably blow away the snapshot tool to ensure it gets rebuilt.

I think we also had a better idea where we would maybe build one copy of JerryScript for the snapshot case and one for the non-snapshot case, naming the libraries differently. Then they can coexist and we won't have to rebuild as often and will not have these kinds of problems where the wrong one is being used.

jimmy-huang:
Yeah, I think keeping two copies of it is a good idea, so I'll drop this and create an issue for that.

PR #980, opened 2017-04-06
[net config] Added net config module

jprestwo:
- OCF and DGRAM can now use the same network/BLE config API's


src/zjs_net_config.c
```diff
@@ -1,6 +1,49 @@
...
+static uint8_t ble_enabled = 0;
+#endif
 
-#include "zjs_ocf_ble.h"
+void zjs_net_config(void)
+{
+#if defined(CONFIG_NET_L2_BLUETOOTH)
+    if (!ble_enabled) {
+        zjs_init_ble_address();
+        if (bt_enable(NULL)) {
```
jimmy-huang: what is the return value of bt_enable here? if it not boolean maybe check if != error?

grgustaf: looks right, zero is success

grgustaf:
+1

pfalcon:
I'm as usual late with review...

This looks definitely good! For a moment I thought that maybe this is a JS-level module which allows to set local IP addresses, etc., but I see it's more an internal C module to consolidate related functions. Well, that's already a good step forwards, and I hope we'll get to JS-level module too eventually (one question is what to model its API on).

PR #981, opened 2017-04-06
[travis] Add Mac OSX build to travis CI

jimmy-huang:


jimmy-huang:
this will depend on #968

grgustaf:
Every time I've looked at the day's history on http://traviscistatus.org, I see a significant backlog of OSX builds during our daytime, so I don't think we want to introduce this unless Mac problems become more common or more critical; it will slow down our code merging too much.

Issue #924, opened 2017-03-28
USAGE FAULT when running HelloWorld.js


makerdiary:
Hi guys,
I am trying to running zephyr.js on nRF52832. I can compile the sample  `HelloWorld.js` now, but when I run it on the hardware, a fault came out. Any ideas about this error?
```
***** USAGE FAULT *****
  Executing thread ID (thread): 0x20006aec
  Faulting instruction address:  0x6bfa
  Unaligned memory access
Fatal fault in essential thread! Spinning...
```

Thanks for your reply.
Zelin

jprestwo:
We have not tested on the nRF52832 but we are always trying to get new Zephyr boards supported. How did you build ZJS? (e.g. make BOARD=nRF52832). Have you tried building and running a Zephyr sample on the nRF52832?

jprestwo:
I tried running hello world on the nRF52-PCA10040 and was able to run it. I followed directions from the Zephyr board support site to setup (download tools):
https://www.zephyrproject.org/doc/boards/arm/nrf52_pca10040/doc/nrf52_pca10040.html

And then built by running:
{code}
make BOARD=nrf52_pca10040 JS=samples/HelloWorld.js
{code}

Then flashed:
{code}
nrfjprog --eraseall -f nrf52
nrfjprog --program outdir/nrf52_pca10040/zephyr.hex -f nrf52
{code}

makerdiary:
@jprestwo Thanks for your quick response.
I just tried to build it for nrf52_pca10040, the same fault came out.
I can build the zephyr samples, and they are running fine.

jprestwo:
Just to make sure we are talking about the same board, is this the one you have:
http://www.digikey.com/product-detail/en/NRF51-DK/1490-1038-ND/5022449?WT.mc_id=IQ_7595_G_pla5022449&wt.srch=1&wt.medium=cpc&WT.srch=1&gclid=COz7t_CV_NICFUlNfgodiJYMqQ

makerdiary:
@jprestwo I am using [nRF52 DK](http://www.nordicsemi.com/eng/Products/Bluetooth-low-energy/nRF52-DK) not nRF51 DK.

grgustaf:
That's really what @jprestwo meant; that's what we have here too, nrF52 DK.

grgustaf:
[...]

PR #985, opened 2017-04-06
[util] Changed malloc to allocate in chunks of 4 bytes

jprestwo:


grgustaf:
Note: the reason we're considering this is in case it helps with #924; otherwise no point I think.

pfalcon:
Weird. I believe C standard *requires* pointer returned by malloc to have a natural alignment for the platform. You guys use Zephyr's (i.e. newlib's) malloc(), do you want to say it's not compliant?

pfalcon:
Not a C standard, but Linux man page for reference:

> The  malloc()  and calloc() functions return a pointer to the allocated memory that is suitably aligned for any kind of variable.

grgustaf:
OK, I checked and on A101 at least it always returns 8-byte aligned pointers. So we can probably close this.

PR #986, opened 2017-04-06
[build] Maintain both snapshot and parser versions of JrS library

grgustaf:
Instead of replacing the same libjerry-core.a with a different version
depending on snapshot vs. parser build, maintain both with different
filenames. Then we don't have to rebuild JerryScript just because we
switched between those two types of builds, and don't need the snapshot
last_build file and related dangers of inconsistency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,9 +144,16 @@ rom_report: zephyr
 					SNAPSHOT=$(SNAPSHOT) \
 					rom_report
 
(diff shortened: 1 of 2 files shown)


jimmy-huang:
+1

jprestwo:
+1 if @grgustaf tested it :)

grgustaf:
I did, including with Jimmy's steps from #936, so I think we're good.

PR #992, opened 2017-04-07
[timers] Only remove interval timers in delete_timer()

jprestwo:
Timeouts automatically get removed since they are a "once"
callback.

### DIFF
--- a/src/zjs_timers.c
+++ b/src/zjs_timers.c
@@ -115,7 +115,10 @@ static bool delete_timer(int32_t id)
             for (int i = 0; i < tm->argc; ++i) {
                 jerry_release_value(tm->argv[i]);
             }
-            zjs_remove_callback(tm->callback_id);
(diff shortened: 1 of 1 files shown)


grgustaf:
This looks like it might need more thought, because delete_timer takes what appears to be a callback ID, so if that no longer exists it could have been reassigned etc. We also already have this logic with "completed" flag that was probably trying to solve something along these lines.

jprestwo:
@grgustaf, there are just two cases:

1. Interval timer: `tm->repeat = 1`, so this will always get it in the new if case. This timer/callback ID *should* still exist at this point because (a) this is NOT a "once" timer, and (b) there is no where else in timers that calls remove callback besides delete_timer().

2. Timeout: `tm->repeat = 0`, so this will never hit the new if case. Since this is a "once" timer, it will get removed automatically; the only thing that needs to get cleaned up is the timer state information.

grgustaf:
+1

PR #993, opened 2017-04-07
[callbacks] Fix memory leak in callbacks (debug build only)

jprestwo:

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -98,7 +98,7 @@ static int zjs_ringbuf_last_error = 0;
 #ifdef DEBUG_BUILD
 static void set_info_string(char *str, const char *file, const char *func)
 {
-    char *i = (char*)(file + strlen(file));
+    const char *i = file + strlen(file);
     while (i != file) {
         if (*i == '/') {
             i++;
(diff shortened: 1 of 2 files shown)


jprestwo:
@grgustaf, per our discussion, don't merge this just yet.

src/zjs_callbacks.c
```diff
@@ -95,16 +96,23 @@ static int zjs_ringbuf_error_max = 0;
 static int zjs_ringbuf_last_error = 0;
 
 #ifdef DEBUG_BUILD
-static void set_info_string(char **str, const char *file, const char *func)
+static void set_info_string(char *str, const char *file, const char *func)
 {
-    uint32_t file_len = strlen(file);
+    char *i = (char*)(file + strlen(file));
```
grgustaf: Just for consistency do (char *)

grgustaf: Wait, actually this should be const char *i on the left and you shouldn't need a cast on the right.

grgustaf: I confused myself earlier but yes as I said I think you should be using const char *i on the left, and not casting on the right.

Better yet come up with a variable name :D.

src/zjs_callbacks.c
```diff
@@ -95,16 +96,23 @@ static int zjs_ringbuf_error_max = 0;
...
-    *str = zjs_malloc(file_len + func_len + 4);
-    if (!(*str)) {
-        ERR_PRINT("failed to create file/func info string\n");
-        return;
+    uint32_t total = file_len + func_len + 4;
+    if (total > MAX_CALLER_CREATOR_LEN) {
+        total = MAX_CALLER_CREATOR_LEN;
     }
-    snprintf(*str, file_len + func_len + 4, "%s:%s()", file, func);
+    snprintf(str, file_len + func_len + 4, "%s:%s()", i, func);
```
grgustaf: You meant to use `total` here; however you can just pass MAX_CALLER_CREATOR_LEN all the time instead, and not bother calculating `total`.

grgustaf:
+1

PR #994, opened 2017-04-07
Make these commands more end-user friendly (easier to copy)

rwaldron:
Signed-off-by: Rick Waldron <waldron.rick@gmail.com>


jimmy-huang:
I like the change, but then we need to update this everywhere else, not just for mac

grgustaf:
+1, makes sense... we can do that, @jimmy-huang

PR #995, opened 2017-04-07
Adds CT_CC_GCC_STATIC_LIBSTDCXX=n to custom .config settings

rwaldron:
Without this, `ct-ng build` will fail

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>


jimmy-huang:
+1

PR #996, opened 2017-04-07
[utils] Add ZJS_DECL_FUNC and related macros

grgustaf:
These simplify the boilerplate of declaring JavaScript to C binding
functions. The one downside is this hides the variable names that
are declared, so you're expected to know what those are.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -49,19 +49,13 @@ const char script_jscode[] = {
 #endif
 
 // native eval handler
-static jerry_value_t native_eval_handler(const jerry_value_t function_obj,
-                                         const jerry_value_t this,
-                                         const jerry_value_t argv[],
-                                         const jerry_length_t argc)
+static ZJS_DECL_FUNC(native_eval_handler)
 {
     return zjs_error("eval not supported");
 }
 
 // native print handler
-static jerry_value_t native_print_handler(const jerry_value_t function_obj,
-                                          const jerry_value_t this,
-                                          const jerry_value_t argv[],
-                                          const jerry_length_t argc)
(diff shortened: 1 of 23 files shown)


grgustaf:
@pfalcon and @martijnthe, I'd value your opinions on whether this is a good idea or not.

martijnthe:
I'm used to having something very similar.
Not sure if it makes sense to include `_API_` in the name. Not all of the uses directly map to a JS API.
I'd hint at "external function" instead –  that's the term that JrS uses for these things.

grgustaf:
Yeah, I'm trying to balance "short and sweet" with "accurate". Updated with ZJS_JERRY_FUNC based on IRC discussion.

poussa:
I'd like to drop the JERRY from the name.

How about ZJS_API ?

grgustaf:
I don't mind that, but it *is* a very JerryScript-centric thing it's doing... replacing four copies of `jerry_value_t` and one `jerry_length_`.

Example:
```diff
-static jerry_value_t zjs_ble_primary_service(const jerry_value_t function_obj,
-                                             const jerry_value_t this,
-                                             const jerry_value_t argv[],
-                                             const jerry_length_t argc)
+static ZJS_JERRY_FUNC(zjs_ble_primary_service)
```

Also @martijnthe pointed out it's not a strict correspondence w/ actual APIs, some of these are internal helper functions for implementing. Some other possibilities could be `ZJS_DECL_FUNC`, `ZJS_DECL_BINDING`. It's kind of good for it to be a bit vague so that you know you may need to look it up to understand it. :)

jimmy-huang:
+1

PR #999, opened 2017-04-08
[zjsbanner] Add our own banner script

grgustaf:
A good, fun Friday task. Now no one will be without lovely banner text,
not even Mac users or people who don't install sysvbanner.


jimmy-huang:
+1

PR #1000, opened 2017-04-08
[ashell] Boot JavaScript file specified by boot.cfg

brianjjones:
Now ashell will run whatever JavaScript file is specified
in boot.cfg. Users can set which file to run on boot using
'cfg' in the ashell menu. ashell operates as usual even if
it starts up a program.

### DIFF
--- a/docs/ashell.md
+++ b/docs/ashell.md
@@ -86,6 +86,7 @@ Commands list:
      run [FILE] Runs the JavaScript program in the file
    parse [FILE] Check if the JS syntax is correct
     stop Stops current JavaScript execution
+ bootcfg [FILE] Set the file that should run at boot
       ls [FILE] List directory contents or file stat
      cat [FILE] Print the file contents of a file
       du [FILE] Estimate file space usage
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -522,8 +522,7 @@ void acm()
     k_fifo_init(&avail_queue);
 
     // Wait until the main loop is running before starting the javascript
-    if (k_sem_take(&mainloop_sem, K_FOREVER) == 0)
-    {
+    if (k_sem_take(&mainloop_sem, K_FOREVER) == 0) {
         ashell_run_boot_cfg();
     }
     else {
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -538,7 +538,7 @@ void acm()
             break;
         // Sleep is needed to allow the javascript in boot.cfg to run
         // while we wait for the connection.
-        k_sleep(2000);
+        k_sleep(1000);
     }
 
     /* 1000 msec = 1 sec */
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, Intel Corporation.
+// Copyright (c) 2016-2017, Intel Corporation.
 
 /**
  * @file
@@ -61,8 +61,6 @@
 extern void __stdout_hook_install(int(*fn)(int));
 
 static const char banner[] = "Zephyr.js DEV MODE " __DATE__ " " __TIME__ "\r\n";
-
-const char filename[] = "jerry.js";
 
 // Jerryscript in green color
 
@@ -523,6 +521,15 @@ void acm()
     k_fifo_init(&data_queue);
     k_fifo_init(&avail_queue);
 
+    // Wait until the main loop is running before starting the javascript
+    if (k_sem_take(&mainloop_sem, K_FOREVER) == 0)
+    {
+        ashell_run_boot_cfg();
+    }
+    else {
+        ERR_PRINT("Main loop failed to start\n");
+    }
+
 #ifdef CONFIG_UART_LINE_CTRL
     uint32_t dtr = 0;
 
@@ -530,6 +537,9 @@ void acm()
         uart_line_ctrl_get(dev_upload, LINE_CTRL_DTR, &dtr);
         if (dtr)
             break;
+        // Sleep is needed to allow the javascript in boot.cfg to run
+        // while we wait for the connection.
+        k_sleep(2000);
     }
 
     /* 1000 msec = 1 sec */
--- a/src/ashell/file-utils.c
+++ b/src/ashell/file-utils.c
@@ -79,7 +79,7 @@ int fs_close_alloc(fs_file_t *fp)
     return res;
 }
 
-bool fs_valid_filename_size(char *filename)
+bool fs_valid_filename(char *filename)
 {
     if (filename == NULL) {
         printf("No filename given\n");
@@ -92,33 +92,24 @@ bool fs_valid_filename_size(char *filename)
     ssize_t extlen;
     ssize_t size = strlen(filename);
 
-    if (fs_exist(filename)) {
-        ptr1 = strchr(ptr1, '.');
+    ptr1 = strchr(ptr1, '.');
 
+    if (ptr1 != NULL) {
+        // Check there aren't multiple periods
+        ptr2 = strchr(ptr1 + 1, '.');
 
-        if (ptr1 != NULL) {
-            // Check there aren't multiple periods
-            ptr2 = strchr(ptr1 + 1, '.');
-
-            if (ptr2 != NULL) {
-                printf("Invalid file extention format\n");
-                return false;
-            }
-
-            namelen = ptr1 - filename;
-            extlen = size - namelen - 1;
+        if (ptr2 != NULL) {
+            printf("Invalid file extension format\n");
+            return false;
         }
-        else
-        {
-            // Filename has no extension
-            namelen = strlen(filename);
-            extlen = 0;
-        }
+
+        namelen = ptr1 - filename;
+        extlen = size - namelen - 1;
     }
     else {
-        // Don't let the cfg be set to a file that doesn't exist
-        printf("File passed to cfg doesn't exist\n");
-        return false;
+        // Filename has no extension
+        namelen = strlen(filename);
+        extlen = 0;
     }
 
     if (namelen == 0) {
@@ -131,7 +122,7 @@ bool fs_valid_filename_size(char *filename)
     }
 
     if (extlen > 3) {
-        printf("File extention is longer than 3 characters\n");
+        printf("File extension is longer than 3 characters\n");
         return false;
     }
 
--- a/src/ashell/file-utils.c
+++ b/src/ashell/file-utils.c
@@ -108,7 +108,7 @@ bool fs_valid_filename(char *filename)
     }
     else {
         // Filename has no extension
-        namelen = strlen(filename);
+        namelen = size;
         extlen = 0;
     }
 
@@ -116,15 +116,9 @@ bool fs_valid_filename(char *filename)
         printf("Filename length is zero\n");
         return false;
     }
-    else if (namelen > 8) {
-        printf("Filename is longer than 8 characters\n");
+    else if (namelen > 8 || extlen > 3) {
+        printf("Filename must be 8.3 format\n");
         return false;
     }
-
-    if (extlen > 3) {
-        printf("File extension is longer than 3 characters\n");
-        return false;
-    }
-
     return true;
 }
--- a/src/ashell/file-utils.c
+++ b/src/ashell/file-utils.c
@@ -78,3 +78,62 @@ int fs_close_alloc(fs_file_t *fp)
     zjs_free(fp);
     return res;
 }
+
+bool fs_valid_filename_size(char *filename)
+{
+    if (filename == NULL) {
+        printf("No filename given\n");
+        return false;
+    }
+
+    char *ptr1 = filename;
+    char *ptr2 = NULL;
+    ssize_t namelen;
+    ssize_t extlen;
+    ssize_t size = strlen(filename);
+
+    if (fs_exist(filename)) {
+        ptr1 = strchr(ptr1, '.');
+
+
+        if (ptr1 != NULL) {
+            // Check there aren't multiple periods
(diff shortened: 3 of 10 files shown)


src/ashell/comms-uart.c
```diff
@@ -523,13 +521,25 @@ void acm()
     k_fifo_init(&data_queue);
     k_fifo_init(&avail_queue);
 
+    // Wait until the main loop is running before starting the javascript
+    if (k_sem_take(&mainloop_sem, K_FOREVER) == 0)
+    {
```
grgustaf: move brace up

src/ashell/comms-uart.c
```diff
@@ -523,13 +521,25 @@ void acm()
...
 #ifdef CONFIG_UART_LINE_CTRL
     uint32_t dtr = 0;
 
     while (1) {
         uart_line_ctrl_get(dev_upload, LINE_CTRL_DTR, &dtr);
         if (dtr)
             break;
+        // Sleep is needed to allow the javascript in boot.cfg to run
+        // while we wait for the connection.
+        k_sleep(2000);
```
grgustaf: Explain this more. What is the 2s for? You expect the JS script to finish in that time? Why is it in a while (1) loop?

src/ashell/file-utils.c
```diff
@@ -78,3 +78,62 @@ int fs_close_alloc(fs_file_t *fp)
...
+    char *ptr1 = filename;
+    char *ptr2 = NULL;
+    ssize_t namelen;
+    ssize_t extlen;
+    ssize_t size = strlen(filename);
+
+    if (fs_exist(filename)) {
+        ptr1 = strchr(ptr1, '.');
+
+
```
grgustaf: remove extra line

src/ashell/file-utils.c
```diff
@@ -78,3 +78,62 @@ int fs_close_alloc(fs_file_t *fp)
...
+            if (ptr2 != NULL) {
+                printf("Invalid file extention format\n");
+                return false;
+            }
+
+            namelen = ptr1 - filename;
+            extlen = size - namelen - 1;
+        }
+        else
+        {
```
grgustaf: move up

src/ashell/file-utils.c
```diff
@@ -78,3 +78,62 @@ int fs_close_alloc(fs_file_t *fp)
...
+    if (fs_exist(filename)) {
+        ptr1 = strchr(ptr1, '.');
+
+
+        if (ptr1 != NULL) {
+            // Check there aren't multiple periods
+            ptr2 = strchr(ptr1 + 1, '.');
+
+            if (ptr2 != NULL) {
+                printf("Invalid file extention format\n");
```
grgustaf: extension

src/ashell/file-utils.c
```diff
@@ -78,3 +78,62 @@ int fs_close_alloc(fs_file_t *fp)
...
+        printf("Filename length is zero\n");
+        return false;
+    }
+    else if (namelen > 8) {
+        printf("Filename is longer than 8 characters\n");
+        return false;
+    }
+
+    if (extlen > 3) {
+        printf("File extention is longer than 3 characters\n");
```
grgustaf: extension

grgustaf: These cases probably can't actually happen because you already know the file exists?

src/ashell/shell-state.c
```diff
@@ -578,6 +576,31 @@ int32_t ashell_check_control(const char *buf, uint32_t len)
...
 }
 
+int32_t ashell_set_bootcfg(char *buf)
+{
+    if (!fs_valid_filename_size(buf)) {
+        return RET_ERROR;
+    }
+
+    fs_file_t *file = fs_open_alloc("boot.cfg", "w+");
+
```
grgustaf: remove this blank line; ties the return value validation visually to getting the value

src/ashell/shell-state.c
```diff
@@ -578,6 +576,31 @@ int32_t ashell_check_control(const char *buf, uint32_t len)
...
+    if (!file) {
+        comms_println("Failed to create boot.cfg file");
+        return RET_ERROR;
+    }
+
+    ssize_t written = fs_write(file, buf, strlen(buf));
+
```
grgustaf: and this

src/ashell/shell-state.c
```diff
@@ -622,6 +646,37 @@ int32_t ashell_help(char *buf)
     return RET_OK;
 }
 
+void ashell_run_boot_cfg() {
+
```
grgustaf: remove blank line, but move the brace down

src/ashell/shell-state.c
```diff
@@ -622,6 +646,37 @@ int32_t ashell_help(char *buf)
...
+        fs_close_alloc(file);
+        return;
+    }
+
+    char data[size+1];
+    count = fs_read(file, data, size);
+    data[size] = '\0';
+    ashell_run_javascript(data);
+    fs_close(file);
+    return ;
```
grgustaf: remove space, but actually just remove the return since it's not needed with now return value

src/ashell/comms-uart.c
brianjjones: No.  This while loop is waiting for the WebUSB or ACM connection, hence the while(1).  The sleep is so that it doesn't eat all the cycles, preventing the javascript from running.

src/ashell/file-utils.c
brianjjones: Yeah looks like I got started thinking I needed to verify it was a valid file format, then realized all I needed to do was check if the file existed.  I'll clean this up

brianjjones: I've renamed it to fs_valid_filename and now use it for when creating a file or renaming it.  cfg simply checks the file being pointed stored in boot.cfg exists.

src/ashell/shell-state.c
```diff
@@ -589,6 +617,7 @@ static const struct ashell_cmd commands[] =
     ASHELL_COMMAND("run",   "[FILE] Runs the JavaScript program in the file" ,ashell_run_javascript),
     ASHELL_COMMAND("parse", "[FILE] Check if the JS syntax is correct"       ,ashell_parse_javascript),
     ASHELL_COMMAND("stop",  "Stops current JavaScript execution"             ,ashell_stop_javascript),
+    ASHELL_COMMAND("cfg",   "[FILE] Set the file that should run at boot"    ,ashell_set_bootcfg),
```
jimmy-huang: should we rename this to something easier to understand? like "boot", "autorun"?, cfg just not easy to understand.

src/ashell/file-utils.c
```diff
@@ -78,3 +78,53 @@ int fs_close_alloc(fs_file_t *fp)
...
+        return false;
+    }
+    else if (namelen > 8) {
+        printf("Filename is longer than 8 characters\n");
+        return false;
+    }
+
+    if (extlen > 3) {
+        printf("File extension is longer than 3 characters\n");
+        return false;
```
rwaldron: Would this ever be used in fs operations that might be expected to open `foo.json`?

brianjjones: This is a limitation of Zephyr's file system API, file extensions can't be longer than 3 characters.  If you try and create a file with a longer extension, it simply doesn't get created. I'll look into seeing if we can't get that expanded.  If not perhaps I can just write a simple conversion that takes all .json files and renames them to .jsn

src/ashell/shell-state.c
brianjjones: I'm open to suggestions, I was just trying to keep the command names short. How about bootcfg?

src/ashell/shell-state.c
```diff
@@ -578,17 +582,42 @@ int32_t ashell_check_control(const char *buf, uint32_t len)
     return RET_OK;
 }
 
+int32_t ashell_set_bootcfg(char *buf)
+{
+    if (!fs_exist(buf)) {
+        comms_println("File passed to cfg doesn't exist\n");
```
grgustaf: This doesn't have to be a concern right now but we need to be careful about how many strings we're adding; these are all taking away ROM space. In this function about a 100 bytes from strings alone. Instead, all "file not found" errors like this across ashell should use the same string; they don't need to mention the command that called it - the user already knows that.

The two rare errors below could be error numbers instead of big strings, eventually.

src/ashell/shell-state.c
```diff
@@ -578,17 +582,42 @@ int32_t ashell_check_control(const char *buf, uint32_t len)
...
 {
     ASHELL_COMMAND("help",  "This help", ashell_help),
-    ASHELL_COMMAND("eval",  "Evaluate JavaScript in real time"                ,ashell_js_immediate_mode),
+    ASHELL_COMMAND("eval",  "Evaluate JavaScript in real time"               ,ashell_js_immediate_mode),
     ASHELL_COMMAND("clear", "Clear the terminal screen"                      ,ashell_clear),
     ASHELL_COMMAND("load",  "[FILE] Saves the input text into a file"        ,ashell_read_data),
     ASHELL_COMMAND("run",   "[FILE] Runs the JavaScript program in the file" ,ashell_run_javascript),
     ASHELL_COMMAND("parse", "[FILE] Check if the JS syntax is correct"       ,ashell_parse_javascript),
     ASHELL_COMMAND("stop",  "Stops current JavaScript execution"             ,ashell_stop_javascript),
+    ASHELL_COMMAND("bootcfg", "[FILE] Set the file that should run at boot"  ,ashell_set_bootcfg),
```
grgustaf: Not a required change but actually 'boot' would be fine and would fit here better, not requiring a bunch of indent adjustment.

src/ashell/shell-state.c
```diff
@@ -622,6 +651,36 @@ int32_t ashell_help(char *buf)
...
+    else if (size > 12) {
+        fs_close_alloc(file);
+        return;
+    }
+
+    char data[size+1];
+    count = fs_read(file, data, size);
+    data[size] = '\0';
+    ashell_run_javascript(data);
+    fs_close(file);
```
grgustaf: You're leaking the file pointer block here because you didn't call fs_close_alloc.

grgustaf: I think it would be clearer to clean up all in one place:

```c
if (size > 0 && size <= 12) {
    // handle good case
}
// clean up in all cases
fs_close_alloc(file);
```

src/ashell/file-utils.c
```diff
@@ -78,3 +78,53 @@ int fs_close_alloc(fs_file_t *fp)
...
+            printf("Invalid file extension format\n");
+            return false;
+        }
+
+        namelen = ptr1 - filename;
+        extlen = size - namelen - 1;
+    }
+    else {
+        // Filename has no extension
+        namelen = strlen(filename);
```
grgustaf: You already got this - size.

src/ashell/file-utils.c
```diff
@@ -78,3 +78,53 @@ int fs_close_alloc(fs_file_t *fp)
...
+        printf("Filename length is zero\n");
+        return false;
+    }
+    else if (namelen > 8) {
+        printf("Filename is longer than 8 characters\n");
+        return false;
+    }
+
+    if (extlen > 3) {
+        printf("File extension is longer than 3 characters\n");
```
grgustaf: Here's another opportunity to economize on errors. How about `if (namelen > 8 || extlen < 3) { printf("Filename must be 8.3 format"); }`

docs/ashell.md
```diff
@@ -86,6 +86,7 @@ Commands list:
      run [FILE] Runs the JavaScript program in the file
    parse [FILE] Check if the JS syntax is correct
     stop Stops current JavaScript execution
+ bootcfg [FILE] Set the file that should run at boot
```
zolkis: *just thinking* -- `bootcfg` needs a better name, we don't have autocomplete. Even `boot` would suffice since the text tells it's run at next boot. Or `default` might also work (default script is the one that is run at next boot).

src/ashell/shell-state.c
zolkis: Good point, you could have globals for common error strings.

brianjjones: I could leave them out and push them to the IDE instead.  Granted, the command line ashell will still fail quietly.

src/ashell/shell-state.c
brianjjones: No problem, changing.

pfalcon: YMMV, but for me, "boot" sounds like a verb, as in: boot now.

src/ashell/shell-state.c
brianjjones: I just also realized that the uart won't be up by that time.  So literally no reason to print an error because it won't show up anyway.  Removing the prints.

docs/ashell.md
brianjjones: boot it is then! Shorter is better.

src/ashell/shell-state.c
```diff
@@ -437,25 +435,26 @@ int32_t ashell_raw_capture(const char *buf, uint32_t len)
 
 int32_t ashell_read_data(char *buf)
 {
-    if (!fs_valid_filename(buf)) {
-        return RET_ERROR;
+    if (shell.state_flags & kShellTransferIhex) {
```
brianjjones: ihex calls load without passing any a *buf, so check it first

src/ashell/shell-state.c
```diff
@@ -655,30 +652,20 @@ void ashell_run_boot_cfg()
...
     ssize_t size = fs_size(file);
-    if (size == 0) {
-        printk("Empty boot.cfg file");
-        fs_close_alloc(file);
-        return;
-    }
-    else if (size > 12) {
-        fs_close_alloc(file);
-        return;
+    if (size > 0 && size <= 12) {
```
brianjjones: This prints won't be seen because the uart isn't up yet, so no reason to keep them.

grgustaf:
+1

PR #1001, opened 2017-04-10
[glcd] Drop unsupported Grove LCD input state API

jimmy-huang:
The Zephyr Grove LCD driver doesn't support text flow direction
set using LCD_IS_ENTRY_RIGHT, GLCD_IS_SHIFT_INCREMENT flags.  So
there's no need to expose the setInputState and getInputState
APIs.  Removing these until support is added in upstream.


src/zjs_grove_lcd.c
```diff
@@ -255,88 +259,68 @@ jerry_value_t zjs_grove_lcd_init()
     jerry_value_t glcd_obj = jerry_create_object();
     zjs_obj_add_function(glcd_obj, zjs_glcd_init, "init");
 
-    // create object properties
-    jerry_value_t val;
-
     // function flags
-    val = jerry_create_number(GLCD_FS_8BIT_MODE);
+    ZVAL_MUTABLE val = jerry_create_number(GLCD_FS_8BIT_MODE);
```
grgustaf: No, this leaks all those values. This is why you shouldn't use MUTABLE. They all still have to be released manually.

grgustaf:
+1

PR #1003, opened 2017-04-10
[headers] Fix header issues and checkheaders script; add to trlite

grgustaf:
We'll see how this goes; the script gets made if you use a symbol in a
header file without first including a header that defines it. This may
work because you have already included that dependency above in the C
file where this header is used, but it's not a great practice.


jimmy-huang:
+1

PR #1005, opened 2017-04-11
[trlite] Run more tests on Linux to catch buffer issues, etc.

grgustaf:
Running these buffer tests would have prevented two buffer issues from
slipping past us last week I think.

Also added several other tests that work on Linux.


jprestwo:
+1

PR #1013, opened 2017-04-11
[fs] Add fs_ to functions in keeping with other modules

grgustaf:
Names like zjs_open and zjs_read were too generic.

Also, remove one unnecessary test for a null string already known to
be non-null.


jimmy-huang:
+1

PR #1014, opened 2017-04-12
Buffer validation

grgustaf:


grgustaf:
I tested all modules affected by the second patch except dgram.

grgustaf:
Also tested with UDP6 on A101, so this should be solid.

grgustaf:
Added two tweaks: moved magic number to first four bytes of private struct, so it should be guaranteed to be safe to read... even if the user passes us a different JS object w/ a different handle type, because every handle has at least four bytes.

Second, I also check that the prototype of the given object matches the one I set.

jimmy-huang:
+1

PR #1016, opened 2017-04-12
[build] ROM=144 default makes more sense to minimize flash times

grgustaf:
If you don't like this, you can export ROM=256 or whatever you like in
your shell. But whenver you test small samples that can fit in the
default 144 it makes more sense to do that because it will flash
faster (and also wear out the flash on the device less quickly).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ OS := $(shell uname)
 
 BOARD ?= arduino_101
 RAM ?= 55
-ROM ?= 256
+ROM ?= 144
 
 # Dump memory information: on = print allocs, full = print allocs + dump pools
 TRACE ?= off


jimmy-huang:
default of 144 make sense, but it could also check if certain components are enabled that will be definitely more than 144kb, then default to 256, like BLE, ASHELL, etc, just an idea

grgustaf:
That's a lot more than a one-line patch though. ;)

I think if we try to do that we need to do it well; you don't want to just have it be weird and random when it guesses right and when it doesn't. So I think it's a non-trivial effort. If you prefer the long flash time to ever failing you can just export ROM to a higher number that works for you.

jimmy-huang:
Yeah, I am okay with the patch, +1

Issue #334, opened 2016-10-27
[BLE] Add argument for ble.disconnect()


cuiyanx:
With one client, the function ble.disconnect() and ble.disconnect(clientAddress) are perfectly working, but with two or more client, it is difference between ble.disconnect() and ble.disconnect(clientAddress).
The function ble.disconnect(clientAddress) is more directional. So, I suggest Adding argument and description for ble.disconnect().

kenchris:
@zolkis ^

grgustaf:
Can someone clarify the issue here? I'm not sure what priority to assign.

cuiyanx:
fixed and closed

PR #1017, opened 2017-04-13
[ble] Support multiple client connections (only up to 4 due to memory constraints)

### LINKED ISSUE #984: BLE needs to support multiple client connections
Currently it only allows one client to connect to it, and once the first client connects, the second client won't be able to connect.

jimmy-huang:
Refactored BLE module to handle multiple connections,
Zephyr default to only 1 client connection, so added a config
to overwrite it to support up to 4 clients.

Also enabled CONFIG_BLUETOOTH_DEBUG_LOG in debug mode to get more
more useful output.

Fixes #984, #334

### DIFF
--- a/src/zjs_ble.c
+++ b/src/zjs_ble.c
@@ -514,21 +514,21 @@ static void zjs_ble_blvl_ccc_cfg_changed(const struct bt_gatt_attr *attr, uint16
 
 static void zjs_ble_connected_c_callback(void *handle, const void *argv)
 {
-    ble_handle_t * h = (ble_handle_t *)handle;
+    ble_handle_t *h = (ble_handle_t *)handle;
     if (!h) {
         ERR_PRINT("handle not found\n");
         return;
     }
 
-    char *addr = (char *)argv;
-    ZVAL arg = jerry_create_string((jerry_char_t *)addr);
+    const char *addr = (const char *)argv;
+    ZVAL arg = jerry_create_string((const jerry_char_t *)addr);
     zjs_trigger_event(h->ble_obj, "accept", &arg, 1, NULL, NULL);
     DBG_PRINT("BLE event: connected, addr %s\n", addr);
 }
 
 static void zjs_ble_disconnected_c_callback(void *handle, const void *argv)
 {
-    ble_handle_t * h = (ble_handle_t *)handle;
+    ble_handle_t *h = (ble_handle_t *)handle;
     if (!h) {
         ERR_PRINT("handle not found\n");
         return;
@@ -549,7 +549,7 @@ static void zjs_ble_connected(struct bt_conn *conn, uint8_t err)
         DBG_PRINT("new connection\n");
         ble_connection_t *new_conn = zjs_ble_new_connection(&ble_handle->connections, conn);
         if (!new_conn) {
-            ERR_PRINT("failed to create new cilent connection\n");
+            ERR_PRINT("failed to create new client connection\n");
             return;
         }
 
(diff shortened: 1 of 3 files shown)


grgustaf:
@jimmy-huang, if you rebase maybe this will work because we removed sensor from ashell?

src/zjs_ble.c
```diff
@@ -375,14 +422,17 @@ static ZJS_DECL_FUNC(zjs_ble_update_value_callback_function)
...
-        if (jerry_get_object_native_handle(this, &ptr)) {
-            ble_characteristic_t *chrc = (ble_characteristic_t *)ptr;
-            if (chrc->chrc_attr) {
-                zjs_buffer_t *buf = zjs_buffer_find(argv[0]);
-                bt_gatt_notify(ble_conn.default_conn, chrc->chrc_attr,
+    uintptr_t ptr;
+    if (jerry_get_object_native_handle(this, &ptr)) {
+        ble_characteristic_t *chrc = (ble_characteristic_t *)ptr;
+        if (chrc->chrc_attr) {
+            zjs_buffer_t *buf = zjs_buffer_find(argv[0]);
```
grgustaf: You may be, but just to be sure - you're guaranteed this will succeed, right?

src/zjs_ble.c
```diff
@@ -464,45 +514,71 @@ static void zjs_ble_blvl_ccc_cfg_changed(const struct bt_gatt_attr *attr, uint16
 
 static void zjs_ble_connected_c_callback(void *handle, const void *argv)
 {
+    ble_handle_t * h = (ble_handle_t *)handle;
```
grgustaf: drop space after first *, also twice below

src/zjs_ble.c
```diff
@@ -464,45 +514,71 @@ static void zjs_ble_blvl_ccc_cfg_changed(const struct bt_gatt_attr *attr, uint16
...
 
+static void zjs_ble_disconnected_c_callback(void *handle, const void *argv)
+{
+    ble_handle_t * h = (ble_handle_t *)handle;
+    if (!h) {
+        ERR_PRINT("handle not found\n");
+        return;
+    }
+
+    char *addr = (char *)argv;
```
grgustaf: You shouldn't be casting away const-ness like this. Can you fix that?

src/zjs_ble.c
```diff
@@ -464,45 +514,71 @@ static void zjs_ble_blvl_ccc_cfg_changed(const struct bt_gatt_attr *attr, uint16
...
 {
     if (err) {
-        DBG_PRINT("Connection failed (err %u)\n", err);
+        ERR_PRINT("Connection failed (err %u)\n", err);
     } else {
-        DBG_PRINT("========== connected ==========\n");
+        DBG_PRINT("new connection\n");
+        ble_connection_t *new_conn = zjs_ble_new_connection(&ble_handle->connections, conn);
+        if (!new_conn) {
+            ERR_PRINT("failed to create new cilent connection\n");
```
grgustaf: client

src/zjs_ble.c
jimmy-huang: what do you mean? you are saying I don't have to to check buffer? it will always exists?

grgustaf: You're *not* checking it so I want to make sure you know it's always going to exist. That means you will never call this on an object that is not a valid buffer.

grgustaf: Looking into it more myself, no it looks like this is directly called by JS so you can't be sure what they'll pass in. So you need to check whether buf is NULL here.

jimmy-huang: @grgustaf added check now, i was looking at the read function before so I thought I already check the buffer

grgustaf:
+1

PR #1030, opened 2017-04-18
WIP: Attempt at self-contained dgram module tests.

pfalcon:
This is a preview of what I'd like to achieve - fully automated tests which exercise (some subset) of network functionality. I wrote these quite some time ago, it initially didn't work because of https://jira.zephyrproject.org/browse/ZEP-1966 , but now they actually do.

Then next step is how to run these. These would run on a Linux host, but in qemu_x86. That's where I'm backlogged so far. I'm posting these just to share the idea (feel free to close this PR when it hanged around enough).

Note that beyond simple tests, this doesn't work. E.g. local TCP connections aren't possible: https://jira.zephyrproject.org/browse/ZEP-2001

### DIFF
--- a/UDP-my/test.js
+++ b/UDP-my/test.js
@@ -1,0 +1,33 @@
+// Copyright (c) 2017, Linaro Limited.
+
+// Sample for "dgram" module implementing an IPv4 UDP echo server.
+// To run it on QEMU, you'll need additional setup for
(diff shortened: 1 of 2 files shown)


grgustaf:
We do have a suite of similar tests now, so I think we can safely close this. Thanks for the examples.

PR #1032, opened 2017-04-18
[ashell] Removing sensor from ashell default configuration

brianjjones:
If sensor is enabled, it will cause GPIO input to be disabled for
all pins.  Because GPIO is used more often than sensor, sensor is
being taken out of the default ashell configuration.  You can
enable sensor again simply by uncommenting it in the file.


grgustaf:
+1

grgustaf:
Also cherry-picked this into zjs-0.3 branch.

PR #1033, opened 2017-04-18
[modules] Acquire instance when returning in subsequent require calls

grgustaf:


grgustaf:
I think this is needed; just stumbled upon it while leafing through code. However, I couldn't come up with a way to test it yet.

jimmy-huang:
+1

PR #1034, opened 2017-04-19
[assert] Remove arguments.callee usage, not allowed in strict mode

grgustaf:
Naming the assert function accomplishes the same thing here.


jimmy-huang:
+1

PR #1037, opened 2017-04-19
[websockets] Initial implementation 

jprestwo:


src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
+// Copyright (c) 2017, Intel Corporation.
+
+// ZJS includes
+#include "zjs_util.h"
```
jimmy-huang: i think these includes should go after the <> ones

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+// generate an accept key given an input key
+static void generate_key(char *key, uint32_t len, char *output, uint32_t olen)
+{
+    char *concat = zjs_malloc(strlen(key) + strlen(magic) + 1);
+    memset(concat, 0, strlen(key) + strlen(magic) + 1);
+    memcpy(concat, key, strlen(key));
+    strcat(concat, magic);
+    char sha_out[20];
+    // compute sha1 hash of concatenated key + magic
+    mbedtls_sha1(concat, strlen(concat), sha_out);
```
jimmy-huang: use strnlen

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+static void dump_packet(ws_packet_t *packet)
+{
+    ZJS_PRINT("FIN=         0x%02x\n", packet->fin);
+    ZJS_PRINT("RSV1=        0x%02x\n",packet->rsv1);
+    ZJS_PRINT("RSV2=        0x%02x\n",packet->rsv2);
+    ZJS_PRINT("RSV3=        0x%02x\n",packet->rsv3);
+    ZJS_PRINT("OPCODE=      0x%02x\n", packet->opcode);
+    ZJS_PRINT("MASK=        0x%02x\n", packet->mask_bit);
+    ZJS_PRINT("PayloadLen=  %d\n",  packet->payload_len);
+    ZJS_PRINT("PayloadMask= [%02x%02x%02x%02x]\n", packet->mask[0], packet->mask[1], packet->mask[2], packet->mask[3]);
```
jimmy-huang: wrap this

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+            ZJS_PRINT("%s\n", line);
+            idx = 0;
+            memset(line, ' ', sizeof(line));
+        }
+    }
+}
+#endif
+
+// encode data into a WS packet.
+static int encode_packet(ws_packet_type type, uint8_t mask, void *payload, uint16_t len, uint8_t *out)
```
jimmy-huang: wrap this

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+    // copy data from read buffer
+    memcpy(zbuf->buffer, con->rptr, len);
+    con->rptr += len;
+    if (con->rptr == con->wptr) {
+        con->rptr = con->wptr = con->rbuf;
+        DBG_PRINT("all data consumed, reseting read buffer\n");
+    }
+    switch (opcode) {
+    case WS_PACKET_CONTINUATION:
+        // continuation frame
```
jimmy-huang: do you need to add break here? otherwise it's gonna trigger_event

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+    default:
+        ERR_PRINT("opcode 0x%02x not recognized\n", opcode);
+        break;
+    }
+}
+
+static void close_connection(ws_connection_t *con)
+{
+    ws_connection_t *cur = connections;
+    if (cur->next == NULL) {
```
jimmy-huang: should check if cur == con, right?

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+
+static void tcp_received(struct net_context *context,
+                         struct net_buf *buf,
+                         int status,
+                         void *user_data)
+{
+    ws_connection_t *con = (ws_connection_t *)user_data;
+    uint32_t len = net_nbuf_appdatalen(buf);
+    uint8_t *buffer = net_nbuf_appdata(buf);
+
```
jimmy-huang: remove line

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+        }
+    }
+
+    if (len && buffer) {
+        struct net_buf *tmp = buf->frags;
+        uint32_t header_len = net_nbuf_appdata(buf) - tmp->data;
+        // move past IP header
+        net_buf_pull(tmp, header_len);
+
+        uint8_t *data = zjs_malloc(len);
```
jimmy-huang: check malloc fail

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+            memcpy(wptr, tmp->data, tmp->len);
+            wptr += tmp->len;
+            tmp = tmp->frags;
+        }
+
+#ifdef DEBUG_BUILD
+        dump_bytes("DATA", data, len);
+#endif
+        if (con->server_handle->max_payload &&
+                (len > con->server_handle->max_payload)) {
```
jimmy-huang: align here

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+    if (ret < 0) {
+        ERR_PRINT("Cannot receive TCP packet (family %d)\n",
+            net_context_get_family(context));
+        // this seems to mean the remote exists but the connection was not made
+        zjs_trigger_event(handle->server, "error", NULL, 0, NULL, NULL);
+        return;
+    }
+}
+
+static jerry_value_t ws_server(const jerry_value_t function_obj,
```
jimmy-huang: use the ZJS_DECL_FUNC here

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+                               const jerry_length_t argc)
+{
+    ZJS_VALIDATE_ARGS_OPTCOUNT(optcount, Z_OBJECT, Z_OPTIONAL Z_FUNCTION);
+
+    int ret;
+    double port = 0;
+    bool backlog = 0;
+    double max_payload = 0;
+
+    server_handle_t *handle = zjs_malloc(sizeof(server_handle_t));
```
jimmy-huang: check of malloc failed

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,793 @@
...
+    memset(handle, 0, sizeof(server_handle_t));
+
+    jerry_value_t server = jerry_create_object();
+
+    zjs_obj_get_double(argv[0], "port", &port);
```
jimmy-huang: should  check if these values actually exists?

src/zjs_web_sockets.c
jprestwo: the only difference between the two is that continuation is part of the previous packet. The data is still given to JS the same though. I have not actually encountered this case yet because there is no way to force it from the Node side.

src/zjs_web_sockets.c
jprestwo: if they don't exist then the port will/backlog/tracking/payload will be unchanged i.e. zero, which is fine.

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+        }
+        return mask_offset + 4 + len;
+    } else {
+        // no need to mask data, just copy payload as is
+        memcpy(out + mask_offset, payload, len);
+        return mask_offset + len;
+    }
+}
+
+// decode WS TCP data into a ws_packet_t type
```
brianjjones: Add a note that packet payload needs to be freed.

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+        ERR_PRINT("only support payload <= 65,535 bytes\n");
+        return -1;
+    }
+    // payload mask, 4 bytes
+    packet->mask[0] = data[mask_offset];
+    packet->mask[1] = data[mask_offset + 1];
+    packet->mask[2] = data[mask_offset + 2];
+    packet->mask[3] = data[mask_offset + 3];
+    packet->payload = zjs_malloc(packet->payload_len + 1);
+    memset(packet->payload, 0, packet->payload_len);
```
brianjjones: Check that zjs_malloc was successful before writing to packet->payload

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+            (len > con->server_handle->max_payload)) {
+            ZVAL error = zjs_error("payload too large");
+            zjs_trigger_event(con->server, "error", &error, 1, NULL, NULL);
+            net_nbuf_unref(buf);
+            zjs_free(data);
+            return;
+        }
+        if (con->state == UNCONNECTED) {
+            ZVAL_MUTABLE protocol_list;
+            con->accept_key = zjs_malloc(64);
```
brianjjones: Check that zjs_malloc was successful

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+                        DBG_PRINT("New protocol requested: %s\n", iter);
+                        iter += len + 1;
+                    }
+                }
+                i++;
+            }
+            // no handler registered, automatically accept connection
+            if (con->accept_handler_id == -1) {
+                // create the accept response
+                char *send_data = zjs_malloc(strlen(accept_header) + strlen(con->accept_key) + 6);
```
brianjjones: Check that zjs_malloc was successful

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+    if (!size) {
+        ERR_PRINT("acceptHandler() protocol string too large\n");
+        return;
+    }
+
+    // create the accept response
+    uint32_t sdata_size = strlen(accept_header) +
+                          strlen(con->accept_key) +
+                          strlen(proto) + 32;
+    char *send_data = zjs_malloc(sdata_size);
```
brianjjones: Check that zjs_malloc was successful

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+                         int error,
+                         void *user_data)
+{
+    int ret;
+    server_handle_t *handle = (server_handle_t *)user_data;
+    DBG_PRINT("connection made, context %p error %d\n", context, error);
+    ws_connection_t *new = zjs_malloc(sizeof(ws_connection_t));
+    memset(new, 0, sizeof(ws_connection_t));
+    new->tcp_sock = context;
+    new->rbuf = zjs_malloc(DEFAULT_WS_BUFFER_SIZE);
```
brianjjones: Check that zjs_malloc was successful for both new and new->rbuf

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+        return;
+    }
+
+    net_nbuf_unref(send_buf);
+}
+
+// generate an accept key given an input key
+static void generate_key(char *key, uint32_t len, char *output, uint32_t olen)
+{
+    char *concat = zjs_malloc(strlen(key) + strlen(magic) + 1);
```
brianjjones: Check that zjs_malloc was successful

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+        zjs_trigger_event(con->conn, "error", &error, 1, NULL, NULL);
+        return;
+    }
+    memset(packet, 0, sizeof(ws_packet_t));
+    if (decode_packet(packet, data, len) < 0) {
+        ERR_PRINT("error decoding packet\n");
+        zjs_free(packet->payload);
+        zjs_free(packet);
+        ZVAL_MUTABLE error = zjs_custom_error("PacketDecodeError", "error decoding packet");
+        jerry_value_clear_error_flag(&error);
```
brianjjones: Why does the error flag need to be cleared before calling the trigger_event?

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,802 @@
...
+    CHECK(net_context_listen(handle->tcp_sock, (int)backlog));
+    CHECK(net_context_accept(handle->tcp_sock, tcp_accepted, 0, handle));
+
+    if (optcount) {
+        zjs_add_event_listener(handle->server, "listening", argv[1]);
+        zjs_trigger_event(handle->server, "listening", NULL, 0, NULL, NULL);
+    }
+
+    handle->server = server;
+    handle->max_payload = (uint16_t)max_payload;
```
brianjjones: I don't see where handle gets freed. Is it happening elsewhere?

src/zjs_web_sockets.c
jprestwo: because you cant use a object that has the error flag set as a parameter. Its a JrS limitation.

brianjjones: Good to know thanks.

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,867 @@
...
+
+// structure to hold a decoded WS packet
+typedef struct ws_packet {
+    uint8_t fin;
+    uint8_t rsv1;
+    uint8_t rsv2;
+    uint8_t rsv3;
+    uint8_t opcode;
+    uint8_t mask_bit;
+    uint32_t payload_len;
```
jprestwo: Regarding mask, do you mean change it to a uint32? or can it remain a uint8 array? Its easier to do the unmasking if its an array.

src/zjs_web_sockets.c
```diff
@@ -0,0 +1,867 @@
...
+                if (cur->accept_key) {
+                    zjs_free(cur->accept_key);
+                }
+                zjs_remove_callback(cur->trigger_msg_id);
+                zjs_remove_callback(cur->accept_handler_id);
+                zjs_free(cur);
+                DBG_PRINT("Freed socket: opened_sockets=%p\n", connections);
+            }
+        }
+    }
```
jprestwo: list code will replace this. If that goes in before this PR I will update it.

src/zjs_web_sockets.c
grgustaf: Array is fine. I'm just talking about position. You have six uint8's before a uint32, so probably two bytes of padding will be inserted in between them to align the uint32.

src/zjs_web_sockets.c
jprestwo: This handle is the server handle. When I set the server objects native handle I register a free function (free_server()), line 790.

src/zjs_web_sockets.c
```diff
@@ -156,13 +156,15 @@ static void tcp_send(struct net_context *context, void *data, uint32_t len)
...
+    char *concat = zjs_malloc(concat_size);
     if (!concat) {
         ERR_PRINT("could not allocate key\n");
         return;
     }
-    memset(concat, 0, strlen(key) + strlen(magic) + 1);
-    memcpy(concat, key, strlen(key));
+    //memset(concat, 0, concat_size);
+    strcpy(concat, key);
+    //memcpy(concat, key, strlen(key));
```
grgustaf: please remove this commented-out code

grgustaf:
+1

PR #1039, opened 2017-04-19
[tests] Callback testing module

jprestwo:
- Can directly create/call/remove callbacks from JS in order to
   test JS callbacks
 - Fixed several bugs in callbacks that were found using this test
 - Added test to trlite

### DIFF
--- a/scripts/trlite
+++ b/scripts/trlite
@@ -392,7 +392,7 @@ if [ "$RUN" == "all" -o "$RUN" == "3" ]; then
 
     # linux runtime tests
     # TODO: Add buffer, it fails in Linux ATM
-    for i in buffer buffer-rw eval error event promise timers callbacks; do
+    for i in buffer buffer-rw callbacks eval error event promise timers; do
         ./outdir/linux/release/jslinux tests/test-$i.js > /tmp/output
         cat /tmp/output
         try_command "t-$i" test ! $(grep "FAIL" /tmp/output)
--- a/scripts/trlite
+++ b/scripts/trlite
@@ -392,7 +392,7 @@ if [ "$RUN" == "all" -o "$RUN" == "3" ]; then
 
     # linux runtime tests
     # TODO: Add buffer, it fails in Linux ATM
-    for i in buffer buffer-rw eval error event promise timers; do
+    for i in buffer buffer-rw eval error event promise timers callbacks; do
         ./outdir/linux/release/jslinux tests/test-$i.js > /tmp/output
         cat /tmp/output
         try_command "t-$i" test ! $(grep "FAIL" /tmp/output)
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -440,7 +440,9 @@ void signal_callback_priv(zjs_callback_id id,
     LOCK();
     DBG_PRINT("pushing item to ring buffer. id=%d, args=%p, size=%lu\n", id,
               args, size);
-    if (id >= 0 && cb_map[id]) {
+    if (id < 0 || cb_map[id]) {
+        DBG_PRINT("callback ID %u does not exist\n", id);
+    } else {
         if (GET_CB_REMOVED(cb_map[id]->flags)) {
             DBG_PRINT("callback already removed\n");
             return;
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -440,50 +440,42 @@ void signal_callback_priv(zjs_callback_id id,
     LOCK();
     DBG_PRINT("pushing item to ring buffer. id=%d, args=%p, size=%lu\n", id,
               args, size);
-    if (id < 0 || cb_map[id]) {
+    if (id < 0 || !cb_map[id]) {
         DBG_PRINT("callback ID %u does not exist\n", id);
-    } else {
-        if (GET_CB_REMOVED(cb_map[id]->flags)) {
-            DBG_PRINT("callback already removed\n");
-            return;
+        return;
+    }
+    if (GET_CB_REMOVED(cb_map[id]->flags)) {
+        DBG_PRINT("callback already removed\n");
+        return;
+    }
+    if (GET_TYPE(cb_map[id]->flags) == CALLBACK_TYPE_JS) {
+        // for JS, acquire values and release them after servicing callback
+        int argc = size / sizeof(jerry_value_t);
+        jerry_value_t *values = (jerry_value_t *)args;
+        for (int i=0; i<argc; i++) {
+            jerry_acquire_value(values[i]);
         }
     }
-
+#ifdef DEBUG_BUILD
+        set_info_string(cb_map[id]->caller, file, func);
+#endif
     int ret = zjs_port_ring_buf_put(&ring_buffer,
-                                    (uint16_t)id,
-                                    0,  // we use value for CB_FLUSH_ONE/ALL
-                                    (uint32_t *)args,
-                                    (uint8_t)((size + 3) / 4));
+            (uint16_t)id,
+            0,  // we use value for CB_FLUSH_ONE/ALL
+            (uint32_t *)args,
+            (uint8_t)((size + 3) / 4));
     if (ret != 0) {
         if (GET_TYPE(cb_map[id]->flags) == CALLBACK_TYPE_JS) {
             // for JS, acquire values and release them after servicing callback
             int argc = size / sizeof(jerry_value_t);
             jerry_value_t *values = (jerry_value_t *)args;
             for (int i=0; i<argc; i++) {
-                jerry_acquire_value(values[i]);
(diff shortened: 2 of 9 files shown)


src/zjs_callbacks.c
```diff
@@ -420,35 +420,40 @@ void signal_callback_priv(zjs_callback_id id,
...
-
-    if (GET_TYPE(cb_map[id]->flags) == CALLBACK_TYPE_JS) {
-        // for JS, acquire values and release them after servicing callback
-        int argc = size / sizeof(jerry_value_t);
-        jerry_value_t *values = (jerry_value_t *)args;
-        for (int i=0; i<argc; i++) {
-            jerry_acquire_value(values[i]);
+    if (id >= 0 && cb_map[id]) {
+        if (GET_CB_REMOVED(cb_map[id]->flags)) {
+            DBG_PRINT("callback already removed");
```
jimmy-huang: add \n

src/zjs_callbacks.c
```diff
@@ -420,35 +420,40 @@ void signal_callback_priv(zjs_callback_id id,
...
             for (int i=0; i<argc; i++) {
-                jerry_release_value(values[i]);
+                jerry_acquire_value(values[i]);
             }
+#ifdef DEBUG_BUILD
+            set_info_string(cb_map[id]->caller, file, func);
+#endif
         }
+        int ret = zjs_port_ring_buf_put(&ring_buffer,
+                (uint16_t)id,
```
jimmy-huang: align these

tests/test-callbacks.js
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2017, Intel Corporation.
+
+var assert = require("Assert.js");
+var cb = require('test_callbacks');
+
+var times = 0;
+function callback(arg1, arg2, arg3)
+{
+    assert((arg1 == "one" && arg2 == "two" && arg3 == "three" && this.prop == "a prop"),  "args matched (" + times + ")");
```
jimmy-huang: you need to wrap these, as it won't fit in the ide

tests/test-callbacks.js
```diff
@@ -0,0 +1,63 @@
...
+    list_count++;
+}
+
+var id = cb.addCallback(callback, {prop: "a prop"});
+cb.signalCallback(id, "one", "two", "three");
+cb.signalCallback(id, "one", "two", "three");
+cb.signalCallback(id, "one", "two", "three");
+
+setTimeout(function() {
+    assert(times == 3, "signalled 3 times");
```
jimmy-huang: use ===

tests/test-callbacks.js
```diff
@@ -0,0 +1,63 @@
...
+    cb.removeCallback(id);
+    cb.signalCallback(id, "one", "two", "three");
+    setTimeout(function() {
+        assert(times == 3, "removed callback not called");
```
jimmy-huang: same here

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+
+setTimeout(function() {
+    assert(times === 3, "signaled 3 times");
+    cb.removeCallback(id);
+    cb.signalCallback(id, "one", "two", "three");
+    setTimeout(function() {
+        assert(times === 3, "removed callback not called");
+        cb.signalCallback(-1);
+        cb.signalCallback(-10);
+        cb.signalCallback(12345);
```
cuiyanx: Maybe use number string and boolean to better.

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+        setTimeout(function() {
+            assert(true, "with bad ID's didn't crash");
```
cuiyanx: Can we judge it? More than just a conclusion, that is not a promise function.

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+var cb = require('test_callbacks');
+
+var times = 0;
+function callback(arg1, arg2, arg3)
+{
+    assert((arg1 == "one" &&
+            arg2 == "two" &&
+            arg3 == "three" &&
+            this.prop == "a prop"),
+            "args matched (" + times + ")");
```
cuiyanx: Please run  once.

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+var list_count = 0;
+function callback_l1(arg1, arg2, arg3)
+{
+    assert((arg1 == "one" &&
+            arg2 == "two" &&
+            arg3 == "three" &&
+            this.prop == "list"),
+            "list args matched (" + list_count + ")");
```
cuiyanx: Same here.

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+    list_count++;
+}
+
+function callback_l2(arg1, arg2, arg3)
+{
+    assert((arg1 == "one" &&
+            arg2 == "two" &&
+            arg3 == "three" &&
+            this.prop == "list"),
+            "list args matched (" + list_count + ")");
```
cuiyanx: here

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+    list_count++;
+}
+
+function callback_l3(arg1, arg2, arg3)
+{
+    assert((arg1 == "one" &&
+            arg2 == "two" &&
+            arg3 == "three" &&
+            this.prop == "list"),
+            "list args matched (" + list_count + ")");
```
cuiyanx: here

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2017, Intel Corporation.
+
+var assert = require("Assert.js");
+var cb = require('test_callbacks');
+
+var times = 0;
+function callback(arg1, arg2, arg3)
```
cuiyanx: Can we set more args? like 20.

tests/test-callbacks.js
```diff
@@ -0,0 +1,80 @@
...
+    cb.signalCallback(id, "one", "two", "three");
+    setTimeout(function() {
+        assert(times === 3, "removed callback not called");
+        cb.signalCallback(-1);
+        cb.signalCallback(-10);
+        cb.signalCallback(12345);
+        setTimeout(function() {
+            assert(true, "with bad ID's didn't crash");
+            var list_id = cb.addCallbackList(callback_l1, {prop: "list"}, -1);
+            cb.addCallbackList(callback_l2, {prop: "list"}, list_id);
```
cuiyanx: Can we add callback list to another callback list?

tests/test-callbacks.js
jprestwo: Changing these to anything other than a number wouldn't really be testing the callback module, but rather the callback test module itself. These numbers are passed directly to the C API's so it would essentially "do nothing" because we would have to throw away the argument.

tests/test-callbacks.js
jprestwo: Since its the main loop (service_callbacks()) which actually calls the callback that may throw an error there is no way we can catch it unfortunately. All you get is a print that an error happened. This is a limitation of JerryScript which would be nice to get figured out but for now its not possible.

And no you cant add a callback list to another list.

scripts/trlite
```diff
@@ -392,7 +392,7 @@ if [ "$RUN" == "all" -o "$RUN" == "3" ]; then
 
     # linux runtime tests
     # TODO: Add buffer, it fails in Linux ATM
-    for i in buffer buffer-rw eval error event promise timers; do
+    for i in buffer buffer-rw eval error event promise timers callbacks; do
```
grgustaf: these are alphabetical to make finding one easy

src/zjs_callbacks.c
```diff
@@ -420,35 +420,40 @@ void signal_callback_priv(zjs_callback_id id,
...
     DBG_PRINT("pushing item to ring buffer. id=%d, args=%p, size=%lu\n", id,
               args, size);
-
-    if (GET_TYPE(cb_map[id]->flags) == CALLBACK_TYPE_JS) {
-        // for JS, acquire values and release them after servicing callback
-        int argc = size / sizeof(jerry_value_t);
-        jerry_value_t *values = (jerry_value_t *)args;
-        for (int i=0; i<argc; i++) {
-            jerry_acquire_value(values[i]);
+    if (id >= 0 && cb_map[id]) {
```
grgustaf: How about change this to
```
if (id < 0 || !cb_map[id]) {
   // some error
   return;
}

Then that odd situation will get a dbg print or something, instead of being ignored... and the indentation won't change so much so easier to review. :)

grgustaf:
+1, needs rebase though

grgustaf:
+1

PR #1041, opened 2017-04-19
[headers] Warn if libc6-dev-i386 package not installed

grgustaf:
This package prevents some non-helpful errors from appearing like:
   fatal error: sys/cdefs/h: no such file

Which might be hard to figure out, otherwise.

### DIFF
--- a/scripts/checkheaders
+++ b/scripts/checkheaders
@@ -63,6 +63,15 @@ fi
 echo RESULTS: $PASSCOUNT/$TOTAL passed
 
 if [ $PASSCOUNT -ne $TOTAL ]; then
+    # on Ubuntu ensure libc6-dev-i386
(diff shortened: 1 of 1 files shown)


brianjjones:
+1

scripts/checkheaders
```diff
@@ -63,6 +63,15 @@ fi
 echo RESULTS: $PASSCOUNT/$TOTAL passed
 
 if [ $PASSCOUNT -ne $TOTAL ]; then
+    # on Ubuntu ensure libc6-dev-i386
```
jimmy-huang: this won't work on mac, do you intend this to be ran on ubuntu only?

grgustaf: Are you sure? If they have `which` then the `which dpkg` test should fail.

PR #1042, opened 2017-04-19
Ashell monkey

jprestwo:
Script which allows you to feed in a list of test files and test them using ashell for semi-automated target testing.

### DIFF
--- a/scripts/ashell-monkey.py
+++ b/scripts/ashell-monkey.py
@@ -1,0 +1,99 @@
+# This tool can test many JS test files at once using ashell. To use, flash a
+# ZJS device with ashell and start it. Then run this script, passing in the
+# files you want to test, as well as the serial port the board is connected to:
+
+#    python scripts/ashell-monkey.py <port> [print] <list of files>
+
+# The port parameter should be something like "/dev/ttyACM0"
+# After the port, you can pass in an optional "print" string, which will print
+# all the ashell output.
+# After the port and print parameters pass in a list of files you would like
+# to test. These should be ZJS test files which end in printing:
+
+#    "TOTAL: X of Y tests passed"
+
+# This is the string that ashell-monkey looks for to determine if the test is
+# over, as well as if it passed or not. Here is an example of how you might
+# run this script:
+
+#    python scripts/ashell-monkey /dev/ttyACM0 print tests/test-buffer.js \
+#        tests/test-fs.js tests/test-error.js tests/test-event.js
+
+# This would run the buffer, fs, error, and event test scripts found under
+# tests/. At the end you should get a message if any tests failed or if all
+# succeeded.
+
+import serial
+import re
+import sys
+from time import sleep
+
+start_idx = 2
+do_print = False
+if sys.argv[2] == "print":
+    do_print = True
+    start_idx = 3
+
+print "TESTING FILES:"
+for i in sys.argv[start_idx:]:
+    print i
+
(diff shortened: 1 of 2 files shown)


src/ashell/jerry-code.c
```diff
@@ -326,8 +326,17 @@ void restore_zjs_api() {
 #ifdef BUILD_MODULE_SENSOR
     zjs_sensor_init();
 #endif
+    jerry_value_t global_obj = jerry_get_global_object();
```
brianjjones: You can take this out, as it has already been fixed by #1044

scripts/ashell-monkey.py
```diff
@@ -0,0 +1,99 @@
...
+        data = ser.read(9999)
+        if len(data) > 0:
+            bytes += data
+            if do_print:
+                print bytes
+
+        if state == "SETUP":
+            if "acm>" in bytes:
+                print 'Ashell is ready'
+                ser.write('load Assert.js\r')
```
brianjjones: You could use 'du assert.js' to determine if Assert.js is already on the filesystem. It only needs to be done once, and it will stay on the FS until its erased.

jprestwo: I could, but I also want to ensure that I have MY Assert.js on the device.

brianjjones: Fair enough, and its not like it takes a bunch of cycles to copy it

scripts/ashell-monkey.py
```diff
@@ -0,0 +1,99 @@
+# This tool can test many JS test files at once using ashell. To use, flash a
```
grgustaf: How about add:
`#!/usr/bin/env python3`

... and copyright below that.

... and you'll have to fix your print statements to use parentheses, required in Python3, hopefully nothing else.

scripts/ashell-monkey.py
```diff
@@ -0,0 +1,99 @@
...
+# to test. These should be ZJS test files which end in printing:
+
+#    "TOTAL: X of Y tests passed"
+
+# This is the string that ashell-monkey looks for to determine if the test is
+# over, as well as if it passed or not. Here is an example of how you might
+# run this script:
+
+#    python scripts/ashell-monkey /dev/ttyACM0 print tests/test-buffer.js \
+#        tests/test-fs.js tests/test-error.js tests/test-event.js
```
grgustaf: Then remove the .py extension and change your examples to say "ashell-monkey" instead of "python ashell-monkey".

And chmod +x it.

scripts/ashell-monkey.py
```diff
@@ -0,0 +1,99 @@
...
+
+# This would run the buffer, fs, error, and event test scripts found under
+# tests/. At the end you should get a message if any tests failed or if all
+# succeeded.
+
+import serial
```
grgustaf: Add documentation above that you need to "pip3 install pyserial" for this to work.

scripts/ashell-monkey.py
grgustaf: Will these still work if you send `\r\n` instead of just `\r`? Ending a line in just `\r` is not normal and it seems like you're exploiting a bug in ashell.

grgustaf:
We theorized we could accomplish the same thing with (Babu's?) command-line tool. So I don't think we're really going to look at this, but marking it 'reconsider later'.

PR #1043, opened 2017-04-19
[callbacks] Lock scheduler on zephyr during important callback operat…

jprestwo:
…ions

 - Call k_sched_lock() and k_sched_unlock() in callbacks to prevent
   other tasks from accessing the callback map/ring buffer at the
   same time. This becomes important for network functionality as
   the zephyr network stack uses multiple tasks, which can still
   call add/signal/remove_callback().

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -95,14 +95,6 @@ static int zjs_ringbuf_error_count = 0;
 static int zjs_ringbuf_error_max = 0;
 static int zjs_ringbuf_last_error = 0;
 
-#ifndef ZJS_LINUX_BUILD
-#define LOCK  k_sched_lock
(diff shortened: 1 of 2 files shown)


src/zjs_callbacks.c
```diff
@@ -95,6 +95,14 @@ static int zjs_ringbuf_error_count = 0;
 static int zjs_ringbuf_error_max = 0;
 static int zjs_ringbuf_last_error = 0;
 
+#ifndef ZJS_LINUX_BUILD
```
jimmy-huang: This can apply to some other modules as well, would it be better if this is in zjs_utils?

jimmy-huang:
+1

brianjjones:
+1

PR #1044, opened 2017-04-20
[ashell] Move common init code to zjs_modules.c for main/ashell

jprestwo:

### DIFF
--- a/src/zjs_modules.c
+++ b/src/zjs_modules.c
@@ -20,6 +20,7 @@
 #include "zjs_event.h"
 #include "zjs_modules.h"
 #include "zjs_performance.h"
+#include "zjs_callbacks.h"
 #ifdef BUILD_MODULE_SENSOR
 #include "zjs_sensor.h"
 #endif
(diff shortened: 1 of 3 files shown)


src/zjs_modules.c
```diff
@@ -206,9 +208,53 @@ static ZJS_DECL_FUNC(native_require_handler)
...
+    #endif
+    zjs_modules_cleanup();
+    jerry_cleanup();
+    return ZJS_UNDEFINED;
+}
+
 void zjs_modules_init()
 {
+    // Add module.exports to global namespace
     ZVAL global_obj = jerry_get_global_object();
```
brianjjones: Need to release these three jerry created items at the end of this function.

jimmy-huang: @brianjjones, the reason to use ZVAL is that it's auto-released, you don't have to call jerry_release() on them, that's why I mentioned in your PR

jimmy-huang:
+1

PR #1045, opened 2017-04-20
[ashell] Fix for calling eval after starting a new application

brianjjones:
Currently jerry_eval will fail when ashell runs it after cleaning
the jerry engine.  This fixes it.

### DIFF
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -326,8 +326,17 @@ void restore_zjs_api() {
 #ifdef BUILD_MODULE_SENSOR
     zjs_sensor_init();
 #endif
+    jerry_value_t global_obj = jerry_get_global_object();
+    jerry_value_t modules_obj = jerry_create_object();
+    jerry_value_t exports_obj = jerry_create_object();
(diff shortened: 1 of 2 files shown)


src/ashell/jerry-code.c
```diff
@@ -326,8 +326,17 @@ void restore_zjs_api() {
 #ifdef BUILD_MODULE_SENSOR
     zjs_sensor_init();
 #endif
+    jerry_value_t global_obj = jerry_get_global_object();
```
jimmy-huang: should just use ZVAL here, so you don't have to free it, this is a weird place to set this also, shouldnt this be in the zjs_modules_init() function?

brianjjones: This is what is done in main.c right now.  I'm fine with moving it to modules_init, but that's for another patch.

brianjjones: @jimmy-huang https://github.com/01org/zephyr.js/pull/1044 will take care of this.

brianjjones:
https://github.com/01org/zephyr.js/pull/1044 takes this fix and adds it to the moving of the init code.

PR #1047, opened 2017-04-20
Ashell single thread

jprestwo:


jimmy-huang:
+1

brianjjones:
-1 Afraid it doesn't work for any of the native calls that use IPM.  I knew there was a reason I didn't move ashell to a single thread, now I remember (sorry for late realization). At the time I tried I didn't have time to dig in, so I shelved it for the moment.  Its probably still possible for this to work - but there's something broken about IPM at the moment that you'll need to fix.

brianjjones:
Works for me now, thanks! +1

PR #1048, opened 2017-04-20
[gpio] Remove BUILD_MODULE_GPIO ifdef around zjs_gpio.c file

grgustaf:
I'm pretty sure this is no longer needed because building the module
is controlled by ZJS_GPIO but want to send out this canary before
changing all the modules. BUILD_MODULE_GPIO is still used in modules.c
and *pins.c to control a few sections though.

Also, fix a string in analyze and add a comment to the top of zjs.conf
describing how it is used, which I found along the way.


brianjjones:
LGTM +1

PR #1049, opened 2017-04-20
[ashell] Fixing using native modules with long names in ashell

brianjjones:
I was checking the length of the filename against the maximum
JS file length. Native modules can be longer and were failing
because of this.

### DIFF
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -202,17 +202,17 @@ static bool add_requires(requires_list_t **list, char *filebuf)
     char *ptr2 = NULL;
     char *filestr = NULL;
     char *ext_check = NULL;
+    bool valid = true;
 
     while (ptr1 != NULL) {
         // Find next instance of "require"
         ptr1 = strstr(ptr1, "require");
-
         if (ptr1 != NULL) {
             // Find the text between the two " " after require
             ptr1 = strchr(ptr1, '"') + 1;
             ptr2 = strchr(ptr1, '"');
             size_t len = ptr2 - ptr1;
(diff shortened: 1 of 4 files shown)


src/zjs_modules.h
```diff
@@ -10,6 +10,7 @@
 #include "jerryscript.h"
 
 #define NUM_SERVICE_ROUTINES 3
+#define MAX_MODULE_LEN 32
```
jimmy-huang: should name this MAX_MODULE_STR_LEN, MODULE_LEN is confusing

src/ashell/jerry-code.c
```diff
@@ -221,9 +221,14 @@ static bool add_requires(requires_list_t **list, char *filebuf)
                 // Check that this is a *.js require and not a zephyr require
                 // Also check that the file hasn't already been loaded
                 ext_check = &filestr[len - 3];
-
                 if (strcmp(ext_check, ".js") == 0 && !list_contains(list, filestr)) {
-                    add_to_list(list, filestr);
+                    if (fs_valid_filename(filestr)) {
+                        add_to_list(list, filestr);
+                    }
+                    else {
```
jimmy-huang: you can just in else statement to free the filestr, and return false immediately, instead of adding this valid bool check, it's cleaner

brianjjones: I was attempting to not have multiple places where I had to clean up the filestr and just do it at the end

brianjjones: @jimmy-huang changed it back

jimmy-huang:
+1

PR #1050, opened 2017-04-21
[sensor] Modularize sensor into separate submodules

jimmy-huang:
Instead of having all the different Sensor implementations
all exists in a single file, it's cleaner to break it out
into separate sub modules, which then can be loaded separately.

The sensor.c just implements the core sensor logic and will
have the individual sensors handle the data through callbacks.

### DIFF
--- a/src/sensors/zjs_sensor_accel.c
+++ b/src/sensors/zjs_sensor_accel.c
@@ -1,7 +1,5 @@
 // Copyright (c) 2017, Intel Corporation.
 
-#ifdef BUILD_MODULE_SENSOR_ACCEL
-
 #include <string.h>
 #include "zjs_common.h"
 #include "zjs_util.h"
@@ -27,16 +25,6 @@ static void onchange(void *h, const void *argv)
     zjs_sensor_trigger_change(obj);
 }
 
-static void onstart(void *h, const void *argv)
-{
-    sensor_handle_t *handle = (sensor_handle_t *)h;
-    jerry_value_t obj = handle->sensor_obj;
-
-    if (jerry_value_has_error_flag(zjs_sensor_start_sensor(obj))) {
-        zjs_sensor_trigger_error(obj, "SensorError", "start failed");
-    }
-}
-
 static void onstop(void *h, const void *argv)
 {
     sensor_handle_t *handle = (sensor_handle_t *)h;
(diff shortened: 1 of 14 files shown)


src/sensors/zjs_sensor_accel.c
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2017, Intel Corporation.
+
+#ifdef BUILD_MODULE_SENSOR_ACCEL
```
grgustaf: Check whether this is really needed here now or not; I was able to remove it for GPIO because our build is now smart enough not to even build the file at all if it's not used. (Same with others.)

src/sensors/zjs_sensor_accel.c
```diff
@@ -0,0 +1,134 @@
...
+    ZVAL global_obj = jerry_get_global_object();
+    zjs_obj_add_function(global_obj, zjs_sensor_constructor, "Accelerometer");
+    g_instance = zjs_malloc(sizeof(sensor_instance_t));
+    if (!g_instance) {
+        ERR_PRINT("failed to allocate sensor_instance_t\n");
+        return NULL;
+    }
+    memset(g_instance, 0, sizeof(sensor_instance_t));
+    return g_instance;
+}
```
grgustaf: I don't like so much code being duplicated in four files. Maybe some of this boilerplate at least could be implemented by a helper function in sensor.c and you call e.g. zjs_sensor_generic_init("Accelerometer") or something.

There are only 14 lines of differences between accel.c and gyro.c, only two of them are "real".

I guess this doesn't have to be dealt with right now, but I don't think we should leave it like this.

jimmy-huang: Yeah, I think I can simplify it a little more

src/zjs_sensor.h
```diff
@@ -1,9 +1,117 @@
...
+    jerry_value_t sensor_obj;
+    zjs_callback_id onchange_cb_id;
+    zjs_callback_id onstart_cb_id;
+    zjs_callback_id onstop_cb_id;
+    struct sensor_handle *next;
+} sensor_handle_t;
+
+typedef struct sensor_instance {
+    void *handles;
+} sensor_instance_t;
```
jimmy-huang: That's how I started with just returning the pointer, then it seems that I would need to store more stuff later, things like supported controllers, and maybe user_data, that's why I created the struct.

grgustaf: OK, I don't mind leaving it.

jimmy-huang:
@grgustaf updated patch

grgustaf:
+1, nice job

PR #1058, opened 2017-04-24
[promise] Updated Jerryscript/ZJS to use promises

jprestwo:
- Updated get/set native handle API's
 - ZJS promise API's use JrS built in promises


jprestwo:
Depends on: https://github.com/01org/zephyr.js/pull/1059

docs/uart.md
```diff
@@ -39,9 +43,8 @@ dictionary UARTOptions {
 
 [NoInterfaceObject]
 interface UARTConnection {
-    Promise<void> init(UARTOptions options);
     // void close();
-    Promise<void> write(Buffer data);
+    void write(Buffer data);
```
zolkis: But now UART just became a synchronous API. Is that warranted?
((The same question applied to the other changes as well).

zolkis:
A general question, since UART became a synchronous API.
Is it more convenient for developers to use synchronous APIs for low level board functions (involving local HW), and use Promises for network functions etc?

Or is it too much overhead to have asynchronous signature (using Promise) and also implement async/await in JerryScript?

pfalcon:
> But now UART just became a synchronous API. Is that warranted?

With the usual disclaimer that JS is not my preferred language, I'm surprised to see such change too. It reminds me a situation with Python's asyncio, where Stream's read() method is an async coroutine, while write() asymmetrically a normal function. This requires write() to either block (can't do that in async code), or have a potentially infinite buffer to slurp anything being fed into it. For MicroPython, we stepped away from that and implemented the proper async write.

The same problem would apply here too IMHO.

> implement async/await in JerryScript?

Well, I heard that Python-style async/await is coming to JS, didn't hear about it coming to JerryScript, may be a while I guess. (Unless my watching the github project doesn't help and I miss a lot; didn't hear about even generators coming.)

jprestwo:
@zolkis, I talked with @poussa and we both agreed that we should take out promises from these low level hardware API's. The overhead is too high for such simple modules. Now this is not to say that we are now requiring these API's to be completely synchronous. GPIO/UART read events are still interrupt driven, making them asynchronous.

We could add a callback to UART.write(), but the underlying C API is synchronous, so we would be "faking" the async nature since its going to block anyways.

poussa:
Let's use promises where the operations are truly asynchronous, like in the networking APIs. The HW APIs are all synchronous and faking async APIs does not make sense.

zolkis:
> @zolkis, I talked with @poussa and we both agreed that we should take out promises from these low level hardware API's. The overhead is too high for such simple modules.

OK, that's a reason. Simplicity of usage would be another. 

> Now this is not to say that we are now requiring these API's to be completely synchronous. GPIO/UART read events are still interrupt driven, making them asynchronous.

Interrupts yes, but that was not the context. It's not about a theoretical label for being synchronous or asynchronous, but about analyzing the use cases and determine if a blocking API could be used or control should be given back until the operation finished.

>The HW APIs are all synchronous

They are not synchronous between the cores, for instance I2C, SPI etc. We can fake them to be synchronous for a number of reasons like given above. But if UART has a synchronous underlying C API, that's good enough reason to keep it synchronous.

In [iot-js-api](https://github.com/01org/iot-js-api/tree/master/board) I2C, SPI, UART are asynchronous (use Promises), and AIO, GPIO, PWM are synchronous.

I would prefer we are consistent on this. So do you want all Board APIs to be synchronous from now, or only UART, and then I2C and SPI continue to be asynchronous and Promise based?

zolkis:
OK, it seems the zjs implementation of I2C serializes the calls anyway, so then let's make all Board APIs synchronous without Promises. I will update iot-js-api.

src/zjs_ocf_client.c
```diff
@@ -930,19 +919,17 @@ static jerry_value_t ocf_get_platform_info(const jerry_value_t function_val,
     // args: device ide
     ZJS_VALIDATE_ARGS(Z_STRING);
     struct ocf_handler *h;
-    jerry_value_t promise = jerry_create_object();
 
     ZJS_GET_STRING(argv[0], device_id, OCF_MAX_DEVICE_ID_LEN + 1);
 
     struct client_resource *resource = find_resource_by_id(device_id);
     if (!resource) {
         ERR_PRINT("resource was not found: %s\n", device_id);
```
grgustaf: If you reuse "resource was not found" from below and use template "%s: %s", it will only save that string once.

src/zjs_ocf_client.c
```diff
@@ -1030,19 +1017,17 @@ static jerry_value_t ocf_get_device_info(const jerry_value_t function_val,
     ZJS_VALIDATE_ARGS(Z_STRING);
 
     struct ocf_handler *h;
-    jerry_value_t promise = jerry_create_object();
 
     ZJS_GET_STRING(argv[0], device_id, OCF_MAX_DEVICE_ID_LEN + 1);
 
     struct client_resource *resource = find_resource_by_id(device_id);
     if (!resource) {
         ERR_PRINT("resource was not found: %s\n", device_id);
```
grgustaf: same one here

grgustaf:
LGTM

PR #1059, opened 2017-04-24
[gpio/uart] Removed promise dependency from GPIO/UART

jprestwo:


jprestwo:
This is in preparation for the big promise patch/update #1058. With JrS promises built in, ashell overflows since GPIO and UART use promises. Since the use of promises in these modules was not required (the operations were sync anyways) we can go with a sync version. This patch needs to and should go in before #1058.

zolkis:
Are these operations synchronous on all boards?

jprestwo:
@zolkis, on all our supported boards they are. Every board I have come into contact with has had a synchronous "open" API for hardware modules. The only API that I can think that *may* be async (on some boards) and use a callback (in C) would be UART.write(), but again in our case even this is synchronous.

zolkis:
Thanks. That simplifies things a lot.

grgustaf:
+1

PR #1060, opened 2017-04-24
[build] JrS feature detection

jprestwo:
- Detect the use of Math/JSON/Promise and generate a feature
   file, similar to how we do it for Zephyr


jprestwo:
This is in preparation for #1058. It goes along with #1059 as well; so that a module that does use promises can enable them at compile time. Currently we use a file, jerry_feature.profile, to enable/disable certain feature but it was still a manual process, requiring you to edit the file before building. This patch enables generation of that feature file, similar to how we generate prj.conf. It greps the script for common JrS features (currently Math, JSON and Promise) and enables them in the newly generated jerry_feature.profile.

There seemed to also be a bug where, if the profile had changed between builds, JerryScript would NOT re-build the library. To work around this we create a backup of the last profile, then generate a new one. If the profiles don't match then we rebuild JerryScript to pick up the new profile changes.

As said before, this patch needs to go in before #1058. It is independent of #1059 but both need to go in before #1058.

brianjjones:
+1

scripts/analyze.sh
```diff
@@ -23,6 +23,12 @@ fi
 PRJFILE=prj.conf.tmp
 ARCPRJFILE=arc/prj.conf.tmp
 CONFFILE=zjs.conf.tmp
+FEATUREFILE=fragments/jerry_feature.profile
+
+if [ -f fragments/jerry_feature.profile ]; then
+    cp fragments/jerry_feature.profile fragments/jerry_feature.profile.bak
+fi
+cat fragments/jerry_feature.base > fragments/jerry_feature.profile
```
grgustaf: cp instead? Does this .profile get updated somewhere?

Makefile
```diff
@@ -54,7 +54,7 @@ BLE_ADDR ?= "none"
 # JerryScript options
 JERRY_BASE ?= $(ZJS_BASE)/deps/jerryscript
 EXT_JERRY_FLAGS ?=	-DENABLE_ALL_IN_ONE=ON \
-			-DFEATURE_PROFILE=$(ZJS_BASE)/jerry_feature.profile \
+			-DFEATURE_PROFILE=$(ZJS_BASE)/fragments/jerry_feature.profile \
```
grgustaf: I'd like to not be leaving generated files randomly all around the tree. The best place for them is outdir/. Also maybe this should be board-specific like outdir/$BOARD?

If not in outdir than FWIW I think root is better than hidden in a subdir.

Makefile
```diff
@@ -227,6 +227,12 @@ analyze: $(JS)
 	@if grep BUILD_MODULE_DGRAM src/Makefile; then \
 		echo "CONFIG_BLUETOOTH_DEVICE_NAME=\"$(DEVICE_NAME)\"" >> prj.conf.tmp; \
 	fi
+	@if [ -e fragments/jerry_feature.profile.bak ]; then \
+		if ! cmp fragments/jerry_feature.profile.bak fragments/jerry_feature.profile; \
+		then \
+			rm -f outdir/$(BOARD)/libjerry-core*.a; \
+		fi \
+	fi
```
grgustaf: What if there's no .bak, shouldn't we then also remove libjerry?

scripts/analyze.sh
```diff
@@ -81,6 +87,21 @@ function check_config_file()
     return 1;
 }
 
+function check_for_feature()
+{
+    # effects: checks script for occurence of a JerrScript build in feature
```
grgustaf: occur **r** ence, Jerr **y** Script

built-in?

grgustaf:
+1

PR #1061, opened 2017-04-24
[ashell] Remove boot.cfg if this is a newly flashed board

brianjjones:
If the board __TIMESTAMP__ is newer than the one in boot.cfg,
remove it.  This allows users to get their ashell build out of a
bad state should the javascript function they specified in boot.cfg
wedges the board on boot.

### DIFF
--- a/src/ashell/shell-state.c
+++ b/src/ashell/shell-state.c
@@ -646,7 +646,6 @@ int32_t ashell_help(char *buf)
         comms_printf("%8s %s\r\n", commands[t].cmd_name, commands[t].syntax);
     }
     //comms_println("TODO: Read help file per command!");
-    ashell_run_boot_cfg();
     return RET_OK;
 }
 
--- a/src/ashell/shell-state.c
+++ b/src/ashell/shell-state.c
@@ -63,6 +63,7 @@ const char MSG_IMMEDIATE_MODE[] =
 const char hex_prompt[] = "[HEX]\r\n";
 const char raw_prompt[] = ANSI_FG_YELLOW "RAW> " ANSI_FG_RESTORE;
 const char eval_prompt[] = ANSI_FG_GREEN "js> " ANSI_FG_RESTORE;
+const char *BUILD_TIMESTAMP = __DATE__ " " __TIME__ "\n";
 
 #define CMD_TRANSFER "transfer"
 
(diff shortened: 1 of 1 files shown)


pfalcon:
This is a good troubleshooting measure, but perhaps removing is too harsh. What about renaming?

brianjjones:
@pfalcon All boot.cfg contains is a string of which file to boot.  The user can easily reset this by calling 'boot myfile.js'.  This erasure would only occur if you made changes to ZJS, rebuilt it, and flashed your board.  In which case would users expect the newly flashed board to still boot the specified javascript? I'm open to suggestions, I'm just trying to keep users from getting in a bad state they can't recover from.

zolkis:
Sounds reasonable. If someone can flash zephyr, likely this is not a problem. If not done, the board may become unusable. If done, there is only a small inconvenience to remember. So the gain clearly outweighs the loss.

src/ashell/shell-state.c
```diff
@@ -594,8 +594,10 @@ int32_t ashell_set_bootcfg(char *buf)
         return RET_ERROR;
     }
 
-    ssize_t written = fs_write(file, buf, strlen(buf));
-    if (written == 0) {
+    ssize_t written = fs_write(file, __TIMESTAMP__, strlen(__TIMESTAMP__));
```
grgustaf: Good find, I didn't know about `__TIMESTAMP__`. However, it looks like this is the last modification time of the source file. So I think you would be better off with `__TIME__` which is apparently the current time when the preprocessor is running.

Also, I'd suggest you use the macro once at the top of the file like this:

```
const char *BUILD_TIMESTAMP = __DATE__ " " __TIME__ "\n";
```

This is just because gcc docs aren't clear on whether this is set once per source file or changes during the course of processing the source file. If the latter, then it's possible for you to hit just right where it would roll over to the next second between when you use it here and down below. Then you'd have a bug where it never matches and you'd erase boot.cfg every time they create it I think.

src/ashell/shell-state.c
```diff
@@ -658,12 +659,31 @@ void ashell_run_boot_cfg()
...
-        count = fs_read(file, data, size);
-        data[size] = '\0';
-        ashell_run_javascript(data);
+    // Check that there is something after the timestamp
+    if (size > tssize) {
+        char ts[tssize];
+        count = fs_read(file, ts, tssize);
+        if (count == tssize && strncmp(ts, __TIMESTAMP__, tssize) == 0) {
+            // skip the space
+            fs_seek(file, 1, FS_SEEK_CUR);
```
grgustaf: If you add the newline to BUILD_TIMESTAMP (or change it to a space) like my example above, then you don't need to skip a space here (or add it above).

src/ashell/shell-state.c
grgustaf: I wasn't too clear: w/ __TIMESTAMP__, I think if you don't touch shell-state.c but rebuild ashell to get other changes, the timestamp won't actually change and you won't remove the boot.cfg.

brianjjones:
@grgustaf Sounds good, added the changes.

grgustaf:
+1

PR #1065, opened 2017-04-25
[tests] Add weak pull-up check and weak pull-down check

cuiyanx:
**TOTAL:** 
a101: 82 of 82 passed
k64f: 104 of 105 passed

### DIFF
--- a/tests/test-a101-pins.js
+++ b/tests/test-a101-pins.js
@@ -26,18 +26,35 @@ for(var i = 0; i < GPIOPins.length; i++) {
     // IO6 and IO9 are defined but unusable as GPIOs currently
     if (pinName == "IO6" || pinName == "IO9") continue;
 
-    var pin = gpio.open({ pin: pins[pinName] });
-    var pinValue = pin.read();
-    assert(typeof pinValue == "boolean",
+    var pinA = gpio.open({ pin: pins[pinName], direction: "in" });
+    var pinAValue1 = pinA.read();
+    var ApullValue = (pinAValue1) ? " weak pull-up" : " weak pull-down";
+
+    assert(typeof pinAValue1 == "boolean",
+           "Arduino101Pins: " + pinName + ApullValue + " on input");
+
+    pinA.write(false);
+    var pinAValue2 = pinA.read();
+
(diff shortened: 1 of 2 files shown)


tests/test-a101-pins.js
```diff
@@ -26,18 +26,35 @@ for(var i = 0; i < GPIOPins.length; i++) {
...
+    var ApullValue = (pinAValue1) ? " weak pull-up" : " weak pull-down";
+
+    assert(typeof pinAValue1 == "boolean",
+           "Arduino101Pins: " + pinName + ApullValue + " on input");
+
+    pinA.write(false);
+    var pinAValue2 = pinA.read();
+
+    assert(typeof pinAValue2 == "boolean" && pinAValue1 == pinAValue2,
            "Arduino101Pins: " + pinName + " input");
```
grgustaf: Can you explain what you're trying to test here? The previous assert you were happy whether it was true or false, just checking that it was a boolean. So are you trying to make sure that value doesn't change, that you weren't allowed to write to the input?

I guess that might be worthwhile... but if pinAValue1 was false you wouldn't really be testing anything by writing 'false' to the pin. In that case you should write true. So maybe you should write !pinAValue1?

grgustaf: Same issue in the other cases below I think.

cuiyanx: @grgustaf If input pin, pin value can not be changed after writing value. Otherwise it must be output pin. So, I write value that is different of default pin value. Usually we set weak pull up on input pins and weak pull down on output pins. Of course, I've seen different settings(like stm32 board). If set weak pull up on input pins, the pin default value must be 'true'. So, I write a fixed value as 'false' to check pin value unchanged.
    In this case, I find that default value of IO3 IO5 IO13 pins are set as 'false' different of other GPIO pins on input pin. Why? PLZ tell me.

grgustaf: OK, I think that makes sense. I have no idea why at this point. When I tried to set pullup/pulldown using Zephyr flags it never seemed to do anything. Maybe I need to try again. It could have something to do with pinmux settings.

PR #1069, opened 2017-04-26
Mock gpio module to run tests on Linux

grgustaf:
This patchset adds a ZJS_GPIO_MOCK #ifdef that controls whether the GPIO module uses Zephyr APIs or replaces them with mock implementations. In the latter case, it also adds a gpio.mock === True to the JS API and a wire(pin1, pin2) function that lets you simulate a wire connection from a JS test.

This lets you do hardware-dependent testing like our test-gpio.js completely in software, while running every line of our GPIO module code w/o modification.

Warning: This contains some black magic. In particular, I substituted a jerry_value_t for an opaque struct device *. I can probably come up with a cleaner way to do that, but first wanted to just get this out there for comment to see whether it's completely insane to go down this path. :)

My biggest concern is that this roughly doubled the code size of the GPIO module, which means it's roughly as likely to contain bugs as the the code itself. So now when we change the module, we may have double the maintenance burden, etc. And this was just what was needed to accomodate the test-gpio.js test case. More sophisticated mocking from JS might be desired. But it's at least an interesting experiment and something to start discussion more concretely.

I'd really like feedback!

### DIFF
--- a/src/zjs_gpio.c
+++ b/src/zjs_gpio.c
@@ -12,6 +12,34 @@
 #include "zjs_util.h"
 #include "zjs_callbacks.h"
 
+#ifdef ZJS_GPIO_MOCK
+static struct device *mock_device_get_binding(const char *name);
+static int mock_gpio_pin_read(struct device *port, uint32_t pin,
+                              uint32_t *value);
+static int mock_gpio_pin_write(struct device *port, uint32_t pin,
+                               uint32_t value);
+static int mock_gpio_pin_configure(struct device *port, uint8_t pin, int flags);
+static void mock_gpio_init_callback(struct gpio_callback *callback,
+                                    gpio_callback_handler_t handler,
+                                    uint32_t pin_mask);
+static int mock_gpio_add_callback(struct device *port,
+                                  struct gpio_callback *callback);
+static int mock_gpio_remove_callback(struct device *port,
+                                     struct gpio_callback *callback);
+static int mock_gpio_pin_enable_callback(struct device *port, uint32_t pin);
+
+#define device_get_binding mock_device_get_binding
+#define gpio_pin_read mock_gpio_pin_read
+#define gpio_pin_write mock_gpio_pin_write
+#define gpio_pin_configure mock_gpio_pin_configure
+#define gpio_init_callback mock_gpio_init_callback
+#define gpio_add_callback mock_gpio_add_callback
+#define gpio_remove_callback mock_gpio_remove_callback
+#define gpio_pin_enable_callback mock_gpio_pin_enable_callback
+
+static jerry_value_t mock_root_obj = 0;
+#endif
+
 static const char *ZJS_DIR_IN = "in";
 static const char *ZJS_DIR_OUT = "out";
 
@@ -168,7 +196,7 @@ static ZJS_DECL_FUNC(zjs_gpio_pin_read)
     bool activeLow = false;
     zjs_obj_get_boolean(this, "activeLow", &activeLow);
 
-    uint32_t value;
+    uint32_t value = 0;
     int rval = gpio_pin_read(gpiodev, newpin, &value);
     if (rval) {
         ERR_PRINT("PIN: #%d\n", newpin);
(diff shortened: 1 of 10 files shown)


grgustaf:
I also didn't clean up the mock metadata I created, and it may not handle the existing metadata properly if the tests get a lot fancier; but since it is only for testing purposes that didn't seem like a critical concern.

src/zjs_gpio.c
```diff
@@ -420,11 +490,398 @@ static ZJS_DECL_FUNC(zjs_gpio_open_async)
...
+    struct value_match_data *data = (struct value_match_data *)user_data;
+    if (value == data->value) {
+        // found match, stop processing and return name
+        data->name = name;
+        return false;
+    }
+    return true;
+}
+
+static jerry_value_t find_property_with_value(jerry_value_t obj,
```
jprestwo: This would be a good function for zjs_util as well as push_array(). Keep it here for now but I think OCF could use it.

src/zjs_gpio.c
```diff
@@ -420,11 +490,398 @@ static ZJS_DECL_FUNC(zjs_gpio_open_async)
...
+        return 0;
+    }
+
+    // check for a pin this is wired to
+    ZVAL wired = zjs_get_property(pin_obj, "wired");
+    if (jerry_value_is_array(wired)) {
+        uint32_t len = jerry_get_array_length(wired);
+        for (uint32_t i = 0; i < len; ++i) {
+            ZVAL pin = jerry_get_property_by_index(wired, i);
+            if (zjs_obj_get_boolean(pin, "state", &flag)) {
```
jprestwo: I am not sure what would *actually* happen, but if you had multiple pins wired to one, and you read that pin wouldn't it be high if any of the pins were high. This just takes the first value found.

src/zjs_gpio.c
```diff
@@ -420,11 +490,398 @@ static ZJS_DECL_FUNC(zjs_gpio_open_async)
...
+typedef struct mock_cb_item {
+    struct mock_cb_item *next;
+    struct gpio_callback *callback;
+    gpio_callback_handler_t handler;
+    uint32_t pin_mask;
+    uint32_t enabled_mask;
+    jerry_value_t port;
+} mock_cb_item_t;
+
+// FIXME: not bothering to free these yet
```
jprestwo: Might as well just free them before you push this; In the cleanup routine.

src/zjs_gpio.c
grgustaf: It's invalid wiring if you have multiple pins driving different voltages; in that case you'd get current flowing and a voltage somewhere between "0" and "1", so it would be undefined basically. So basically, not our problem here.

src/zjs_gpio.c
grgustaf: Thanks, addressed in an additional patch.

src/zjs_gpio.c
```diff
@@ -13,6 +15,74 @@
...
+#if defined(ZJS_LINUX_BUILD) || defined(QEMU_BUILD)
+#define ZJS_GPIO_MOCK
+#endif
+
+#ifdef ZJS_GPIO_MOCK
+#ifdef ZJS_LINUX_BUILD
+// provide Zephyr dependencies
+struct device;
+struct gpio_callback;
+typedef void (*gpio_callback_handler_t)(struct device *port,
```
grgustaf: Good idea, this actually let me remove the evil casting between struct device * and jerry_value_t.

src/zjs_gpio.c
```diff
@@ -13,6 +15,73 @@
 #include "zjs_callbacks.h"
 #include "zjs_promise.h"
 
+#if defined(ZJS_LINUX_BUILD) || defined(QEMU_BUILD)
+#ifndef ZJS_GPIO_MOCK
+#define ZJS_GPIO_MOCK
```
pfalcon: It would depend on the definition of "QEMU_BUILD" (there're different hw targets for QEMU for example), but I'd consider it too coarse a logic for "if it's qemu, use mock GPIO". A QEMU target may mock GPIO on its own for example.

Perhaps, use numeric values for ZJS_GPIO_MOCK, and #if (not #ifdef) to test it?

pfalcon:
Per @grgustaf ping on IRC:

I'd generally consider it a very good idea to add "dummy" GPIO implementation for Linux/QEMU which would e.g. just print to console "GPIO xxx is ON", "GPIO xxx is OFF". That would allow to to run blinky against it, total bliss ;-).

If there's going beyond that, towards mock implementation with behavior setup/record - well, only cooler of course, though the risks mentioned: 1) needs to be written (done already, but as mentioned above, with all the effort went into C impl, it can do only so much, with desire to mock in JS still on the horizon); 2) needs to be maintained. Regarding the last, I'd personally put this new mock code in a separate file, like zjs_gpio_mock.c (and zjs_gpio_mock.h).

Overall, looks great, will improve test coverage noticeably (taking into account that GPIO is the most basic HW feature).

grgustaf:
@pfalcon yeah I agree that splitting it out would be cleaner, I'll see how well I can do that.

src/zjs_gpio.c
grgustaf: I'll just drop that for now, don't have it working anyway.

src/zjs_gpio_mock.c
```diff
@@ -0,0 +1,403 @@
...
+    zjs_set_property(pin_obj, "pin", jerry_create_number(pin));
+
+    return 0;
+}
+
+void mock_gpio_init_callback(struct gpio_callback *callback,
+                             gpio_callback_handler_t handler,
+                             uint32_t pin_mask)
+{
+    mock_cb_item_t *item = (mock_cb_item_t *)malloc(sizeof(mock_cb_item_t));
```
jimmy-huang: check malloc fail

src/zjs_gpio_mock.h
```diff
@@ -0,0 +1,77 @@
...
+
+DEVICE mock_gpio_device_get_binding(const char *name);
+int mock_gpio_pin_read(DEVICE port, uint32_t pin, uint32_t *value);
+int mock_gpio_pin_write(DEVICE port, uint32_t pin, uint32_t value);
+int mock_gpio_pin_configure(DEVICE port, uint8_t pin, int flags);
+void mock_gpio_init_callback(struct gpio_callback *callback,
+                                    gpio_callback_handler_t handler,
+                                    uint32_t pin_mask);
+int mock_gpio_add_callback(DEVICE port, struct gpio_callback *callback);
+int mock_gpio_remove_callback(DEVICE port,
```
jimmy-huang: align indent

src/zjs_gpio_mock.h
```diff
@@ -0,0 +1,77 @@
...
+#define gpio_add_callback mock_gpio_add_callback
+#define gpio_remove_callback mock_gpio_remove_callback
+#define gpio_pin_enable_callback mock_gpio_pin_enable_callback
+
+DEVICE mock_gpio_device_get_binding(const char *name);
+int mock_gpio_pin_read(DEVICE port, uint32_t pin, uint32_t *value);
+int mock_gpio_pin_write(DEVICE port, uint32_t pin, uint32_t value);
+int mock_gpio_pin_configure(DEVICE port, uint8_t pin, int flags);
+void mock_gpio_init_callback(struct gpio_callback *callback,
+                                    gpio_callback_handler_t handler,
```
jimmy-huang: align indent

tests/test-gpio.js
```diff
@@ -21,7 +26,7 @@ var mark = 0;
 
 var edgeInterval = setInterval(function () {
     var count = 0;
-    pinA = gpio.open({ pin: pins.IO7 });
+    pinA = gpio.open({pin: pins.IO7});
```
jimmy-huang: i actually prefer the leaving the spaces before, is there a reason to change this?

grgustaf: Just following this since I noticed it:

https://google.github.io/styleguide/javascriptguide.xml#Code_formatting
`var obj = {a: 1, b: 2, c: 3};  // No space after { or before }.`

jimmy-huang:
+1

PR #1073, opened 2017-04-26
[general] Linked list macros to simplify code

jprestwo:

### DIFF
--- a/src/zjs_util.h
+++ b/src/zjs_util.h
@@ -298,7 +298,24 @@ void zjs_loop_init(void);
 #define UNLOCK() do {} while (0)
 #endif
 
-#define ZJS_FIND_NODE(type, list, cmp_element, cmp_to) \
+// Type definition to be used with macros below
+// struct list_item {
+//     int value;
+//     struct list_item *next;
+// } list_item_t;
+// list_item_t *my_list = NULL;
+
+// Looks for and returns a node found in a list.
+// @param type          Struct type of list
+// @param list          Pointer to list head
+// @param cmp_element   Name of element in node struct to compare
+// @param cmp_to        Value to compare node element to
+
+// Example:
+//    list_item_t *item = ZJS_FIND_NODE(list_item_t, my_list, value, 42);
+
+// The above will return a list item whos `value` parameter == 42
+#define ZJS_LIST_FIND(type, list, cmp_element, cmp_to) \
    ({ \
        type *found = NULL; \
        type *cur = list; \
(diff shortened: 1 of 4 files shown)


src/zjs_util.h
```diff
@@ -275,4 +275,66 @@ int zjs_validate_args(const char *expectations[], const jerry_length_t argc,
...
+       found; \
+   })
+
+#define ZJS_APPEND_NODE(type, list, p) \
+    { \
+        type **pnext = &list; \
+        while (*pnext != NULL) { \
+            pnext = &(*pnext)->next; \
+        } \
+    *pnext = p; \
```
grgustaf: indentation

src/zjs_util.h
```diff
@@ -275,4 +275,66 @@ int zjs_validate_args(const char *expectations[], const jerry_length_t argc,
...
+           } \
+           cur = cur->next; \
+       } \
+       found; \
+   })
+
+#define ZJS_APPEND_NODE(type, list, p) \
+    { \
+        type **pnext = &list; \
+        while (*pnext != NULL) { \
```
grgustaf: Maybe drop the != NULL?

src/zjs_util.h
```diff
@@ -275,4 +275,66 @@ int zjs_validate_args(const char *expectations[], const jerry_length_t argc,
         (zjs_validate_args((const char *[]){ __VA_ARGS__, NULL }, argc, argv) \
                 <= ZJS_INVALID_ARG) ? 1 : 0
 
+#define ZJS_FIND_NODE(type, list, cmp_element, cmp_to) \
```
grgustaf: How about add a comment block above defining a sample struct like:

```c
// Type definition for examples below:
// struct mylistitem {
//     struct mylistitem *next;
//     int myvalue;
// } mylistitem_t;
// mylistitem_t *mylist = NULL;
```

Variable names like (type, list, field_name, value) might be clearer?

And then above each macro show a one-line example like:

```c
/**
 * Returns node from list with field field_name matching value, or NULL
 *
 * (Uses simple equality, will not work for string fields.)
 *
 * Example:
 *   mylistitem_t *item = ZJS_FIND_NODE(mylistitem_t, mylist, myvalue, 42);
*/
```

src/zjs_util.h
```diff
@@ -275,4 +275,66 @@ int zjs_validate_args(const char *expectations[], const jerry_length_t argc,
...
+    { \
+        type **pnext = &list; \
+        while (*pnext != NULL) { \
+            pnext = &(*pnext)->next; \
+        } \
+    *pnext = p; \
+    }
+#define ZJS_PREPEND_NODE(type, list, p) \
+    { \
+        ((type *)p)->next = list; \
```
grgustaf: You shouldn't need to cast to (type *) and it's dangerous if they pass the wrong thing.

src/zjs_util.h
```diff
@@ -275,4 +275,66 @@ int zjs_validate_args(const char *expectations[], const jerry_length_t argc,
...
+        list = p; \
+    }
+
+#define ZJS_REMOVE_NODE(type, list, p) \
```
grgustaf: Add a note that this doesn't free the item, they should do that after removing it.

grgustaf: Probably this should return the removed node so they can tell the difference between removing and NULL (not found).

src/zjs_util.h
```diff
@@ -275,4 +275,66 @@ int zjs_validate_args(const char *expectations[], const jerry_length_t argc,
...
+        while (cur->next) { \
+            if (cur->next == p) { \
+                cur->next = ((type *)p)->next; \
+                break; \
+            } \
+            cur = cur->next; \
+        } \
+    }
+
+#define ZJS_FREE_LIST(type, list, callback) \
```
grgustaf: Change all functions to start with ZJS_LIST prefix, you can then probably leave off _NODE.

Probably this should just call zjs_free directly instead of a callback? I assume that's your real use case?

Otherwise it's more like FOREACH except that it sets the list to NULL at the end. Maybe a FOREACH would be useful.

src/zjs_util.h
jprestwo: My reasoning for this was kinda dumb, but for the unit test code I just create list items on the stack, so I am passing in &item, which wont work, since you cant do &item->next. This is why I did the casting. Probably better just to do a malloc in the unit tests code, OR just create a pointer that points to &item and pass that.

src/zjs_util.h
jprestwo: The reason for the callback is that there may be more than just the node that needs to be freed, jerry_value's or other pointers in the structure. We could just rename this to ZJS_LIST_FOREACH.

grgustaf:
+1

PR #1076, opened 2017-04-26
[nRF52] Added experimental support for nRF52

jprestwo:
- analyze was missing a case statement.
 - Zephyr BLE config options should be the same as A101

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ the BOARD name:
 make JS=samples/OcfServer.js BOARD=nrf52840_pca10056
 ```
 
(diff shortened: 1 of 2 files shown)


pfalcon:
+1. It's definitely nice to see support for more boards, though this one is a bit of cheating, as it reuses the config for an existing board ;-). In general case, config (configs - for networked vs non-networked, ashell vs non-ashell builds) is different, and that's what so far very hard to handle in Z.js.

brianjjones:
+1 LGTM

grgustaf:
+1

PR #1079, opened 2017-04-27
[sensor] Remove onstatechange event callback

jimmy-huang:
Updating implementation and samples due to change of latest W3C spec
no longer has onstatechange.

### DIFF
--- a/src/zjs_sensor.c
+++ b/src/zjs_sensor.c
@@ -174,7 +174,7 @@ static void zjs_sensor_set_state(jerry_value_t obj, enum sensor_state state)
         return;
     }
 
-    // update state property and trigger onstatechange event
+    // update state property and trigger onactivate event if neccessary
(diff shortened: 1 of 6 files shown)


src/zjs_sensor.c
```diff
@@ -174,7 +174,7 @@ static void zjs_sensor_set_state(jerry_value_t obj, enum sensor_state state)
         return;
     }
 
-    // update state property and trigger onstatechange event
+    // update state property and trigger onactivate event if neccessary
```
grgustaf: necessary

Where does it get triggered though, I don't see that code?

jimmy-huang: it's already triggering it in the old code, that's why it didn't show up in the diff, at the end of the func it does that

grgustaf:
+1

PR #1080, opened 2017-04-28
[ashell] Fix ashell after it broke with semaphore change

jprestwo:


brianjjones:
Tested and it works for me, +1

jimmy-huang:
+1

PR #1081, opened 2017-04-28
[boards] New board support/docs

jprestwo:
- Added 'flash' target to make to support BOARD=nucleo_f411re
 - Moved around analyze.sh lines for UART/FS for possible
   support on other targets
 - Added more info in README about supported boards and features

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,17 @@ else
 JERRYLIB=outdir/$(BOARD)/libjerry-core-parser.a
 endif
 
+.PHONY: flash
+flash:  analyze generate $(JERRYLIB) $(ARC)
+	@make -f Makefile.zephyr flash	BOARD=$(BOARD) \
+									VARIANT=$(VARIANT) \
+									CB_STATS=$(CB_STATS) \
+									PRINT_FLOAT=$(PRINT_FLOAT) \
+									SNAPSHOT=$(SNAPSHOT) \
+									BLE_ADDR=$(BLE_ADDR) \
+									ASHELL=$(ASHELL) \
+									NETWORK_BUILD=$(NET_BUILD)
+
 # Build for zephyr, default target
 .PHONY: zephyr
 zephyr: analyze generate $(JERRYLIB) $(ARC)
--- a/Makefile
+++ b/Makefile
@@ -158,13 +158,13 @@ endif
 .PHONY: flash
 flash:  analyze generate $(JERRYLIB) $(ARC)
 	@make -f Makefile.zephyr flash	BOARD=$(BOARD) \
-									VARIANT=$(VARIANT) \
-									CB_STATS=$(CB_STATS) \
-									PRINT_FLOAT=$(PRINT_FLOAT) \
-									SNAPSHOT=$(SNAPSHOT) \
-									BLE_ADDR=$(BLE_ADDR) \
-									ASHELL=$(ASHELL) \
-									NETWORK_BUILD=$(NET_BUILD)
+		VARIANT=$(VARIANT) \
+		CB_STATS=$(CB_STATS) \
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -633,23 +633,35 @@ For additional information, see [here](https://www.zephyrproject.org/doc/getting
...
+
+* X - Supported
+* E - Experimental, not formally tested by QA but basic functionality verified.
+* NT - Not tested at all
+* Blank - Not Supported
+
+| Module    | Linux | Arduino 101 | K64F  | nRF52 | Arduino DUE | ST F411RE |
+| :---:     | :---: |    :---:    | :---: | :---: |    :---:    |   :---:   |
+|HelloWorld |   X   |      X      |   X   |    X  |      X      |     X     |
+|  ADC      |       |      X      |   X   |   NT  |      NT     |     NT    |
```
jimmy-huang: I thought ADC doesn't not work on k64f yet as there's no ADC driver in Zephyr?

README.md
```diff
@@ -633,23 +633,35 @@ For additional information, see [here](https://www.zephyrproject.org/doc/getting
...
+|  PWM      |       |      X      |   X   |   NT  |      NT     |     NT    |
```
grgustaf: Jimmy's right, I don't think we've ever tested PWM on K64F either. That should work but might require some pinmux changes.

jimmy-huang: well, i am pretty sure ADC won't work for k64f, because the AIO module solely uses IPM, and we never implemented ADC to use non-ipm calls

Makefile
```diff
@@ -155,6 +155,17 @@ else
 JERRYLIB=outdir/$(BOARD)/libjerry-core-parser.a
 endif
 
+.PHONY: flash
+flash:  analyze generate $(JERRYLIB) $(ARC)
+	@make -f Makefile.zephyr flash	BOARD=$(BOARD) \
+									VARIANT=$(VARIANT) \
```
pfalcon: For the *standard* tab size of 8 this looks hugely right-indented (easily visible on github).

README.md
```diff
@@ -633,23 +633,35 @@ For additional information, see [here](https://www.zephyrproject.org/doc/getting
...
+successfully. Below is a complete table of modules and target support. Some
+Zephyr targets listed are experimental and have not been fully tested. For this
+reason we have the following possibilities for support:
+
+* X - Supported
+* E - Experimental, not formally tested by QA but basic functionality verified.
+* NT - Not tested at all
+* Blank - Not Supported
+
+| Module    | Linux | Arduino 101 | K64F  | nRF52 | Arduino DUE | ST F411RE |
```
pfalcon: I'd say table format won't scale for a long time, as more targets are added ;-).

pfalcon:
+1 for any progress to support more boards, but stuff like:

~~~
     if [ $BOARD = "arduino_101" ]; then
         echo "CONFIG_FS_FAT_FLASH_DISK_W25QXXDV=y" >> $PRJFILE
+        echo "CONFIG_DISK_ACCESS_FLASH=y" >> $PRJFILE
+    else
+        echo "CONFIG_DISK_ACCESS_RAM=y" >> $PRJFILE
     fi
~~~

won't scale, there will be maze of if's soon (well, arguably, there already is). The only solution is to move different axes of configuration (per-board, network vs non-network, etc.) to config fragments. The sooner we get there, the better ;-).

grgustaf:
@pfalcon, see #1104 for a big step in that direction...

README.md
grgustaf: @jimmy-huang, It will because you're going to have to rewrite that junk as soon as come up against a working second platform. :)

grgustaf:
+1

PR #1084, opened 2017-05-02
[ocf] Added start API to ocf object

jprestwo:
- Now, multiple resources can be registered


jimmy-huang:
+1

PR #1085, opened 2017-05-02
[build] Fixed analyze for ARC

jimmy-huang:
Some of the analyze scripts should belong to the ARC, also adding -q to
suppress output of grep so it doesn't show up multiple times during
build


grgustaf:
+1

PR #1086, opened 2017-05-02
[ashell] Remove error print if a cfg file doesn't exist

brianjjones:


jimmy-huang:
+1

PR #1087, opened 2017-05-02
[sensor] Support polling mode for Ashell

jimmy-huang:
Polling mode is enabled by default when GPIO module is
enabled, instead of using triggers to monitor change events,
the ARC will poll the sensor data, this patch will allow
Ashell to have both GPIO and Sensor enabled.


grgustaf:
+1

PR #1093, opened 2017-05-03
[net] Add null pointer check to tcp_recv

### LINKED ISSUE #1091: [net] BUS FAULT occurs when break connection with netcat
### Description

BUS FAULT occurs when break connection with netcat
 
### Test Code

[TCPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ6.js); 
 
### Steps to Reproduction

1. `make BOARD=frdm_k64f JS=samples/TCPEchoServ6.js`
2.  cp outdir/frdm_k64f/zephyr.bin /media/<USERNAME>/MBED/
3. Connect TCP server by `nc -6 2001:db8::1 4242` on Linux
4. Disconnect by `CTRL+C`

### Actual Result

BUS FAULT occurs when break connection with netcat
selection_049

### Expected Result

No error occurs.
 
### Test Builds

---|---|---|---|---
master | 78bc531 | FRDM-K64F | May 3, 2017 | Fail

jprestwo:
Fixes #1091

### DIFF
--- a/src/zjs_net.c
+++ b/src/zjs_net.c
@@ -259,13 +259,12 @@ static void tcp_received(struct net_context *context,
 {
     sock_handle_t *handle = (sock_handle_t *)user_data;
 
-    if (handle) {
+    if (handle && buf) {
         start_socket_timeout(handle, handle->timeout);
 
         uint32_t len = net_nbuf_appdatalen(buf);
         uint8_t *data = net_nbuf_appdata(buf);
 
-        DBG_PRINT("received data, context=%p, data=%p, len=%u\n", context, data, len);
         // TODO: Probably not the right way to check if a socket is closed
         if (len == 0 && data == NULL) {
             // socket close
@@ -274,10 +273,13 @@ static void tcp_received(struct net_context *context,
             jerry_value_clear_error_flag(&error);
             zjs_trigger_event(handle->socket, "error", &error, 1, NULL, NULL);
             zjs_trigger_event(handle->socket, "close", NULL, 0, post_closed, handle);
+            net_buf_unref(buf);
             return;
         }
 
         if (len && data) {
+            DBG_PRINT("received data, context=%p, data=%p, len=%u\n", context, data, len);
+
             memcpy(handle->wptr, data, len);
             handle->wptr += len;
 


src/zjs_net.c
```diff
@@ -259,13 +259,12 @@ static void tcp_received(struct net_context *context,
...
 
-    if (handle) {
+    if (handle && buf) {
         start_socket_timeout(handle, handle->timeout);
 
         uint32_t len = net_nbuf_appdatalen(buf);
         uint8_t *data = net_nbuf_appdata(buf);
 
-        DBG_PRINT("received data, context=%p, data=%p, len=%u\n", context, data, len);
         // TODO: Probably not the right way to check if a socket is closed
```
pfalcon: Indeed, that's not the right way. Incoming buf == NULL is exactly what signals EOF (peer closed connection to be exact).

jprestwo: Ok that makes sense. So what does if mean if buf is valid, but data == NULL and len == 0? When I wrote this and was testing, I would always get that case when I closed the other end of the connection.

pfalcon: I can only quote a comment from MicroPython implementation:

~~~
        // Keep repeating while we're getting empty fragments
        // Zephyr IP stack appears to feed empty net_buf's with empty
        // frags for various TCP control packets.
~~~

Note that this no longer seems to happen with Zephyr 1.8-pre.

jprestwo: Yeah that was exactly what I saw too. The buf was valid but there was no data inside. And no explanation to what it meant. I think you are right that buf == NULL means the socket was closed, but I would be curious to find out what the other cases mean.

I made the assumptions for websockets that if the data pointer inside the buffer was != NULL, but length was zero, that means it was a TCP ACK. This was purely based off trial and error.

Assuming:
`buf` = argument passed into read callback
`data` = net_nbuf_appdata(buf)
`len` = net_nbuf_appdatalen(buf)

There are 4 cases I have seen:
1. `buf` == NULL
     Means socket has been closed
2. `buf` != NULL, `data` != NULL, `len` == 0
     ??? (Maybe an ACK, or some other control packet)
3. `buf` != NULL, `data` == NULL, `len` == 0
    ???
4. `buf` != NULL, `data` != NULL, `len` != 0
    Data received

grgustaf:
@jprestwo can you open an issue for the remaining concerns, but it seems like we should get this patch in.

grgustaf:
+1

src/zjs_net.c
pfalcon: @jprestwo : I don't think there's any meaning in 2,3,4, these are just artifacts of 1.7 implementation, as I mentioned, these are gone in Z master (to be 1.8). A lot of things changes in 1.8, the patch will be all red and green, and important thing to keep in mind, it's not just mechanical search&replace.

buf == NULL is officially documented though:

~~~
 * @param pkt Network buffer that is received. If the pkt is not NULL,
 * then the callback will own the buffer and it needs to to unref the pkt
 * as soon as it has finished working with it.  On EOF, pkt will be NULL.
 * @param status Value is set to 0 if some data or the connection is
 * at EOF, <0 if there was an error receiving data, in this case the
 * pkt parameter is set to NULL.
~~~

PR #1095, opened 2017-05-03
[sensor] Fix calling get_state calling into JerryScript from interrupt

jimmy-huang:
Fixes a previous patch that accidently call into JerryScript engine
from the IPM interrupt again since this is not safe.  Intead, store
state in the handle and query that instead.


grgustaf:
+1

PR #1097, opened 2017-05-04
Board api

grgustaf:


grgustaf:
@zolkis In case you want to take a look at what I've done so far.

zolkis:
Core functionality looks good. 

I need to think about a way to expose the pin mapping that is in the impl to developers as a doc (would that be zephyr docs?). I will perhaps remove the board docs from iot-js-api and replace with a single file with external links to various mappings.

Questions:
- which API do you implement? the latest changes in my [fork](https://github.com/zolkis/iot-js-api/tree/master/board), or the one in master?
- what do you say about supporting two namespaces for pin names in the API: OS (Zephyr pin mapping) and board (the labels printed on the board)? Apps can give hints which namespace they are using, and if not defined, first the OS namespace is checked, then the board namespace. 
- what do you say about the port support; in the API spec I assume there may be ports referred by symbolic name in boards that support it, or then a port would be a "manual" aggregation of pins by the impl (a convenience).

grgustaf:
@zolkis I've based on zolkis fork because I like the changes and don't want more rework. :)

What I wrote here was before you added specifying the namespace; that is, I "auto-detect" first scanning for a match among simple names (1, "IO1", etc...) and only if not found, I see if "GPIO_0.3" format is present. For now at least, the sets are disjoint so this works. But at some point there could perhaps be overlap in which case your solution is a good one. Basically until that day, if I ignore it, you won't be able to tell. :)

I haven't tried to implement the port support yet but I like the idea. I think I would use the real zephyr port API and so that will only work if all the pins are truly within a single hardware port. So I would try to check for that and fail if they try to create one that crosses ports. I probably won't have great feedback until I actually attempt to implement it. My priority is to migrate all existing functionality to use the new board-based pins and such, and try to fit your APIs (PWM, AIO, etc.)

zolkis:
All sounds good. 
As for namespaces, both could be checked anyway, just that the app provided one is checked first. Of course we could also implement a policy that fails if nothing found in the specified namespace. A config item I guess from impl point of view, but the spec should choose one of the behaviour. I'd say the second would be reasonable: if no namespace is explicitly specified, first check the OS namespace, then the board namespace, otherwise only check the provided namespace.

grgustaf:
Still plan to fix up documentation although this could perhaps be merged before that. Other than that, I think it's in good shape now.

grgustaf:
Realized there are another half dozen at least tests in tests/ that require updates to GPIO API calls. :(

samples/AutoButton.js
```diff
@@ -20,20 +20,12 @@ var delay = 10;
...
-    edge: 'rising'
-});
+var input = gpio.open({pin: "IO7", mode: 'in', edge: 'rising'});
 
 // simulate button presses on this output
-var output = gpio.open({
-    pin: pins.IO12,
-    direction: 'out'
-});
+var output = gpio.open({pin: "IO10"});
```
zolkis: When only pin is specified, the simple form `gpio.open("IO10")` also works, but of course you can choose to be explicit and use an init dictionary.

grgustaf: Yeah, I kind of varied the usage throughout the samples to perhaps highlight the different options.

src/zjs_board.c
```diff
@@ -0,0 +1,188 @@
...
+    {"PWM2", NULL,  NULL, PTC,  2,   3},
+    {"PWM3", NULL,  NULL, PTC,  3,   3},
+    {"PWM4", NULL,  NULL, PTC, 12,   4},
+    {"PWM5", NULL,  NULL, PTC,  4,   5},
+    {"PWM6", NULL,  NULL, PTD,  0,   6},
+    {"PWM7", NULL,  NULL, PTD,  2,   7},
+    {"PWM8", NULL,  NULL, PTD,  3,   8},
+    {"PWM9", NULL,  NULL, PTD,  1,   9}
+    // TODO: More pins at https://developer.mbed.org/platforms/FRDM-K64F/
+};
```
jprestwo: This pin mapping is going to get rather large as more targets are added. I liked the idea of keeping the pin maps in their own file.

brianjjones:
LGTM

src/zjs_board.c
grgustaf: That's true, but let's cross that bridge when we come to it. For now I want to move things out of the pins modules because those will be disappearing as require() modules at least (Not yet because pwm and aio still use them.) The way this board api works now, they will be implicitly defined given how you build with BOARD.

docs/board.md
```diff
@@ -0,0 +1,56 @@
...
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+The Board API provides information about the circuit board ZJS is running on.
+
+The Board API intends to follow the [iot-js-api specification](https://github.com/01org/iot-js-api/tree/master/board),
+but both that and ZJS are under a lot of change at the moment.
```
jimmy-huang: changes

docs/gpio.md
```diff
@@ -11,6 +11,9 @@ Introduction
 The GPIO API supports digital I/O pins. Pins can be configured as inputs or
 outputs, with some board-specific limitations.
 
+The GPIO API intends to follow the [iot-js-api specification](https://github.com/01org/iot-js-api/tree/master/board/gpio.md),
+but both that and ZJS are under a lot of change at the moment.
```
jimmy-huang: changes

docs/gpio.md
```diff
@@ -22,78 +25,88 @@ specific API functions.
...
 dictionary GPIOInit {
-    unsigned long pin;
+    number or string pin;
     boolean activeLow = false;
-    string direction = "out";  // in, out
-    string edge = "any";       // none, rising, falling, any
-    string pull = "none";      // none, up, down
+    string mode = "out";        // in, out
+    string edge = "none";       // none, rising, falling, any
+    string pull = "state";      // none, up, down
```
jimmy-huang: it says none, up, down, but default is "state"? should be state = "none"?

samples/Board.js
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2017, Intel Corporation.
+
+var board = require('board');
+
+console.log("Connected to board:", board.name);
```
jimmy-huang: maybe add a space after :

samples/ButtonLEDs.js
```diff
@@ -9,24 +9,22 @@ console.log("GPIO test with two LEDs and a button...");
...
 var gpio = require("gpio");
-var pins = require("arduino101_pins");
 
 // LED1 and LED2 are onboard LEDs on Arduino 101
-var pinA = gpio.open({pin: pins.LED1, activeLow: true});
-var pinB = gpio.open({pin: pins.LED2, activeLow: false});
-var pinIn = gpio.open({pin: pins.IO4, direction: 'in', edge: 'rising'});
+var pinA = gpio.open({pin: "LED1", activeLow: true});
+var pinB = gpio.open({pin: "LED2", activeLow: false});
+var pinIn = gpio.open({pin: "IO4", mode: 'in', edge: 'rising'});
```
jimmy-huang: make all double quotes or single quote

samples/Electroshock.js
```diff
@@ -37,8 +33,7 @@ function resetDebounce() {
 function updateBuzzer() {
     var value = button.read();
     if (value != lastState) {
```
jimmy-huang: !==

samples/OcfSensorServer.js
```diff
@@ -49,9 +49,10 @@ server.register(resourceInit).then(function(resource) {
         if (request.resource.properties) {
             var recvProps = request.resource.properties;
             if (recvProps.light != undefined) {
```
jimmy-huang: !==

samples/TrafficLight.js
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) 2016-2017, Intel Corporation.
 
-// Reimplementation of Arduino - Basics - Blink example
-//   - Toggles onboard LEDs as if they were a traffic light
+// Toggles onboard LEDs as if they were a traffic light
```
jimmy-huang: they were traffic lights?

samples/TwoButtons.js
```diff
@@ -11,16 +11,15 @@ console.log("GPIO test with two buttons controlling two LEDs...");
...
-var pins = require("arduino101_pins");
 
 // LED1 and LED2 are onboard LEDs on Arduino 101
-var led1 = gpio.open({pin: pins.LED1, activeLow: true});
-var led2 = gpio.open({pin: pins.LED2, activeLow: false});
-var btn1 = gpio.open({pin: pins.IO3, direction: 'in', edge: 'any'});
-var btn2 = gpio.open({pin: pins.IO4, direction: 'in', edge: 'any'});
+var led1 = gpio.open({pin: "LED1", activeLow: true});
+var led2 = gpio.open({pin: "LED2", activeLow: false});
+var btn1 = gpio.open({pin: 2, mode: 'in', edge: 'any'});
```
jimmy-huang: again, use double quote or single quote all around.

samples/arduino/basics/Blink-k64f.js
```diff
@@ -11,20 +11,16 @@ console.log("Starting Blink-k64f example...");
...
+var pin = gpio.open("LEDG");
 
 // remember the current state of the LED
-var toggle = false;
+var toggle = 0;
 
 // schedule a function to run every 1s (1000ms)
 setInterval(function () {
-    toggle = !toggle;
+    toggle = 1 -toggle;
```
jimmy-huang: space before toggle

samples/arduino/basics/Blink.js
```diff
@@ -18,21 +18,20 @@
...
 
 // import gpio module
 var gpio = require("gpio");
-var pins = require("arduino101_pins");
 
 // LED0 is one of the onboard LEDs on the Arduino 101
-// 'out' direction is default, could be left out
-var pin = gpio.open({pin: pins.LED0, direction: 'out', activeLow: true});
+// 'out' mode is default, could be left out
+var pin = gpio.open({pin: "LED0", mode: 'out', activeLow: true});
```
jimmy-huang: again, quotes

samples/arduino/digital/Button.js
```diff
@@ -14,10 +14,11 @@ console.log("Starting Button example...");
 
 // import gpio module
 var gpio = require("gpio");
-var pins = require("arduino101_pins");
 
-var led = gpio.open({pin: pins.LED0, direction: 'out', activeLow: true});
-var button = gpio.open({pin: pins.IO4, direction: 'in', edge: 'any'});
+var led = gpio.open({pin: "LED0", mode: 'out', activeLow: true});
+var button = gpio.open({pin: "IO4", mode: 'in', edge: 'any'});
```
jimmy-huang: quotes

samples/arduino/starterkit/Love-o-Meter.js
```diff
@@ -16,30 +16,18 @@ var aio = require('aio');
...
-var led2 = gpio.open({
-    pin: pins.IO7,
-    direction: 'out'
-});
-var led3 = gpio.open({
-    pin: pins.IO8,
-    direction: 'out'
-});
+var sensor = aio.open({device: 0, pin: pins.A0});
+var led1 = gpio.open("LED0");//4);
```
jimmy-huang: are these comments? maybe just do gpio.open("LED0"); // pin 4

samples/tests/ButtonEnable.js
```diff
@@ -15,22 +15,21 @@ console.log("Button enable test...");
...
 var gpio = require("gpio");
-var pins = require("arduino101_pins");
 
-var led1 = gpio.open({pin: pins.LED1, activeLow: true});
-var led2 = gpio.open({pin: pins.LED2, activeLow: false});
-var btn1 = gpio.open({pin: pins.IO2, direction: 'in', edge: 'any'});
-var btn2 = gpio.open({pin: pins.IO4, direction: 'in', edge: 'any'});
+var led1 = gpio.open({pin: "LED1", activeLow: true});
+var led2 = gpio.open({pin: "LED2", activeLow: false});
+var btn1 = gpio.open({pin: "IO2", mode: 'in', edge: 'any'});
```
jimmy-huang: quotes

scripts/analyze.sh
```diff
@@ -314,8 +314,10 @@ if check_for_require ocf || check_config_file ZJS_OCF; then
 fi
 
 if check_for_require gpio || check_config_file ZJS_GPIO; then
+    >&2 echo Using module: GPIO
     MODULES+=" -DBUILD_MODULE_GPIO"
     echo "CONFIG_GPIO=y" >> $PRJFILE
+    echo "export ZJS_BOARD=y" >> $CONFFILE
```
jimmy-huang: the check below also set export ZJS_BOARD=y, so you are setting it twice for GPIO?  not that it hurts anything.

scripts/trlite
```diff
@@ -446,7 +446,7 @@ if [ "$RUN" == "all" -o "$RUN" == "4" ]; then
     try_command "ashell" make $VERBOSE ashell ROM=250
 
     # build ide version
-    try_command "ide" make $VERBOSE ide ROM=250
+    try_command "ide" make $VERBOSE ide ROM=251
```
jimmy-huang: probably better make it 256, since the PME will increase to it  anyway,  I tested with the 0.9 SDK and the largest we can do is 259.

docs/gpio.md
grgustaf: yep, thanks :)

samples/Board.js
grgustaf: there's already a space due to the , which you dislike :)

docs/board.md
grgustaf: no, this is fine

samples/TrafficLight.js
grgustaf: no, this is fine.

samples/arduino/starterkit/Love-o-Meter.js
grgustaf: thanks, that was a tmp change for my own testing

scripts/analyze.sh
grgustaf: It shouldn't hurt anything and it is needed. If they require('board') they need board; if they require('gpio') they need board even if they didn't require('board'), but they don't need the JS board API part. This will need some rework at some point, perhaps the pin stuff should be separated from the board module. The board has become utterly trivial now, basically just a name. When I started this patch it was much bigger. :)

scripts/trlite
grgustaf: I merged your patch already so this will probably get taken care of in rebase. I'll try that in minute.

samples/Board.js
jimmy-huang: oh yeah, i keep thinking like printf formatting :)

jimmy-huang:
lgtm, +1

PR #1101, opened 2017-05-04
[ws] Forgot to check in websocket file

jprestwo:
- Added net and ws to trlite


grgustaf:
This will probably make VM1 take much longer than VM2 which will tend to be the long pole of Travis, but we can adjust it again 

+1

PR #1104, opened 2017-05-04
[analyze] Replace analyze with manifest based modules

jprestwo:


src/zjs_events.json
```diff
@@ -0,0 +1,7 @@
+{
+    "module":"events",
+    "require":"events",
+    "targets":["arduino_101", "frdm_k64f"],
+    "src": ["zjs_event.c"],
+    "zjs_config": ["-DBUILD_MODULE_EVENT"]
+}
```
grgustaf: You didn't mean to have both zjs_event.json and zjs_events.json.

Makefile
```diff
@@ -258,7 +255,7 @@ update:
 # set up prj.conf file
 -.PHONY: setup
 setup:
-	@echo "# This is a generated file" > prj.conf
+	@echo "# This is a generated file" >> prj.conf
```
grgustaf: This should be the top line of the file, so if it's not now, can you rearrange things so it will be?

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+
+find_deps(board)
+if force != ['']:
+    for i in force:
+        find_deps(i)
+
+#print("DEPENDENCIES:\n" + str(dependencies))
+#print("ZCONF:\n" + str(zephyr_config))
+#print("SRC:\n" + str(src_list))
+#print("ZJS:\n" + str(zjs_config))
```
grgustaf: Either remove or make conditional to some debug flag you set at the top maybe?

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+import json
+import re
+import shutil
+
+force = ['']
+featurefile = ''
+
+for i in sys.argv:
+    if i == sys.argv[0]:
+        continue
```
grgustaf: You can say `for i in sys.argv[1:]` above instead.

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+    option = i.split('=')[0]
+    value = i.split('=')[1]
```
grgustaf: `option, value = i.split('=')[:2]`

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+    elif option == 'PRJBASE':
+        prjbase = value
+    elif option == 'PRJCONF':
+        prjconf = value
+    elif option == 'MAKEFILE':
+        makefile = value
+    elif option == 'MAKEBASE':
+        makebase = value
+    elif option == 'PROFILE':
+        featurefile = value
```
grgustaf: Not required, but this could all be simplified with something like:

```python
args = {}   # also, set any defaults here
...
    option, value = ...
    args[option] = value
```

Then later refer to:
args['BOARD'] etc.

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+
+def add_to_list(list1, list2):
+    if list1 == None:
+        return list2
+    if list2 == None:
+        return list1
+    list1_set = set(list1)
+    list2_set = set(list2)
+    diff = list2_set - list1_set
+    return list1 + list(diff)
```
grgustaf: You might be doing this to preserve, order but I'm not sure you need to do that. You could replace it with this:
`return list(set(list1) | set(list2))`

And you might not even really need to convert to list, it's still iterable.

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+    global zjs_config
+    global jrs_config
+    with open("src/zjs_" + name + ".json") as f:
+        data = json.load(f)
+    if 'depends' in data:
+        depends = data['depends']
+        dependencies = add_to_list(dependencies, depends)
+        for i in depends:
+            l = find_deps(i)
+            dependencies = add_to_list(l, dependencies)
```
grgustaf: I think you're doing this twice... `find_deps` above is already going to have added these dependencies to your global.

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+    if 'targets' in data:
+        found = False
+        for i in data['targets']:
+            if i == board:
+                found = True
+                break
+        if not found:
+            return
```
grgustaf: This can be simplified:
```python
if 'targets' in data:
    if board not in data['targets']:
        return
```

or

```python
if board not in data.get('targets', []):
    return
```

src/zjs_buffer.json
```diff
@@ -0,0 +1,7 @@
+{
+    "module":"buffer",
+    "constructor":"Buffer",
+    "targets":["arduino_101", "frdm_k64f"],
```
grgustaf: I'm not sure if you can solve this now, but this breaks scalability because we have to list every supported platform. Perhaps if there is no "targets" it could mean all platforms should work?

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+        find_deps(i)
+
+#print("DEPENDENCIES:\n" + str(dependencies))
+#print("ZCONF:\n" + str(zephyr_config))
+#print("SRC:\n" + str(src_list))
+#print("ZJS:\n" + str(zjs_config))
+
+write_zconf(zephyr_config, prjconf)
+write_makefile(src_list, makefile)
+write_jrsconfig(jrs_config)
```
grgustaf: missing newline

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+        if not found:
+            return
+    if 'zephyr_conf' in data:
+        z_conf = data['zephyr_conf']
+        if 'all' in z_conf:
+            zephyr_config = add_to_list(zephyr_config, z_conf['all'])
+        if board in z_conf:
+            zephyr_config = add_to_list(zephyr_config, z_conf[board])
+    if 'zjs_config' in data:
+        zjs_config = add_to_list(zjs_config, data['zjs_config'])
```
grgustaf: You should probably make this edit the first list in place, rather than return it. Lists are mutable in Python. Really you haven't defined "add_to_list" (append) but "merge_lists".

If you do that, these two lines could become:
```python
add_to_list(zjs_config, data.get('zjs_config', []))
```

Same can be all over the place to trim things down.

scripts/analyze
```diff
@@ -0,0 +1,168 @@
...
+            if re.search(r"require\([\'\"]%s[\"\']\)" % data["require"], text):
+                find_deps(data["require"])
+        if 'constructor' in data:
+            if re.search(r"%s\(.*" % data["constructor"], text):
+                find_deps(data["module"])
+
+find_deps(board)
+if force != ['']:
+    for i in force:
+        find_deps(i)
```
grgustaf: Since lists are mutable, you could just pass dependencies into find_deps and not have to make it a global.

scripts/analyze.sh
```diff
@@ -1,560 +0,0 @@
...
-    feature_on CONFIG_DISABLE_ES2015_TYPEDARRAY_BUILTIN
-fi
-
-if check_for_feature "JSON\."; then
-    >&2 echo Using JrS module: JSON
-    feature_on CONFIG_DISABLE_JSON_BUILTIN
-fi
-
-echo "# ---------------------------" >> $CONFFILE
-echo $MODULES
```
grgustaf: Doesn't deleting feel good?

src/zjs_arduino101_pins.json
```diff
@@ -0,0 +1,7 @@
+{
+    "module":"arduino101_pins",
+    "require":"arduino101_pins",
+    "targets":["arduino_101"],
+    "zjs_config": ["-DBUILD_MODULE_A101"],
```
grgustaf: Try to be consistent with a space after `:`, hopefully some judicious sed could make the work easy.

grgustaf: This filename should match C: zjs_a101_pins.json

src/zjs_ashell.json
```diff
@@ -0,0 +1,39 @@
+{
```
grgustaf: I wonder if these "virtual" modules could at least go in their own dir maybe? Just keep the ones that parallel real source files?

src/zjs_frdm_k64f.json
```diff
@@ -0,0 +1,6 @@
+{
```
grgustaf: zjs_k64f_pins.json

Do `ls` and look over the list to see if there are other one-offs.

src/zjs_ocf_common.h
```diff
@@ -8,8 +8,6 @@
 
 #include "oc_api.h"
 #include <stdio.h>
-#include <signal.h>
-#include <pthread.h>
 //#include "port/oc_signal_main_loop.h"
```
grgustaf: Just random cleanup? Maybe remove comment too?

scripts/analyze
pfalcon: Even better: 
~~~
option, value = i.split('=', 1)
~~~
will allow to have values with "=" inside (2nd arg to split() is how many splits to make, not how many parts to get).

scripts/analyze
pfalcon: +1 to @grgustaf's comment

scripts/analyze
pfalcon: +1 again. I also don't think preserving original order would be important, but might make sense to sort the result to have stable result from run to run:

~~~
sorted(set(list1) | set(list2))
~~~

src/zjs_aio.json
```diff
@@ -0,0 +1,10 @@
+{
+    "module":"aio",
```
pfalcon: Here you don't have space after `:`...

src/zjs_aio.json
```diff
@@ -0,0 +1,10 @@
...
+    "require":"aio",
+    "targets":["arduino_101"],
+    "zephyr_conf": {
+        "all":["CONFIG_ADC=y"]
+    },
+    "zjs_config": ["-DBUILD_MODULE_AIO"],
```
pfalcon: ... and here you have space after `:`...

Would be nice to have a small consistency with this stuff (apparently, with space after `:`).

src/zjs_arduino101_pins.json
pfalcon: > Try to be consistent with a space after :,

Oops, keep duplicating @grgustaf 's suggestions ;-)

src/zjs_buffer.json
pfalcon: +1

pfalcon:
This looks really great! That's what I call a great maintainers' work - no contributor could come up with such patch, because there're so many decision points, so review would be a chore for both a contributor and maintainers, because many things potentially would need changing. But maintainers know how they want to see it, and in this case, these scheme is really flexible from a quick enough look.

I agree with many @grgustaf's comments on how it can be quickly improved, even for this initial version.

src/zjs_i2c.json
```diff
@@ -0,0 +1,12 @@
...
+    "require": "i2c",
+    "depends": ["buffer", "ipm"],
+    "zephyr_conf": {
+        "all": [
+            "CONFIG_I2C=y"
+        ]
+    },
+    "zjs_config": ["-DBUILD_MODULE_I2C"],
+    "src": ["zjs_i2c.c", "zjs_i2c_ipm_handler.c"]
+}
```
brianjjones: Needs the following added I think...
"src": ["zjs_i2c.c", "zjs_i2c_handler.c", "zjs_i2c_ipm_handler.c"]
"header": ["zjs_i2c.h", "zjs_i2c_handler.h", "zjs_i2c_ipm_handler.h"],
"init" : [zjs_i2c_init],
"cleanup" : ["zjs_i2c_cleanup"]

brianjjones:
+1

jimmy-huang:
AIO doesn't works either

jimmy-huang:
same goes for I2C, GroveLCD, Sensors, all the arc-specif CONFIGs is now in prj.conf where it should've been in the arc/prj.conf

brianjjones:
What Jimmy said, CONFIG_I2C should only be in arc/prj.conf

scripts/analyze
```diff
@@ -0,0 +1,298 @@
...
+def get_dep_list(dlist, board, tree):
+    for dep in tree:
+        if 'targets' not in tree[dep]:
+            merge_lists(dlist, [dep])
+        get_dep_list_helper(dlist, dep, board, tree)
+
+def gather_deps_helper(js, dep, dlist, zlist, zjslist, jrslist, srclist, tree):
+    if dep in tree:
+        entry = tree[dep]
+        if 'targets' in entry:
```
grgustaf: Same thing, you can drop this if because of the get below.

scripts/analyze
```diff
@@ -0,0 +1,298 @@
...
+        for i in gbl_inits:
+            file.write("    {\"%s\", %s" % (i, gbl_inits[i]["inits"][0]))
+            if 'cleanups' in gbl_inits[i]:
+                file.write(", %s" % gbl_inits[i]["cleanups"][0])
+            file.write("},\n")
+        file.write("};\n")
+
+if args['BOARD'] == 'linux':
+    restrict = args['RESTRICT'].split(',')
+    restrict = [i.strip() for i in restrict]
```
grgustaf: Doesn't really matter but I doubt these items can have any whitespace since they came from an arg RESTRICT=foo,bar,baz I think?

scripts/analyze
```diff
@@ -0,0 +1,298 @@
...
+                if re.search(r"require\([\'\"]%s[\"\']\)" % i, js):
+                    js_modules += [i]
+            for i in js_modules:
+                with open('modules/' + i) as mod:
+                    f.write(mod.read())
+                    f.write("\n")
+            f.write(js)
+        js_file = args['JS']
+    else:
+        js_file = args['SCRIPT']
```
grgustaf: Diff between JS= and SCRIPT= is unclear here; part of why intent should be documented in comments at the top and/or here.

scripts/analyze
jprestwo: Make was adding a space between them if I escaped newline like:

```
RESTRICT=foo,\
bar,\
baz
```

Your right it shouldn't do that but this is why I did it.

grgustaf:
+1

Issue #1099, opened 2017-05-04
[timer] Error range is too large on setTimeout function


cuiyanx:
### Description

Error range is too large on setTimeout function. 
 

### Test Code

**Simple timer test1:**
```
console.log("Simple timer test1");
var performance = require("performance");

var timer1, timer2, timer3, timer4, timer5, timer6, timer7, timer8;
var begin1, begin2, begin3, begin4, begin5, begin6, begin7, begin8;
var runtime1, runtime2, runtime3, runtime4, runtime5, runtime6, runtime7, runtime8;

console.log("timer", "begin", "runtime", "delay", "error");

begin1 = performance.now();
timer1 = setTimeout(function () {
    runtime1= performance.now();
    console.log("timer1", begin1, runtime1, "3000", runtime1 - begin1);
}, 3000);

begin2 = performance.now();
timer2 = setTimeout(function () {
    runtime2= performance.now();
    console.log("timer2", begin2, runtime2, "4000", runtime2 - begin2);
}, 4000);

begin3 = performance.now();
timer3 = setTimeout(function () {
    runtime3= performance.now();
    console.log("timer3", begin3, runtime3, "5000", runtime3 - begin3);
}, 5000);

begin4 = performance.now();
timer4 = setTimeout(function () {
    runtime4= performance.now();
    console.log("timer4", begin4, runtime4, "6000", runtime4 - begin4);
}, 6000);

begin5 = performance.now();
timer5 = setTimeout(function () {
    runtime5= performance.now();
    console.log("timer5", begin5, runtime5, "7000", runtime5 - begin5);
}, 7000);

begin6 = performance.now();
timer6 = setTimeout(function () {
    runtime6= performance.now();
    console.log("timer6", begin6, runtime6, "8000", runtime6 - begin6);
}, 8000);

begin7 = performance.now();
timer7 = setTimeout(function () {
    runtime7= performance.now();
    console.log("timer7", begin7, runtime7, "9000", runtime7 - begin7);
}, 9000);

begin8 = performance.now();
timer8 = setTimeout(function () {
    runtime8= performance.now();
    console.log("timer8", begin8, runtime8, "10000", runtime8 - begin8);
}, 10000);
```
**Simple timer test2:**
```
console.log("Simple timer test2");
var performance = require("performance");

var timer1, timer2, timer3, timer4, timer5, timer6, timer7, timer8;
var begin1, begin2, begin3, begin4, begin5, begin6, begin7, begin8;
[...]

PR #1105, opened 2017-05-05
[pwm] Fix timer error range on setTimeout function

cuiyanx:
**ISSUE:** #1099 
**TOTAL:** 12 of 12 passed


jimmy-huang:
+1

PR #1110, opened 2017-05-05
[ashell] Adding echo command

brianjjones:
This adds the echo command, which when turned off will prevent
commands from being printed in the console.

Original patch by Sudarsana Nagineni <sudarsana.nagineni@intel.com>

### DIFF
--- a/src/ashell/comms-shell.c
+++ b/src/ashell/comms-shell.c
@@ -542,7 +542,7 @@ uint32_t ashell_process_finish()
 
 void ashell_print_status()
 {
-    printk("Shell Status\n");
+    DBG("Shell Status\n");
 
     malloc_stats();
 
--- a/src/ashell/comms-shell.c
+++ b/src/ashell/comms-shell.c
@@ -48,11 +48,13 @@ const char *comms_get_prompt()
 
 #define MAX_LINE 90
 #define MAX_ARGUMENT_SIZE 32
+#define CMD_ECHO_OFF "echo off\n"
 
 static ashell_line_parser_t app_line_cb = NULL;
 static char *shell_line = NULL;
 static uint8_t tail = 0;
 static bool ashell_is_done = false;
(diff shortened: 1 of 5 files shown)


nagineni:
IDE side changes here - https://github.com/01org/zephyrjs-ide/pull/127/commits/dc2223e

nagineni:
I just noticed that this PR has the old commit of echo command functionality. I will create a new one with the correct fix.

brianjjones:
@nagineni FYI the reason ashell is failing the build is due to lack of memory.  I'm going to have to do a bit of work to get things to fit again.  Sorry for that, I'll keep you posted.

brianjjones:
@nagineni OK got it shrunk down, just let me know where your new changes are and I'll use them instead.

nagineni:
@brianjjones you can cherry-pick from my branch here https://github.com/nagineni/zephyr.js/commits/echo_mode  or I can create a separate PR with all these changes.

brianjjones:
@nagineni The only difference I see in that branch is the "comms_println("echo_on");" and echo_off prints.  I intentionally removed those as @poussa had asked we cut down on the chatter seen in the console.  Is there a reason for the print other than extra information?  If so could it possibly be a DBG print instead?

nagineni:
@brianjjones Those messages are needed to inform the IDE back that the echo mode is off/on so that the IDE can prevent printing duplicate prompts (i.e 'acm>') received from ashell on the console. Note that we are already handling those messages in the IDE and skipping them so they don't appear on the console.

This is the output we see in the console when the echo mode is off:

JS: console.log("Hello World!!");

Output:

```
acm> 
Hello World!!
acm> 
```

nagineni:
I just created a new PR with all the changes here https://github.com/01org/zephyr.js/pull/1120

brianjjones:
Closing this one in favor of #1120

PR #1111, opened 2017-05-06
[main] Quick fix: Timer added in callback goes unnoticed

grgustaf:
The semaphore change recently broke some things. This showed up for me
in samples/AlarmClock.js but also may be the root cause of part of

Timers were being checked first, then callbacks, so if for example a
GPIO input interrupt came that caused a callback, and that callback
set a timer, it still thought there were no timers. This fix may or
may not be sufficient, but it's probably not the best way to do it
and the problem should be considered more carefully.


jimmy-huang:
+1

PR #1112, opened 2017-05-06
[trlite] Run JS tests with a new bash function for better output

grgustaf:


jimmy-huang:
+1

Issue #1115, opened 2017-05-09
[ocf] Can not find resource on OCF client


cuiyanx:
### Description

OCFClient can not find resource, it will be hanging.
 

### Test Code

Add console.log on [/samples/OcfClient.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfClient.js):
    Add `console.log("----1----");` on line 44;
    Add `console.log("----2----");` on line 50;

Like this:
```
console.log("----1----");
client.findResources({ resourceType:"core.light" }, onfound).then(function(resource) {
    console.log("findResources() was successful, deviceId=" + resource.deviceId);
}).catch(function(error) {
    console.log("findResources() returned an error: " + error.name);
});
console.log("----2----");
```

Result:
![selection_001](https://cloud.githubusercontent.com/assets/22387607/25834967/e7ad2b2e-34ad-11e7-96fa-380679d3f08d.jpg)

 

### Steps to Reproduction

Run test case and check console logs.
 

### Actual Result

Can not find resource and will be hanging.
 

### Expected Result

Find resource and work well.
 

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 3950beb | Arduino 101 | May 9, 2017 | Fail

qiaojingx:
Verified with commit 011f0fe + patch #1125 . This issue is not reproducible.
Will close this issue after patch merged.

cuiyanx:
Fixed and closed

PR #1119, opened 2017-05-09
[tests] Add start api and multiple resources for OCF server

cuiyanx:
**Wait issue fix:** #1115 

### DIFF
--- a/tests/test-ocfserver.js
+++ b/tests/test-ocfserver.js
@@ -5,6 +5,9 @@ console.log("Test OCF server");
 var ocf = require("ocf");
 var assert = require("Assert.js");
 
+var bleAddress = "F1:E2:D3:C4:B5:A6";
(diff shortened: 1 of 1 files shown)


tests/test-ocfserver.js
```diff
@@ -5,6 +5,9 @@ console.log("Test OCF server");
 var ocf = require("ocf");
 var assert = require("Assert.js");
 
+var bleAddress = "F1:E2:D3:C4:B5:A6";
+ocf.setBleAddress(bleAddress);
+
```
jprestwo: Note that this will only work on the A101, on the FRDM board it will likely throw an error. The (IMO better) way to set the BLE address is via the command line with:
`BLE_ADDR="<address>"`

But if the only test target is arduino 101, then this is fine.

grgustaf:
+1

PR #1124, opened 2017-05-09
[fs] Add size to stats object

### LINKED ISSUE #1106: [fs] Add size to stats
Can we add the file size to the stat object returned by the statSync() so that the following sample would work:

```js
var fd = fs.openSync(file, "r");
var stats = fs.statSync(file);
var buf = new Buffer(stats.size);
fs.readSync(fd, buf, 0, stats.size, 0);
```

Node.js fs API has it.

jprestwo:
Fixes #1106


grgustaf:
+1

PR #1128, opened 2017-05-09
[BLE] Fixes disconnect callback not getting called

### LINKED ISSUE #1088: [ble] Disconnect can't trigger callback function
### Description

Disconnect can't trigger callback function
 
### Test Code

[BLE.js](https://github.com/01org/zephyr.js/blob/master/samples/BLE.js); [test-ble.js](https://github.com/01org/zephyr.js/blob/master/tests/test-ble.js)
 
### Steps to Reproduction

1. `make JS=samples/BLE.js ROM=256`
2. `make dfu`
3. Connect A101 with Nexus 5 via nRF app.
4. Disconnect ble connection from nRF app side.

### Actual Result

No disconnection info is outputted while connection broke up in nRF app.
selection_062
 
### Expected Result

BLE disconnection info prints when try to disconnect from nRF app.
 
### Test Builds

---|---|---|---|---
master | 78bc531 | Arduino 101 | May 3, 2017 | Fail

jimmy-huang:
The issue is that the c callback is immediately removed
after the callback is signaled, and the main loop have not
serviced the callback yet.  We don't need to keep a separate
callback for each connection anyway, so instead use
a singal C callback for connect and disconnects.

Fixes #1088

### DIFF
--- a/docs/ble.md
+++ b/docs/ble.md
@@ -33,7 +33,7 @@ specific API functions.
 
 [NoInterfaceObject]
 interface BLE {
-    void disconnect();
+    void disconnect(string address);
     void on(string eventType, EventCallback callback);
     void startAdvertising(string name, string[] uuids, string url);
     void stopAdvertising();
@@ -94,9 +94,11 @@ API Documentation
 -----------------
 ### BLE.disconnect
 
-`void disconnect();`
+`void disconnect(string address);`
 
 Disconnect the remote client.
+
+The `address` is the address of the connected client.
 
 ### BLE.on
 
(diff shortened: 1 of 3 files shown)


grgustaf:
@jimmy-huang, @jprestwo - from the description of the first patch it sure sounds like you're looking for zjs_add_callback_once?

jprestwo:
@jimmy-huang and I discussed and yes a way to add a C callback "one time" is a clean solution. There are a few other places where this would be beneficial. The workaround currently is to remove the callback inside the C callback, which works, but you do have to know the ID.

grgustaf:
Oh I see, we don't already have that function for C, only for JS callbacks? Well we could either merge this first and then address that when we like, or do it now - either way is fine.

grgustaf:
+1

PR #1132, opened 2017-05-10
[trlite] Fix output of linux tests when run with -l

grgustaf:


jimmy-huang:
+1

PR #1134, opened 2017-05-10
[ashell] Fixing duplicate characters on UART

brianjjones:
This fixes the duplicate character issue on UART, removes a bunch
of unused strings, and functions that were redundant.

### DIFF
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -309,12 +309,9 @@ void comms_write_buf(const char *buf, int len)
     while (data_transmitted == false);
     uart_irq_tx_disable(dev_upload);
 }
-void comms_print(const char *buf, bool newline)
+void comms_print(const char *buf)
 {
     comms_write_buf(buf, strnlen(buf, MAX_LINE_LEN));
-
-    if (newline)
-        comms_write_buf("\r\n", 2);
 }
 
 /**
@@ -378,7 +375,7 @@ void comms_runner_init()
     setbuf(stdout, NULL);
 
     ashell_help("");
-    comms_print(comms_get_prompt(), false);
+    comms_print(comms_get_prompt());
     process_state = 0;
 
     atomic_set(&uart_state, UART_INIT);
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -284,7 +284,8 @@ static void comms_interrupt_handler(struct device *dev)
 
 static int comms_out(int c)
 {
-    comms_write_buf((char*)((int*)&c), 1);
+    char ch = (char)c;
+    comms_write_buf(&ch, 1);
     return 1;
 }
 
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -148,21 +148,6 @@ void fifo_recycle_buffer(struct comms_input *data)
     }
     k_fifo_put(&avail_queue, data);
 }
-void comms_clear(void)
-{
-    void *data = NULL;
-    do {
-        if (data != NULL)
-            zjs_free(data);
-        data = k_fifo_get(&avail_queue, K_NO_WAIT);
-    } while (data);
-
-    do {
-        if (data != NULL)
-            zjs_free(data);
-        data = k_fifo_get(&data_queue, K_NO_WAIT);
-    } while (data);
-}
 
 /**************************** UART CAPTURE **********************************/
 
@@ -233,10 +218,6 @@ static void comms_interrupt_handler(struct device *dev)
             bytes_received += bytes_read;
             tail += bytes_read;
 
-            for (uint32_t t = 0; t<bytes_read; t++) {
-                printk("%c",buf[t]);
-            }
-
             /* We don't want to flush data too fast otherwise we would be allocating
             * but we want to flush as soon as we have processed the data on the task
             * so we don't queue too much and delay the system response.
@@ -303,7 +284,7 @@ static void comms_interrupt_handler(struct device *dev)
 
 static int comms_out(int c)
 {
-    comms_writec((char)c);
+    comms_write_buf((char*)((int*)&c), 1);
     return 1;
 }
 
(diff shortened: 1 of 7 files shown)


src/ashell/comms-uart.c
```diff
@@ -322,31 +303,18 @@ void comms_write_buf(const char *buf, int len)
...
         return;
 
-    struct device *dev = dev_upload;
-    uart_irq_tx_enable(dev);
-
+    uart_irq_tx_enable(dev_upload);
     data_transmitted = false;
-    uart_fifo_fill(dev, buf, len);
-
+    uart_fifo_fill(dev_upload, (const uint8_t *)buf, len);
```
pfalcon: FYI, this is not a correct way to use this function. It returns number of bytes actually written, which can be less than requested, including 0: https://github.com/zephyrproject-rtos/zephyr/blob/cc334c72732bbb23d7809f5f010ecb32345afd62/include/uart.h#L248

brianjjones: This is how all the Zephyr examples use it, my understanding is it means I don't want to do anything specific if the uart fails to copy my entire buffer.  I hate to block here and keep trying to get the uart_fifo_fill to succeed, as perhaps it could go on forever.  I'm open to suggestions however, is it expected we would keep trying until we are sure we had sent all the bytes? At what point do we give up? Perhaps @grgustaf has some insight?

Thanks for the comment.

grgustaf: You're already busy waiting on data_transmitted so if that's anywhere close to okay, you should be fine to expand the loop to retry on the remaining bytes, until all your bytes are gone.

So basically:
```
while (len) {
    len -= uart_blah()
}

would maybe do it? Let's hear from our resident UART expert @jprestwo. :)

Comment (grgustaf) [inline src/ashell/comms-uart.c]: I see you're right, some samples do exactly what you did, others ignore the return value just assume it works, without even the data transmitted loop.

The function @pfalcon pointed to shows that it will return 0 if there is no such api in the driver; so if that could ever happen we would hang in my above code. Maybe then 0 should be considered an error? Not sure, the samples and documentation are insufficient.

`test_uart_fifo.c` shows this in `test_fifo_fill`:
```c
		data_transmitted = false;
		while (!uart_fifo_fill(uart_dev, (uint8_t *) &fifo_data[i], 1));
		while (data_transmitted == false);
```

so I guess they're expecting 0 to be a recoverable error.

The only other sample code that ever checks the return value of the function is in `drivers/bluetooth/hci/h4.c` which considers it an error if it's not 1 (they were trying to write just one byte).

Maybe we should consider 0 an error for now to avoid hanging? Otherwise there doesn't seem to be any way to distinguish "the buffer's just full, I can't write any bytes" from "something's not working and will never work".

pfalcon: > The function @pfalcon pointed to shows that it will return 0 if there is no such api in the driver; so if that could ever happen we would hang in my above code.

Well, that's not the problem to care about. If CONFIG_UART_INTERRUPT_DRIVEN is defined, that function can't be non-empty, it's part of the core interrupt-driven API. The problem is like above - you can't expect it to write everything you asked it for.

Code like that (not checking return) is the underlying cause of all those "can't get lines longer 90 chars" and "it doesn't work on any other hardware" problems.

src/ashell/jerry-code.c
```diff
@@ -356,28 +356,6 @@ int javascript_parse_code(const char *file_name, bool show_lines)
...
-                } else {
-                    if (!isprint(byte)) {
-                        comms_printf("(%x)", byte);
-                    } else
-                        comms_writec(byte);
-                }
-            }
-            comms_println("[END]");
-        }
-
```
grgustaf: Are you taking this out to save space? Seems kind of neat but I didn't know about it. You could give a 'v' arg on the command line when displaying code and get line numbers?

src/ashell/comms-uart.c
```diff
@@ -322,31 +303,18 @@ void comms_write_buf(const char *buf, int len)
...
-}
-
-void comms_print(const char *buf)
-{
-    comms_write_buf(buf, strnlen(buf, MAX_LINE_LEN * 4));
+    uart_irq_tx_disable(dev_upload);
 }
-
-void comms_println(const char *buf)
+void comms_print(const char *buf, bool newline)
```
grgustaf: If you do insist on removing writec, then it's cleaner to use

 ```c
write_buf(" ", 1)
```
than
```c
char space = ' ';
write_buf(&space, 1);
```

src/ashell/jerry-code.c
brianjjones: Yes its basically redundant code, all its doing is printing the file with lines.  If someone REALLY wants to print the file with lines I can add that to the ashell_print_file command, that's where it belongs.  But again I don't see the point personally, the IDE has line numbers and is far easier to read.  And size is a real problem now.

src/ashell/comms-uart.c
brianjjones: Agreed, that was leftover from testing.  And yes I don't want to have several comms_write_* functions, it makes it confusing (to me at least) what exactly is being done in each.

brianjjones: Looking at it, it seems to make more sense to remove the newline arg and just add them to the msg strings.  I still get to remove the extra function, and its clearer what is being done.

src/ashell/shell-state.c
```diff
@@ -282,37 +272,38 @@ int32_t ashell_print_file(char *buf)
...
         return RET_OK;
     }
 
     fs_seek(file, 0, SEEK_SET);
-    if (lines)
-        comms_printf("%5d  ", line++);
 
+    int lineStart = 0;
     do {
-        count = fs_read(file, data, 4);
```
brianjjones: Rather than reading and printing one character at a time, it does up to 80.  Note comms_printf is no longer used for critical output as it will occasionally add extra characters.

src/ashell/shell-state.c
```diff
@@ -644,12 +618,14 @@ static const struct ashell_cmd commands[] =
 
 int32_t ashell_help(char *buf)
 {
-    comms_println("'A Shell' bash\r\n");
-    comms_println("Commands list:");
+    comms_print("'A Shell' bash\r\n\r\n");
+    comms_print("Commands list:\r\n");
     for (uint32_t t = 0; t < ASHELL_COMMANDS_COUNT; t++) {
-        comms_printf("%8s %s\r\n", commands[t].cmd_name, commands[t].syntax);
+        comms_print(commands[t].cmd_name);
```
brianjjones: Again using comms_print and comms_write_buf instaed of comms_printf to avoid extra chars.

brianjjones:
To make this clear, comms_printf and comms_writec appear to be the culprits in adding extra characters.  I removed comms_writec and now use comms_write_buf exclusively.  I also removed some functions I felt were redundant.  Sorry I didn't break those into another patch, was on a cleanup mission and didn't think about it until I had put it all together.  Let me know if anything else still isn't clear.

src/ashell/comms-uart.c
```diff
@@ -303,7 +284,7 @@ static void comms_interrupt_handler(struct device *dev)
 
 static int comms_out(int c)
 {
-    comms_writec((char)c);
+    comms_write_buf((char*)((int*)&c), 1);
```
grgustaf: Could you change that to this:

```
char ch = (char)c;
comms_write_buf(&c, 1);
```

grgustaf:
+1, I verified this helps the problem a lot. But that's on output (like `cat [filename]`) not on input echoing (like `load [filename]` + paste). So more to be done!

PR #1137, opened 2017-05-12
[ashell] Update docs to mention generic udev rules for the WebUSB device

nagineni:
Update ashell docs to mention generic udev rules which are compatible
on all Linux distros. We don't need to disable the modem manager when
there is a udev rule that prevents the modem manager interfering with
the device, so remove instructions that were added for disabling the
modem manager.


nagineni:
@zolkis Could you please check this on Fedora?

zolkis:
Works for me on Fedora 25.

grgustaf:
+1

PR #1139, opened 2017-05-12
[fs] Fix truncate check for wrong return value

jimmy-huang:
FS.truncate() will always fail since it doesn't check the wrong return
value from Zephyr


jprestwo:
+2

jimmy-huang:
@jprestwo you can add it along with your other patch :)

grgustaf:
+1

PR #1141, opened 2017-05-12
[general] Turn on jerryscript error messages

jprestwo:


jimmy-huang:
you can probably try with 256 instead, I think maybe you are using 0.8.2 on your system, so it didn't fit, with 0.9, the size is smaller.

poussa:
+1 Must have

PR #1142, opened 2017-05-12
[fs] Fix issue where it the max string it can write is 256

jimmy-huang:

### DIFF
--- a/src/zjs_fs.c
+++ b/src/zjs_fs.c
@@ -826,7 +826,7 @@ static ZJS_DECL_FUNC_ARGS(zjs_fs_write_file, uint8_t async)
     char *data = NULL;
     uint32_t length;
     if (jerry_value_is_string(argv[1])) {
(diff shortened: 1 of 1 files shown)


src/zjs_fs.c
```diff
@@ -826,7 +826,7 @@ static ZJS_DECL_FUNC_ARGS(zjs_fs_write_file, uint8_t async)
     char *data = NULL;
     uint32_t length;
     if (jerry_value_is_string(argv[1])) {
-        size = 256;
+        size = jerry_get_string_size(argv[1]);
         data = zjs_alloc_from_jstring(argv[1], &size);
```
grgustaf: How about just set size to 0, then alloc will get as much as necessary. Not really any different, one less func call.

jimmy-huang:
@grgustaf updated

grgustaf:
+1

PR #1154, opened 2017-05-15
[ashell] Disable raw mode control messages in the console

nagineni:
This disables the raw mode control/transfer messages in
the console when echo is off.


grgustaf:
+1

brianjjones:
+1

PR #1157, opened 2017-05-16
[ashell] Disable PME from Ashell/IDE to gain more ROM

jimmy-huang:


poussa:
+1

grgustaf:
This is fine but after disabling UART we may not need more space savings right now, in which case we could leave it in. Let's keep it on the back burner for the moment.

poussa:
yeah, let's keep PME on ide around for now.

jimmy-huang:
I am closing this until further ROM is needed.

PR #1159, opened 2017-05-16
[ashell] Further fixes for character duplication in ashell

brianjjones:

### DIFF
--- a/src/ashell/comms-shell.c
+++ b/src/ashell/comms-shell.c
@@ -403,6 +403,7 @@ void ashell_process_line(const char *buf, uint32_t len)
 uint32_t ashell_process_data(const char *buf, uint32_t len)
 {
     uint32_t processed = 0;
+    // printed Is used to make sure we don't re-print characters
     uint8_t printed = cur;
     bool flush_line = false;
     if (shell_line == NULL) {
@@ -478,11 +479,13 @@ uint32_t ashell_process_data(const char *buf, uint32_t len)
                 break;
             }
         }
-
+        // We have a newline, flush the current line
         if (flush_line) {
             DBG("Line %u %u \n", cur, end);
             shell_line[cur + end] = '\0';
-            comms_print_strncpy(shell_line, printed, cur);
+            if (comms_get_echo_mode()) {
+                comms_write_buf(shell_line + printed, cur - printed);
+            }
             comms_write_buf("\r\n", 2);
 
             uint32_t length = strnlen(shell_line, MAX_LINE);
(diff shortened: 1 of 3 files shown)


src/ashell/comms-shell.c
```diff
@@ -80,10 +80,6 @@ static void insert_char(char *pos, char c, uint8_t end)
 {
     char tmp;
 
-    /* Echo back to console */
```
brianjjones: Rather than printing here one character at a time, I now use comms_print_strncpy to print all text in the uart that hasn't been echoed yet.  This prevents duplicated characters and improves performance when pasting large amounts of text.

src/ashell/comms-uart.c
```diff
@@ -299,23 +299,40 @@ static int comms_out(int c)
 
 void comms_write_buf(const char *buf, int len)
 {
-    if (len == 0)
-        return;
+    if (comms_get_echo_mode()) {
```
grgustaf: I'm a little concerned about this because write_buf is called ~20 times; I hope none of those other ones are supposed to be getting printed even though we 've turned off "echo"?

src/ashell/comms-uart.c
```diff
@@ -299,23 +299,40 @@ static int comms_out(int c)
...
+
 void comms_print(const char *buf)
 {
-    comms_write_buf(buf, strnlen(buf, MAX_LINE_LEN));
+    if (comms_get_echo_mode()) {
+        comms_write_buf(buf, strnlen(buf, MAX_LINE_LEN));
+    }
+}
+
+void comms_print_strncpy(const char *buf, uint16_t start, uint16_t end)
```
brianjjones: Sweet thanks.

src/ashell/comms-shell.c
```diff
@@ -483,10 +479,13 @@ uint32_t ashell_process_data(const char *buf, uint32_t len)
...
         }
-
+        // We have a newline, flush the current line
         if (flush_line) {
             DBG("Line %u %u \n", cur, end);
             shell_line[cur + end] = '\0';
+            if (comms_get_echo_mode()) {
+                comms_write_buf(shell_line + printed, cur - printed);
+            }
             comms_write_buf("\r\n", 2);
```
grgustaf: This should be within the echo check too, right?

src/ashell/comms-shell.c
```diff
@@ -80,10 +80,6 @@ static void insert_char(char *pos, char c, uint8_t end)
 {
     char tmp;
 
-    /* Echo back to console */
-    if (echo_mode)
-        comms_write_buf(&c, 1);
```
grgustaf: I still don't understand where there's another line in this insert_char function:
```
        comms_write_buf(&tmp, 1);
```

Can you make sure that's okay and maybe add some comment describing this strange function's purpose. :)

src/ashell/comms-shell.c
brianjjones: ah yes

src/ashell/comms-shell.c
brianjjones: It seems fine, I never see it called.  It has to do with handling ANSI values.

grgustaf:
+1, still slightly iffy on insert_char implementation but @brianjjones has done a fair amount of testing and we haven't seen anything big.

PR #1161, opened 2017-05-16
[ashell] Update Windows instructions

nagineni:
Update Windows instructions in ashell docs and mention how to
install the WinUSB driver with the Zadig utility.


poussa:
+1

PR #1168, opened 2017-05-17
[ashell] Buffer stdout before sending on wire

grgustaf:
Waits until a newline or 80 chars on stdout arrive before sending.

This seems to solve, or at least radically decrease likelihood,
of duplicated characters on redirected stdout.

### DIFF
--- a/src/ashell/comms-uart.c
+++ b/src/ashell/comms-uart.c
@@ -282,8 +282,14 @@ static void comms_interrupt_handler(struct device *dev)
 
 static int comms_out(int c)
 {
+    static char buf[80];
+    static int size = 0;
     char ch = (char)c;
-    comms_write_buf(&ch, 1);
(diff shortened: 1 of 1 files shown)


jimmy-huang:
+1

pfalcon:
There's only one step left (after this patch, not now) to switch to circular buffer and output it in ISR ;-), which will work on any hardware (needs testing). Or maybe we all even will be able to use upstream implementation: https://github.com/zephyrproject-rtos/zephyr/commit/3cb13b96873b26bf92e43b019011ed0a1def5232#diff-7a4cc31c1c7007e220a2592064272fbb

brianjjones:
Doesn't this mean that if someone fails to put a \n at the end of their printf, it won't show up until the buffer hits 80?  Or are we assuming that all printf will contain a \n ?  If so it looks fine to me.

PR #1169, opened 2017-05-17
[ashell] Fix add_requires to allow single as well as double quotes

grgustaf:
Also, improve parsing of require() by requiring parentheses, requiring
the string to be found within them, but allowing whitespace around them.

Note some of the remaining foibles with comments.

### DIFF
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -221,6 +221,7 @@ static bool skip_char_with_whitespace(char **ptr, char match)
     if (**ptr != match) {
         return false;
     }
+    ++(*ptr);
     skip_whitespace(ptr);
     return true;
 }
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -195,53 +195,118 @@ static void eval_list(requires_list_t **list)
     }
 }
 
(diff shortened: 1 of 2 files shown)


grgustaf:
You can test this just by running `require('gpio');` in the IDE.

jimmy-huang:
+1

src/ashell/jerry-code.c
```diff
@@ -195,53 +195,122 @@ static void eval_list(requires_list_t **list)
...
+    // modifies: *ptr
+    //  effects: skips any whitespace in *ptr before and after a match char or
+    //             until null terminator is reached
+    //           if first non-whitespace char found is match, returns true;
+    //             otherwise, false
+    skip_whitespace(ptr);
+    if (**ptr != match) {
+        return false;
+    }
+    skip_whitespace(ptr);
```
brianjjones: You need a ++(*ptr); before this line.  Otherwise skip_whitespace won't do anything as ptr is currently pointing at '('

grgustaf: Thanks!

brianjjones:
@grgustaf Please hold off on this one, I think I have a more elegant solution that doesn't require an extra function.  Will show you first thing tomorrow.

Thanks.

PR #1171, opened 2017-05-17
[util] Search up the object hierarchy for full function path

jprestwo:
- When printing out the error message function, this will
   search up the object tree to find the full object path
   of the function.


grgustaf:
+1, I found a bug where the buffer can have some garbage but I've already fixed it in a follow-on patch I'll submit soon.

PR #1184, opened 2017-05-19
[buffer] Fix buffer.toString() to bail if there is 0xff in ascii

jprestwo:
- This prevents buffer from composing a string that could be used
   to access hidden ZJS properties


jimmy-huang:
+1

PR #1186, opened 2017-05-19
[general] Fix reversed order args from recent ZJS_DECL_FUNC patch

grgustaf:


jimmy-huang:
+1

PR #1196, opened 2017-05-24
[tests] Add test case for board APIs

cuiyanx:
**TOTAL:**  2 of 2 passed


jimmy-huang:
@cuiyanx why can't this be automated test instead of manual?

cuiyanx:
@jimmy-huang Because we can only get the board information in this way, and there is no object to match(we're not sure which board to be used). Finally, we still have to do manual comparisons.

jimmy-huang:
@cuiyanx, ok sounds all right.

PR #1198, opened 2017-05-24
[net] Update tcp client to add 'enter' for finished transmission

cuiyanx:

### DIFF
--- a/samples/TCPClient6.js
+++ b/samples/TCPClient6.js
@@ -35,8 +35,9 @@ net.on('up', function() {
 		}
 
 		client.on("data", function(data) {
+			count++;
 			console.log("got data: " + data.toString('ascii'));
-			client.write(new Buffer("write data " + count++ + "\r\n"));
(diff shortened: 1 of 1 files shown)


jimmy-huang:
+1

samples/TCPClient6.js
```diff
@@ -31,12 +31,12 @@ net.on('up', function() {
...
 			console.log("connect successful");
 			is_connected = true;
-			client.write(new Buffer("initial data"));
+			client.write(new Buffer("initial data\r\n"));
 		}
 
 		client.on("data", function(data) {
 			console.log("got data: " + data.toString('ascii'));
-			client.write(new Buffer("write data " + count++));
+			client.write(new Buffer("write data " + count++ + "\r\n"));
```
jimmy-huang: maybe do the counter inrement on a different line, then append to the string, it's probably cleaner than ++ +

cuiyanx: Yeah, I think so, but I will try to keep intact.

PR #1202, opened 2017-05-24
[aio] Fixed AIO read slow on ARC

jimmy-huang:
The adc_read() on Zephyr has a latency of 100ms which causes
the AIO read to be very slow.  This patch now spawns a thread
that fetch ADC for any pins that are enabled so that
the main thread on ARC will just look up the fetched value.

### DIFF
--- a/arc/src/main.c
+++ b/arc/src/main.c
@@ -50,9 +50,9 @@ static struct zjs_ipm_message *end_of_queue_ptr = msg_queue + QUEUE_SIZE;
 #define STACK_PRIORITY 7
 static char __stack stack[STACK_SIZE];
 static struct device *adc_dev = NULL;
-static uint8_t pin_enabled[ARC_AIO_LEN] = {};
-static uint32_t pin_values[ARC_AIO_LEN] = {};
-static uint32_t pin_last_values[ARC_AIO_LEN] = {};
+static atomic_t pin_enabled[ARC_AIO_LEN] = {};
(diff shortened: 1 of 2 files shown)


grgustaf:
Is this a bug in Zephyr?

jimmy-huang:
@grgustaf I think so, because it's adc_read() that's taking so long, that's a native zephyr call

jimmy-huang:
https://www.zephyrproject.org/doc/1.7.0/kernel/common/common_atomic.html

It does say that it support a 32bit value, no?

jprestwo:
@jimmy-huang, ah yes your right. LGTM.

scripts/analyze.sh
```diff
@@ -525,7 +525,10 @@ if check_for_feature "Accelerometer\|Gyroscope\|AmbientLightSensor\|TemperatureS
             fi
         fi
         if $sensor_light; then
+            # Ambient Light sensor only works with AIO at the moment
+            MODULES+=" -DBUILD_MODULE_AIO"
             echo "CONFIG_ADC=y" >> $ARCPRJFILE
+            echo "export ZJS_AIO=y" >> $CONFFILE
```
grgustaf: You'll presumably have to fix this after the JSON patch goes in.

jimmy-huang:
@grgustaf rebased

grgustaf:
+1

Issue #1206, opened 2017-06-01
[net] There is an error on the type of 'family' attribute in AddressOptions dictionary


cuiyanx:
### Description

In [/docs/net.md](https://github.com/01org/zephyr.js/blob/master/docs/net.md):
```
dictionary AddressOptions {
    Number port;          // Port the client should connect to (required)
    String host;          // Host the client should connect to
    String localAddress;  // Local address to bind to
    Number localPort;     // local port to bind to
    String family;        // Version of IP stack, deafults to 4
}
```
The type of 'family' attribute in doc is 'string';

 In [/samples/TCPClient6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPClient6.js):
`client.connect({port: 4242, host:'fe80::5ef3:70ff:fe78:1080', localAddress:'2001:db8::1', localPort: 8484, family: 6}, connected);
`

But the type of 'family' attribute in sample is 'number';

### Actual Result

Description and implementation does not match.
 

### Expected Result

Please update doc to same as sample.
 

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 6180feca | Arduino 101 | June 1, 2017 | Fail

cuiyanx:
Fixed and closed

PR #1207, opened 2017-06-01
[net] Update the type of 'family' attribute to same as sample

cuiyanx:
Issue: #1206 


jimmy-huang:
+1

PR #1208, opened 2017-06-01
[sensor] k64f support

jimmy-huang:
Refactor sensor and decouple IPM logic into a separate sensor_board
module.  Added initial support for Accelerometer for the FRDM-K64F
board.


src/zjs_sensor_frdm.json
```diff
@@ -0,0 +1,15 @@
+{
```
grgustaf: Rename this file from frdm to k64f, that's what we've been using elsewhere as a short abbreviation.

src/zjs_sensor_gyro.json
```diff
@@ -2,7 +2,7 @@
     "module": "sensor_gyro",
     "constructor": "Gyroscope",
     "depends": ["sensor"],
-    "targets": ["arduino_101"],
+    "targets": ["arduino_101", "frdm_k64f"],
```
grgustaf: Not sure this makes sense if as per IRC it's not actually going to work on k64f.

src/zjs_sensor_board_frdm.c
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2016-2017, Intel Corporation.
```
grgustaf: As below use `k64f` not `frdm` in filename, too generic.

jimmy-huang:
@grgustaf updated patch

grgustaf:
+1

PR #1209, opened 2017-06-02
[SPI] Initial PR for SPI

brianjjones:
To run the test, connect pins 11 and 12 to each other on the A101.
This causes the MOSI to send its data to the MISO.


brianjjones:
This is the first pass of the SPI API implementation.  I'm sure there are improvements that can be made, comments appreciated.  See https://github.com/01org/iot-js-api/blob/master/board/aio.md for more info.

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+
+    if (spi_slave_select(handle->spi_device, jerry_get_number_value(argv[0]))) {
+        return zjs_error("SPI slave select failed\n");
+    }
+
+    // If only a slave is given, this must be a single read or its invalid
+    if (argc == 1) {
+        if (handle->topology != ZJS_TOPOLOGY_SINGLE_READ) {
+            return ZJS_STD_ERROR(RangeError, "Missing args for transceive");
+        }
```
zolkis: "Missing transmit buffer"? (perhaps the message needs to be aligned with other APIs)

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+    enum direction dirArg = ZJS_SPI_DIR_UNDEFINED;
+
+    // We might have been passed a 'direction' arg, get it and validate
+    if (argc > 1 && jerry_value_is_string(argv[argc - 1])) {
+        // Find which argument contains the direction arg
+        if (argc == 2)
+            arg = argv[1];
+        else
+            arg = argv[2];
+
```
zolkis: Wait, the direction is always the 3rd arg, if buffer is not passed, `null` needs to be passed instead.

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+        uint32_t dirLen = jerry_get_string_size(arg) + 1;
+        zjs_copy_jstring(arg, dirString, &dirLen);
+
```
zolkis: Maybe here you could check if dir is defined, if not, then takes READ_WRITE.

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+        if (strncmp(dirString, "read-write", 11) == 0)
+            dirArg = ZJS_SPI_DIR_READ_WRITE;
+        else if (strncmp(dirString, "read", 5) == 0)
+            dirArg = ZJS_SPI_DIR_READ;
+        else if (strncmp(dirString, "write", 6) == 0)
+            dirArg = ZJS_SPI_DIR_WRITE;
+
```
zolkis: and if defined, but not one of these values, throw TypeError

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+        // If there are 3 args and the last one is not a direction or
+        // If there are 2 args and the direction calls for a write return error
```
zolkis: This condition is not valid; if there are 2 args, the first is a device, the second is a buffer or `null`, and direction is read-write.

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+        if ((argc == 3 && dirArg == ZJS_SPI_DIR_UNDEFINED) ||
+            (argc == 2 && (dirArg == ZJS_SPI_DIR_WRITE || dirArg == ZJS_SPI_DIR_READ_WRITE)))
+            return ZJS_STD_ERROR(RangeError, "Invalid args");
+
```
zolkis: this  (104 to 107) IMHO needs rewriting

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+    }
+
```
zolkis: before continuing, I'd check here is direction matches with topology; this would simplify the code in continuation

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+        // If we need to write a buffer
+        if (handle->topology != ZJS_TOPOLOGY_SINGLE_READ && dirArg != ZJS_SPI_DIR_READ) {
+            // If topology is single write and has a dirArg other than write
```
zolkis: at this point it should be enough to check direction

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+                zjs_copy_jstring(argv[1], txBuf->buffer, &len);
+            }
+            // If this is a read / write
+            if (handle->topology != ZJS_TOPOLOGY_SINGLE_WRITE &&
+                dirArg != ZJS_SPI_DIR_WRITE) {
+                rxBuf_obj = zjs_buffer_create(len, &rxBuf);
+                // Send the data and read from the device
+                if (spi_transceive(handle->spi_device, txBuf->buffer , txBuf->bufsize, rxBuf->buffer, rxBuf->bufsize) != 0) {
+                    return zjs_error("SPI transceive failed\n");
+                }
```
zolkis: throw SystemError?

src/zjs_spi.c
```diff
@@ -0,0 +1,310 @@
...
+                rxBuf_obj = jerry_create_null();
+            }
+        }   // This is a read only operation
+        else {
+            rxBuf_obj = zjs_buffer_create(MAX_READ_BUFF, &rxBuf);
+            // Read the data from the device
+            if (spi_read(handle->spi_device, rxBuf->buffer, rxBuf->bufsize) !=0) {
+                return zjs_error("SPI transceive failed\n");
+            }
```
zolkis: throw SystemError?

zolkis:
BTW the "spec" is also open, can be found at
https://github.com/zolkis/iot-js-api/blob/master/board/spi.md

Should be checked, since the implementation seems to be based on somewhat older version.

jimmy-huang:
@brianjjones is there supposed to be a sample JS with this patch?

src/zjs_spi.h
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2017, Intel Corporation.
+
+#ifndef __zjs_spi_h__
+#define __zjs_spi_h__
+
+#include "jerryscript.h"
+
+enum SPITopology {
+    ZJS_TOPOLOGY_FULL,
```
brianjjones: Sure, probably clarifies things.

src/zjs_spi.h
```diff
@@ -0,0 +1,40 @@
...
+    ZJS_TOPOLOGY_SINGLE_READ,
+    ZJS_TOPOLOGY_SINGLE_WRITE,
+    ZJS_TOPOLOGY_MULTIPLEXED,
+    ZJS_TOPOLOGY_DAISY_CHAIN
+};
+
+enum direction {
+    ZJS_SPI_DIR_UNDEFINED,
```
brianjjones: Originally I had intended this to be for when a direction wasn't passed in as an arg.  However, its not really necessary so I'll remove it to keep things clear.

src/zjs_spi.c
brianjjones: So you couldn't call..
transceive(0, read); ?  

Probably makes more sense to pass NULL to keep it consistent, certainly easier on my end ;)

src/zjs_spi.c
brianjjones: I set it as the default now since I removed the UNDEFINED enum

brianjjones:
@jimmy-huang  Yeah, good catch thanks.

src/zjs_spi.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2017, Intel Corporation.
+
+#ifndef __zjs_spi_h__
+#define __zjs_spi_h__
+
+#include "jerryscript.h"
+
+enum spi_topology {
+    ZJS_TOPOLOGY_FULL_DUPLEX,
+    ZJS_TOPOLOGY_SINGLE_READ,
```
zolkis: I have renamed this to TOPOLOGY_READ because it may not be a single read. In some doc it figures like "single read" though.

src/zjs_spi.h
```diff
@@ -0,0 +1,38 @@
...
+    ZJS_TOPOLOGY_SINGLE_WRITE,
```
zolkis: I have renamed this to TOPOLOGY_Write because it may not be a single write. In some doc it figures like "single write" though.

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+        zjs_obj_get_uint32(init, "phase", &phase);
+        if (phase != 0 && phase != 2)
+            return ZJS_STD_ERROR(TypeError, "Invalid phase");
+
+        // Connection type
+        zjs_obj_get_string(init, "topology", topologyStr, 13);
+        if (strncmp(topologyStr, "full-duplex", 12) == 0)
+            topology = ZJS_TOPOLOGY_FULL_DUPLEX;
+        else if (strncmp(topologyStr, "single-read", 12) == 0)
+            topology = ZJS_TOPOLOGY_SINGLE_READ;
```
zolkis: "read"

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+        else if (strncmp(topologyStr, "single-write", 12) == 0)
```
zolkis: "write"

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    ZJS_GET_HANDLE(this, spi_handle_t, handle, spi_type_info);
+
+    if (handle->closed == true) {
+        ZJS_PRINT("SPI bus is closed\n");
+        return jerry_create_null();
+    }
+
+    size_t len;
+    jerry_value_t buffer;
+    zjs_buffer_t *txBuf;
```
jimmy-huang: use lowercase underscore like tx_buf

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    zjs_buffer_t *rxBuf;
```
jimmy-huang: rx_buf

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    jerry_value_t rxBuf_obj = jerry_create_null();
```
jimmy-huang: rx_buf_obj

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    jerry_value_t txBuf_obj;
```
jimmy-huang: tx_buf_obj

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    // If only a slave is given, this must be a single read or its invalid
+    if (argc == 1) {
+        if (handle->topology != ZJS_TOPOLOGY_READ) {
+            return ZJS_STD_ERROR(RangeError, "Missing transmit buffer");
+        }
+    }
+
+    char dirString[13];
```
jimmy-huang: dir_string

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    enum spi_direction dirArg;
```
jimmy-huang: dir_arg

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+            char *tmpBuf = zjs_alloc_from_jstring(argv[1], NULL);
+            strncpy(txBuf->buffer, tmpBuf, txBuf->bufsize);
+            zjs_free(tmpBuf);
+        }
+        // If this is a read / write
+        if (dirArg == ZJS_SPI_DIR_READ_WRITE) {
+            rxBuf_obj = zjs_buffer_create(txBuf->bufsize, &rxBuf);
+            // Send the data and read from the device
+            if (spi_transceive(handle->spi_device, txBuf->buffer , txBuf->bufsize, rxBuf->buffer, rxBuf->bufsize) != 0) {
+                return ZJS_STD_ERROR(SystemError, "SPI transceive failed");
```
jimmy-huang: leak here with buffer obj

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+        // This is a write only operation, return a NULL buffer
+        else {
+            if (spi_write(handle->spi_device, txBuf->buffer , txBuf->bufsize) !=0) {
+                return ZJS_STD_ERROR(SystemError, "SPI transceive failed");
```
jimmy-huang: again, leak here

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+            }
+            rxBuf_obj = jerry_create_null();
+        }
+    }   // This is a read only operation
+    else {
+        rxBuf_obj = zjs_buffer_create(MAX_READ_BUFF, &rxBuf);
+        // Read the data from the device
+        if (spi_read(handle->spi_device, rxBuf->buffer, rxBuf->bufsize) !=0) {
+            return ZJS_STD_ERROR(SystemError, "SPI transceive failed");
```
jimmy-huang: leak here

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+        init = argv[0];
+
+    // Default values
+    uint32_t bus = 0;
+    double speed = 10;
+    bool msbFirst = true;
+    uint32_t bits = 8;
+    uint32_t polarity = 0;
+    uint32_t phase = 0;
+    char topologyStr[13] = "";
```
jimmy-huang: topology_str, we only use camelcase in Javascript, use lower case for all C

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    char busStr[9];
```
jimmy-huang: bus_str

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+    enum spi_topology topology = ZJS_TOPOLOGY_FULL_DUPLEX;
+    uint32_t frameGap = 0;
```
jimmy-huang: frame_gap

src/zjs_spi.h
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2017, Intel Corporation.
+
+#ifndef __zjs_spi_h__
+#define __zjs_spi_h__
+
+#include "jerryscript.h"
+
+enum spi_topology {
```
jimmy-huang: It seems there's no need for these to be public, so maybe move these to the source file instead.

src/zjs_spi.h
```diff
@@ -0,0 +1,38 @@
...
+    ZJS_TOPOLOGY_FULL_DUPLEX,
+    ZJS_TOPOLOGY_READ,
+    ZJS_TOPOLOGY_WRITE,
+    ZJS_TOPOLOGY_MULTIPLEXED,
+    ZJS_TOPOLOGY_DAISY_CHAIN
+};
+
+enum spi_direction {
+    ZJS_SPI_DIR_READ,
```
jimmy-huang: It seems topology has the same mapping as direction?  READ == READ, WRITE == WRITE, and all others = READ/WRITE?    Maybe you can just consolidate into one enum?

src/zjs_spi.h
```diff
@@ -0,0 +1,38 @@
...
+    ZJS_SPI_DIR_WRITE,
+    ZJS_SPI_DIR_READ_WRITE
+};
+
+typedef struct spi_handle {
+    struct device *spi_device;
+    enum spi_topology topology;
```
jimmy-huang: I see that you don't keep the direction in the handle, that's why i think you could've drop the direction.

src/zjs_spi.c
```diff
@@ -0,0 +1,314 @@
...
+
+static jerry_value_t zjs_spi_prototype;
+
+static void zjs_spi_callback_free(void *native)
+{
+    // requires: handle is the native pointer we registered with
+    //             jerry_set_object_native_handle
+    //  effects: frees the spi item
+    spi_handle_t *handle = (spi_handle_t *)native;
+    handle->closed = true;
```
jimmy-huang: not sure if you really need to set closed to true, since you are freeing the handle immediately.

brianjjones: I do because after you free it, the object still exists in JS until garbage collection.  The user can try and use the SPI object again before this happens, need a way to make sure they don't.

src/zjs_spi.h
brianjjones: Yeah I think I can just use full_duplex for the read_write direction

src/zjs_spi.c
```diff
@@ -0,0 +1,319 @@
...
+            tx_buf_obj = zjs_buffer_create(jerry_get_string_size(buffer), &tx_buf);
+            // zjs_copy_jstring adds a null terminator, which we don't want
+            // so make a new string instead and remove it.
+            char *tmpBuf = zjs_alloc_from_jstring(argv[1], NULL);
+            strncpy(tx_buf->buffer, tmpBuf, tx_buf->bufsize);
+            zjs_free(tmpBuf);
+        }
+        // If this is a read / write
+        if (dir_arg == ZJS_TOPOLOGY_FULL_DUPLEX) {
+            rx_buf_obj = zjs_buffer_create(tx_buf->bufsize, &rx_buf);
```
grgustaf: `tx_buf` can be NULL here if it failed above, so you need to check it somewhere and fail I guess.

Do you know somehow how much data they're going to be sending you back? Using the size of the TX buffer doesn't make sense to me. I think what you want to do instead is call the SPI function passing it a C buffer you allocate a maximum expected size like 256 bytes or whatever. Once you've got that data back, THEN you create your Buffer object of the right size and fill it in.

(I could be wrong, I don't know anything about SPI. Are you guaranteed TX messages = RX messages perhaps?)

src/zjs_spi.c
```diff
@@ -0,0 +1,319 @@
...
+            return ZJS_STD_ERROR(RangeError, "Missing transmit buffer");
+        }
+    }
+
+    char dir_str[13];
+    // Set the direction default based on the topology.
+    enum spi_topology dir_arg = handle->topology;
+
+    // If we have a 'direction' arg, get it and validate
+    if (argc == 3) {
```
brianjjones: We don't consider too many args to be invalid?

src/zjs_spi.c
```diff
@@ -0,0 +1,319 @@
...
+        if (dir_arg == ZJS_TOPOLOGY_READ && !jerry_value_is_null(argv[1])) {
+            return ZJS_STD_ERROR(NotSupportedError, "Buffer should be null when direction is read");
+        }
+    }
+
+    // If we need to write a buffer
+    if (dir_arg != ZJS_TOPOLOGY_READ) {
+        buffer = argv[1];
+        // Figure out if the buffer is an array or string, handle accordingly
+        if (jerry_value_is_array(argv[1])) {
```
brianjjones: I think it is far easier for the user to be able to pass an array and not be burdened with creating the Buffer object themselves.  And yes, I was following the example of how it would be used in JS.

src/zjs_spi.c
brianjjones: If it's full duplex, the RX is guaranteed to be the same size as the TX.  As per the Zephyr SPI API docs.

 * @brief Read and write the specified amount of data from the SPI driver.
 *
 * This routine is meant for full-duplex transmission.
 * Only equal length is supported(tx_buf_len must be equal to rx_buf_len).

src/zjs_spi.c
grgustaf: 1) It's a squishy language, and 2) It's good enough for C. :)

I forget, I've done some searching for common practice on the topic. Just now I opened Chrome inspector and typed Date() with a whole bunch of arguments and it was fine with it, just ignored them after the first three.

But if we decide to do it, it makes sense to do it uniformly everywhere using macros rather than repeating the logic in every function. So I took the occasional appearance of such logic out when I added ZJS_VALIDATE_ARGS.

src/zjs_spi.c
grgustaf: It's a little easier to say [1, 2, 3] than Buffer.from([1, 2, 3]) which I don't think we support yet anyway... but it also means you're reimplementing a lot of logic that will have weird corner cases.

I guess we could implement common handling for it in C and reuse that code.

src/zjs_spi.c
```diff
@@ -0,0 +1,322 @@
...
+                        tx_buf->buffer[i] = 0;
+                    }
+                }
+            }
+            else {
+                 return ZJS_STD_ERROR(Error, "Failed to get TX buffer");
+            }
+        }
+        else {
+            tx_buf_obj = zjs_buffer_create(jerry_get_string_size(buffer), &tx_buf);
```
grgustaf: tx_buf can also be NULL here.

grgustaf:
+1, just do that NULL check in your follow-on doc patch I guess.

src/zjs_spi.c
brianjjones: Sorry wasn't clear where the null check was missing.  Ill check argv[1] for being null if a write is going to occur.

PR #1210, opened 2017-06-05
[sensor] Fix bug where CONFIG_ADC=y is added for all sensors

jimmy-huang:
CONFIG_ADC is only needed for the AmbientLight sensor, but the
analyze script incorrectly adds it for all sensors.


jimmy-huang:
this is fixed after the new analyze, closing this PR

PR #1211, opened 2017-06-05
[general] Use Zephyr defined types

jimmy-huang:
Update Zephyr and rebase ZJS to move away from the C99 {u}int{8,16,32,64}_t
types to Zephyr defined u{8,16,32,64}_t and s{8,16,32,64}_t.


grgustaf:
+1

PR #1232, opened 2017-06-12
Update status of supported modules in README

haoxli:
- K64F: File System, OCF, Dgram, Net
- F411RE: Buffer, Console, Event, File System, Performance, Timers


grgustaf:
+1

PR #1240, opened 2017-06-13
[webusb] Define device interface GUID

nagineni:
This defines device interface GUID in the extended properties
descriptor so that the Windows generic driver WinUSB will
be loaded automatically when the device is connected to the
host. This makes the installation process simple on Windows.

Also, we do not need to install the driver manually anymore, so
removing Zadig utility instructions from the docs.


poussa:
This is great!

brianjjones:
LGTM, thanks! +1

Issue #1151, opened 2017-05-15
[net] Only first string can be received by TCPEchoServ6.js on QEMU


qiaojingx:
### Description

[reg][net] Only first string can be received by TCPEchoServ6.js on QEMU

### Test Code

[TCPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ6.js); 

### Steps to Reproduction

1. `make BOARD=qemu_x86 qemu JS=samples/TCPEchoServ6.js`
2. `nc -6 2001:db8::1 4242` to connect with TCPEchoServ6.js
3. Send string

### Actual Result

Only first string can be received. 
![selection_005](https://cloud.githubusercontent.com/assets/19813546/26051664/65919494-3995-11e7-87a5-8ed9baa93b58.png)

### Expected Result

String can be received continuously.

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | bd07595 | QEMU | May 15, 2017 | Pass
master | e5f62c0 - e24821c| QEMU | May 15, 2017 | Fail

### Additional Information

This issue is probably caused by commit e5f62c0.

grgustaf:
Fixed by PR #1231.

qiaojingx:
Verified with commit 819a41a + #1231. This issue is not reproducible.
Will close it after patch merged.

grgustaf:
Qemu was further broken by #1104 but that is fixed by #1247.

qiaojingx:
Verified with commit 78c9749. This issue has been fixed.

PR #1247, opened 2017-06-15
[ipss] Make ipss inclusion its own module

grgustaf:

### DIFF
--- a/src/zjs_ipss.json
+++ b/src/zjs_ipss.json
@@ -1,0 +1,6 @@
+{
+    "module": "ipss",
+    "targets": ["arduino_101"],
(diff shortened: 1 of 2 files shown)


grgustaf:
This fixes a bug I found after including the JSON Analyze patch. When you try the qemu networking steps  from #1151, you would get this error:

```
/home/pepsi/dev/zjs/project/src/zjs_net.c:1064: undefined reference to `zjs_net_config'
collect2: error: ld returned 1 exit status
```

This patch fixes it for me, but I'm not sure if it's the right solution so I'd like @jprestwo to review.

jprestwo:
LGTM

jimmy-huang:
+1

PR #1253, opened 2017-06-15
[ashell] Restore default modules included in ashell/ide mode

grgustaf:
Some were dropped accidentally in the recent analyze/json patch.

Also, remove no longer used zjs.conf.dev file.


grgustaf:
@jimmy-huang when I try to run sensorbledemo.js from IDE with this patch, it longer gives an error, but it doesn't work either. It never prints out that it gets to advertising. I'll do some more testing but wanted you to be able to take a look.

jimmy-huang:
+1

PR #1258, opened 2017-06-16
[modules] Fix bug where module string printed can be empty

### LINKED ISSUE #1191: Missing module not displayed in the error message
If the JavaScript app requires a module that is missing, the console displays an error message stating so but does not display what module is missing.

Steps to reproduce this:
1. Run a ZJS image that does *not* include support for `iotivity-constrained`
2. Load [this JavaScript sample](https://github.com/01org/SmartHome-Demo/blob/master/ocf-servers/zjs-servers/led.js) via the Web IDE
3. Run it and the following message will appear on the console:
```
acm> In function require():
NotSupportedError: module not found: ''
```
I would expect the last line to actually display:
```
NotSupportedError: module not found: 'ocf'
```

grgustaf:
Fixes #1191.

### DIFF
--- a/src/zjs_modules.c
+++ b/src/zjs_modules.c
@@ -92,12 +92,15 @@ static ZJS_DECL_FUNC(native_require_handler)
         return SYSTEM_ERROR("exports object not found");
     }
 
-    for (int i = 0; i < 4; i++) {
-        // Strip the ".js"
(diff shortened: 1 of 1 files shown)


brianjjones:
+1 LGTM

PR #1274, opened 2017-06-19
[SPI] Adding docs for SPI.  Also adding a NULL check

brianjjones:


grgustaf:
+1

PR #1285, opened 2017-06-20
[trlite] Add testing of all the modules that should work on k64f

grgustaf:


jimmy-huang:
+1

PR #1287, opened 2017-06-21
[modules] Fixing including JS modules that include arc modules

brianjjones:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,7 +235,7 @@ analyze: $(JS)
 
 	./scripts/analyze	V=$(V) \
 		SCRIPT=$(JS) \
-		JS_OUT=JS_TMP \
+		JS_OUT=$(JS_TMP) \
 		BOARD=$(BOARD) \
 		JSON_DIR=src/ \
 		FORCE=$(ASHELL) \
@@ -323,7 +323,7 @@ cleanlocal:
 	@rm -f prj.conf.tmp
 	@rm -f prj.mdef
 	@rm -f zjs.conf.tmp
-	@rm -f JS_TMP
+	@rm -f $(JS_TMP)
 
 # Explicit clean
 .PHONY: clean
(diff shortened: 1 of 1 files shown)


jprestwo:
+1 LGTM

jimmy-huang:
I see a couple other places in the Makefile that uses $(JS) instead of js.tmp, shouldn't those be changed as well? maybe fix that in this patch too, and instead of hard-coding js.tmp, create another variable called JS_TMP or something

@./scripts/convert.sh $(JS)

jprestwo:
@jimmy-huang, in some cases we do want to pass in just the $(JS) file. There are two "script" type parameters, JS_OUT is for the output JS file (after we append JS requires) and SCRIPT is for the original script.

BUT scripts/convert.sh should be using js.tmp otherwise you wont get any of the appended JS requires. @brianjjones said this patch fixed the issue but maybe it was not using convert.sh in his test?

jimmy-huang:
yeah, i think convert.sh need to use js.tmp for Linux builds as well

brianjjones:
This update should take care of all that.

Makefile
```diff
@@ -353,17 +354,17 @@ ifeq ($(SNAPSHOT), on)
...
 	else \
-		cat js.tmp > outdir/jsgen.tmp; \
+		cat $(JS_TMP) > outdir/jsgen.tmp; \
 	fi
 	@outdir/snapshot/snapshot outdir/jsgen.tmp > outdir/include/zjs_snapshot_gen.h
 else
 	@echo Creating C string from JS application...
 ifeq ($(BOARD), linux)
-	@./scripts/convert.sh $(JS) outdir/include/zjs_script_gen.h
+	@./scripts/convert.sh $(JS_TMP) outdir/include/zjs_script_gen.h
```
jimmy-huang: remove the linux then, since they are the same now

jimmy-huang:
+1

Issue #1294, opened 2017-06-21
[spi] SPI bus can only read and write array and string


cuiyanx:
### Description

Now, SPI bus can only read and write array and string. I test buffer data as `number` and `boolean`, SPI bus will be hang.
 

### Test Code

Change `"Hello World\0"` to `true` or `1024` in [/samples/SPI.js](https://github.com/01org/zephyr.js/blob/master/samples/SPI.js) on line 10.

 

### Actual Result

Logs: `SPI test starting..`
 

### Expected Result

 Can read and write number or boolean or some buffer data.

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 15d3847 | Arduino 101 | June 21, 2017 | Fail

brianjjones:
Currently numbers and bools are not supported.  But I don't see why they couldn't be so I'll look into adding it.

grgustaf:
I really don't think that makes sense. Accepting Buffer, however, makes a lot of sense. :)

grgustaf:
Marking P2, but this is for handling Buffer.

brianjjones:
https://github.com/01org/zephyr.js/pull/1315 will add the ability to pass it a Buffer

qiaojingx:
Verified with commit 3403f93. Buffer object can be passed.

cuiyanx:
fixed and closed

PR #1295, opened 2017-06-21
[tests] Add test for SPI APIs

cuiyanx:
**ISSUE:**   #1294
**TOTAL:** 22 of 22 passed


tests/test-spi.js
```diff
@@ -0,0 +1,117 @@
...
+assert.throws(function() {
+    spiBus = spi.open({speed:20, bus:128, polarity:0, phase:0, bits:16});
+}, "SPIBusObject: set SPI bus as '128'");
+
+spiBus = spi.open({speed:20, bus:1, polarity:0, phase:0, bits:16});
+assert(typeof spiBus === "object" && spiBus !== null,
+       "SPIBusObject: be defined");
+
+var spiBusOptions = [
+    [spiBus.bus, "object", "bus", 1],
```
grgustaf: As discussed elsewhere, these properties are not part of the documented SPIBus API and shouldn't be expected.

cuiyanx: I refer to this document [/board/spi.md](https://github.com/zolkis/iot-js-api/blob/master/board/spi.md). If we do not support, I will remove it.

tests/test-spi.js
```diff
@@ -0,0 +1,83 @@
...
+    [1, 2],
+    [1, 2, 3],
+    [1, 2, 3, 4, 5, 6, 7, 8],
+    [80, 90]
+];
+
+var checkNumber = [
+    ["0102", "010203", "0102030405060708", "000a"],
+    ["0102", "010203", "0102030405060708", "505a"],
+    ["0102", "010200", "0102030405060708", "505a"]
```
brianjjones: While this test should be correct, currently Zephyr will sometimes return gibberish when the the data size isn't divisible by the bits number.  So you'll see random bits added to the last byte of the return buffer.  This appears to be a Zephyr issue, if its an issue at all. It might be the expected result.

brianjjones:
LGTM, do note my comment about using an array that isn't divisible by the number of bits being sent.  I'll ping the Zephyr team and see if this is expected behavior.

brianjjones:
Another option is adding trailing zeros to the transmit buffer.

tests/test-spi.js
grgustaf: Oh I see, @zolkis what do you think here? It does seem like in our constrained environment we should leave out reporting back to the user the settings they used to configure the bus. They should know what they did.

Or we could make it optional for the implementation?

tests/test-spi.js
brianjjones: I'm fine with making it an error if you try and pass an incorrect number of bytes.  I'd also be OK with just zeroing out the rest.

grgustaf:
Rebase on master with #1364 and the build should pass.

jimmy-huang:
+1

PR #1297, opened 2017-06-21
[SPI] Changing the SPI example to print with hex instead of ascii

brianjjones:
If the MOSI / MISO pins are not connected, the print will fail
with the message SPI error: buffer has invalid ascii. To avoid
this, I'm changing it to print using HEX.


grgustaf:
+1

PR #1306, opened 2017-06-22
[zephyr] Update to Zephyr 1.8.0

grgustaf:


jimmy-huang:
+1

jimmy-huang:
We should merge this now, just note that ADC is broken with this patch, but I filed a bug upstream https://jira.zephyrproject.org/browse/ZEP-2363, will be tracking that closely.

PR #1309, opened 2017-06-23
[sensor] Added Magnetometer sensor support

jimmy-huang:
It will support the fxos8700 magnetometer sensor on the FRDM_K64F


grgustaf:
+1

PR #1310, opened 2017-06-23
[uart] Fix for 1279 UART not working

brianjjones:
The lock in signal_callback is not freeing before UART is
triggered.  This is causing UART to be locked out when it tries
to write.  Moving the unlock up fixes the issue.  Also adding a
needed item to the zjs_uart.json file.

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -465,7 +465,11 @@ void signal_callback_priv(zjs_callback_id id,
             0,  // we use value for CB_FLUSH_ONE/ALL
             (u32_t *)args,
             (u8_t)((size + 3) / 4));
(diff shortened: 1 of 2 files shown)


src/zjs_callbacks.c
```diff
@@ -465,6 +465,7 @@ void signal_callback_priv(zjs_callback_id id,
             0,  // we use value for CB_FLUSH_ONE/ALL
             (u32_t *)args,
             (u8_t)((size + 3) / 4));
+    UNLOCK();
```
jimmy-huang: don't you need to move the zjs_loop_unblock() too for Linux builds?

jimmy-huang:
+1

grgustaf:
Merging by proxy...

PR #1314, opened 2017-06-27
[general] Clean up for formatting

jimmy-huang:

### DIFF
--- a/arc/src/arc_aio.c
+++ b/arc/src/arc_aio.c
@@ -1,13 +1,13 @@
 // Copyright (c) 2017, Intel Corporation.
 
 // Zephyr includes
-#include <zephyr.h>
-#include <kernel.h>
 #include <adc.h>
+#include <kernel.h>
+#include <zephyr.h>
 
 // ZJS includes
-#include "arc_common.h"
 #include "arc_aio.h"
+#include "arc_common.h"
 
 // AIO thread
 #define STACK_SIZE 1024
@@ -76,8 +76,7 @@ u32_t arc_pin_read(u8_t pin)
     }
 
     // read from buffer, not sure if byte order is important
-    u32_t raw_value = (u32_t) seq_buffer[0]
-                    | (u32_t) seq_buffer[1] << 8;
+    u32_t raw_value = (u32_t)seq_buffer[0] | (u32_t)seq_buffer[1] << 8;
 
     return raw_value;
 }
@@ -85,8 +84,7 @@ u32_t arc_pin_read(u8_t pin)
 void arc_process_aio_updates()
 {
     for (int i = 0; i <= 5; i++) {
-        if (pin_send_updates[i] &&
-            pin_values[i] != pin_last_values[i]) {
+        if (pin_send_updates[i] && pin_values[i] != pin_last_values[i]) {
             // send updates only if value has changed
             // so it doesn't flood the IPM channel
             struct zjs_ipm_message msg;
--- a/arc/src/arc_aio.h
+++ b/arc/src/arc_aio.h
@@ -3,6 +3,9 @@
 #ifndef __arc_aio_h__
 #define __arc_aio_h__
 
+// ZJS includes
+#include "zjs_ipm.h"
+
 void arc_aio_init();
 
 void arc_aio_cleanup();
--- a/arc/src/arc_curie_pme.c
+++ b/arc/src/arc_curie_pme.c
@@ -20,11 +20,12 @@ void curie_pme_begin(void)
     reg_write16(NSR, saved_nsr);
 }
 
-void curie_pme_configure(u16_t global_context,
-                         PATTERN_MATCHING_DISTANCE_MODE distance_mode,
-                         PATTERN_MATCHING_CLASSIFICATION_MODE classification_mode,
-                         u16_t min_if,
-                         u16_t max_if)
+void curie_pme_configure(
+    u16_t global_context,
+    PATTERN_MATCHING_DISTANCE_MODE distance_mode,
+    PATTERN_MATCHING_CLASSIFICATION_MODE classification_mode,
+    u16_t min_if,
+    u16_t max_if)
 {
     reg_write16(GCR, (global_context | (distance_mode << 7)));
     reg_write16(NSR, reg_read16(NSR) | (classification_mode << 5));
@@ -33,12 +34,13 @@ void curie_pme_configure(u16_t global_context,
 }
 
 // clear all commits in the network and make it ready to learn
-void curie_pme_forget(void) { reg_write16(FORGET_NCOUNT, 0); }
+void curie_pme_forget(void)
+{
+    reg_write16(FORGET_NCOUNT, 0);
+}
 
 // mark --learn and classify--
-u16_t curie_pme_learn(u8_t *pattern_vector,
-                      s32_t vector_length,
-                      u16_t category)
+u16_t curie_pme_learn(u8_t *pattern_vector, s32_t vector_length, u16_t category)
 {
     if (vector_length > MAX_VECTOR_SIZE)
         vector_length = MAX_VECTOR_SIZE;
@@ -51,7 +53,8 @@ u16_t curie_pme_learn(u8_t *pattern_vector,
 
     // Mask off the 15th bit-- valid categories range from 1-32766,
     // and bit 15 is used to indicate if a firing neuron has degenerated
-    reg_write16(CAT, (reg_read16(CAT) & ~CAT_CATEGORY) | (category & CAT_CATEGORY));
+    reg_write16(CAT,
+                (reg_read16(CAT) & ~CAT_CATEGORY) | (category & CAT_CATEGORY));
 
     return reg_read16(FORGET_NCOUNT);
 }
@@ -68,7 +71,7 @@ u16_t curie_pme_classify(u8_t *pattern_vector, s32_t vector_length)
     }
 
     reg_write16(LCOMP, current_vector[vector_length - 1]);
-    reg_read16(IDX_DIST); // Sort by distance by David Florey
+    reg_read16(IDX_DIST);
 
     return (reg_read16(CAT) & CAT_CATEGORY);
 }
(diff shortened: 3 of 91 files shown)


jimmy-huang:
@grgustaf I am using clang-format so it will make certain assumptions when you do right alignment on pointers, for example, when you do a cast, it will do (char *)val instead of (char*)val, which I think makes more sense given that you are aligning the * to the right anyway, but it changes our existing styling, I tried to keep most of our style conventions, like using 2 spaces before end of the line comments, etc, if there's a clang rule for it, if you can come up with the correct configuration where we can embed the settings into the project, it will make code maintenance a lot easier, like if you prefer other tools like uncrustify like @jprestwo suggested.

arc/src/arc_aio.c
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) 2017, Intel Corporation.
 
 // Zephyr includes
-#include <zephyr.h>
 #include <adc.h>
+#include <zephyr.h>
```
grgustaf: Alphabetizing includes, I like this.

arc/src/arc_aio.c
```diff
@@ -75,17 +75,15 @@ u32_t arc_pin_read(u8_t pin)
     }
 
     // read from buffer, not sure if byte order is important
-    u32_t raw_value = (u32_t) seq_buffer[0]
-                    | (u32_t) seq_buffer[1] << 8;
+    u32_t raw_value = (u32_t)seq_buffer[0] | (u32_t)seq_buffer[1] << 8;
```
grgustaf: Using one line when it fits in 80 chars, I like this.

arc/src/arc_aio.h
```diff
@@ -3,6 +3,9 @@
 #ifndef __arc_aio_h__
 #define __arc_aio_h__
 
+// ZJS includes
+#include "zjs_ipm.h"
+
```
grgustaf: This isn't cleanup - why this?

arc/src/arc_curie_pme.c
```diff
@@ -20,11 +20,12 @@ void curie_pme_begin(void)
...
-                         PATTERN_MATCHING_DISTANCE_MODE distance_mode,
-                         PATTERN_MATCHING_CLASSIFICATION_MODE classification_mode,
-                         u16_t min_if,
-                         u16_t max_if)
+void curie_pme_configure(
+    u16_t global_context,
+    PATTERN_MATCHING_DISTANCE_MODE distance_mode,
+    PATTERN_MATCHING_CLASSIFICATION_MODE classification_mode,
+    u16_t min_if,
+    u16_t max_if)
```
grgustaf: I don't like the first arg on the next line, and I don't using a whole line per arg when others would fit. I'd even consider making shorter versions of the nasty type names, at least PATTERN_MATCHING_ -> PM_.

Was this necessary for some reason?

arc/src/arc_curie_pme.c
```diff
@@ -33,12 +34,13 @@ void curie_pme_configure(u16_t global_context,
 }
 
 // clear all commits in the network and make it ready to learn
-void curie_pme_forget(void) { reg_write16(FORGET_NCOUNT, 0); }
+void curie_pme_forget(void)
```
grgustaf: All for expanding this out.

I think the (void) is a good idea in .h files but I don't know that it matters in the definition. Although maybe it does if there *is* no declaration, so maybe it's a good practice.

arc/src/arc_curie_pme.c
```diff
@@ -33,12 +34,13 @@ void curie_pme_configure(u16_t global_context,
...
+{
+    reg_write16(FORGET_NCOUNT, 0);
+}
 
 // mark --learn and classify--
-u16_t curie_pme_learn(u8_t *pattern_vector,
-                      s32_t vector_length,
-                      u16_t category)
+u16_t curie_pme_learn(u8_t *pattern_vector, s32_t vector_length, u16_t category)
```
grgustaf: Yep, I like this. (I will try to comment only the first time I see each type of change. :D)

arc/src/arc_curie_pme.c
```diff
@@ -51,7 +53,8 @@ u16_t curie_pme_learn(u8_t *pattern_vector,
 
     // Mask off the 15th bit-- valid categories range from 1-32766,
     // and bit 15 is used to indicate if a firing neuron has degenerated
-    reg_write16(CAT, (reg_read16(CAT) & ~CAT_CATEGORY) | (category & CAT_CATEGORY));
+    reg_write16(CAT,
+                (reg_read16(CAT) & ~CAT_CATEGORY) | (category & CAT_CATEGORY));
```
grgustaf: Personally I would break this at the | instead, or just read reg_read16() into a variable to make it fit.

arc/src/arc_curie_pme.c
```diff
@@ -187,12 +190,13 @@ void curie_pme_end_restore_mode(void)
 }
 
 // mark -- getter and setters--
-PATTERN_MATCHING_DISTANCE_MODE curie_pme_get_distance_mode(void) // L1 or LSup
+PATTERN_MATCHING_DISTANCE_MODE curie_pme_get_distance_mode(void)  // L1 or LSup
```
grgustaf: I like the two spaces, to my eyes it makes it look more like a comment and less like an operation.

otoh i'm not sure what this comment means here and it could go *above* the function instead

arc/src/arc_curie_pme.c
```diff
@@ -187,12 +190,13 @@ void curie_pme_end_restore_mode(void)
...
 {
     return (GCR_DIST & reg_read16(GCR)) ? LSUP_DISTANCE : L1_DISTANCE;
 }
 
-void curie_pme_set_distance_mode(PATTERN_MATCHING_DISTANCE_MODE mode) // L1 or LSup
+void curie_pme_set_distance_mode(
+    PATTERN_MATCHING_DISTANCE_MODE mode)  // L1 or LSup
```
grgustaf: OK now I see what the comment means I guess. I think it's dumb, just drop it. You look  up DISTANCE_MODE if you want to know what it means. Or add param documentation above. It's not clear enough this way.

arc/src/arc_curie_pme.c
```diff
@@ -232,7 +236,8 @@ u16_t curie_pme_get_committed_count(void)
     return (getFORGET_NCOUNT() & 0xff);
 }
 
-PATTERN_MATCHING_CLASSIFICATION_MODE curie_pme_get_classifier_mode(void) // RBF or KNN
+PATTERN_MATCHING_CLASSIFICATION_MODE
+curie_pme_get_classifier_mode(void)  // RBF or KNN
```
grgustaf: Again I'd rather move the comment up with more text and/or create a shortcut name for this return type.

arc/src/arc_curie_pme.h
```diff
@@ -86,35 +89,35 @@ u16_t curie_pme_write_vector(u8_t *pattern_vector, s32_t vector_length);
...
+    AIF = 0x14,           // Active Influence Field
+    MINIF = 0x18,         // Minimum Influence Field
+    MAXIF = 0x1C,         // Maximum Influence Field
+    TESTCOMP = 0x20,      // Write Test Component
+    TESTCAT = 0x24,       // Write Test Category
+    NID = 0x28,           // Network ID
+    GCR = 0x2C,           // Global Context Register
+    RSTCHAIN = 0x30,      // Reset Chain
+    NSR = 0x34,           // Network Status Register
+    FORGET_NCOUNT = 0x3C  // Forget Command / Neuron Count
```
grgustaf: yeah that's nicer

arc/src/main.c
```diff
@@ -75,9 +78,12 @@ static void queue_message(struct zjs_ipm_message *incoming_msg)
     k_sem_give(&arc_sem);
 }
 
-static void ipm_msg_receive_callback(void *context, u32_t id, volatile void *data)
+static void ipm_msg_receive_callback(void *context,
+                                     u32_t id,
```
grgustaf: I'd leave this one on the same line.

src/ashell/comms-shell.h
```diff
@@ -23,6 +23,10 @@ char *ashell_get_token_arg(char *str);
 bool ashell_check_parameter(const char *buf, const char parameter);
 
 u32_t ashell_get_argc(const char *str, u32_t nsize);
-const char *ashell_get_next_arg_s(const char *str, u32_t nsize, char *str_arg, u32_t max_arg_size, u32_t *length);
+const char *ashell_get_next_arg_s(const char *str,
+                                  u32_t nsize,
+                                  char *str_arg,
+                                  u32_t max_arg_size,
+                                  u32_t *length);
```
grgustaf: I'd only wrap the one that runs over 80 chars.

arc/src/arc_aio.h
jimmy-huang: This is needed to compile when including arc_aio.h, the problem only surfaced after the re-ordering of #includes since probably this header was included before arc_aio.h previously

arc/src/arc_curie_pme.c
jimmy-huang: I am specifically telling clang to try to use 1 arg per line when the whole line doesn't fit, only in the function declarations, but not when calling the functions.  Since I've seen how clang treats it differently on different places, I rather enforce the rules to have a consistent style so I can just run the script without going back to manually revert the style, but if you prefer the fit everything you can on the same line, I am okay with it.

arc/src/arc_curie_pme.c
jimmy-huang: Like I said previously, if it fits in the same line, it will put it as 1 line, but if it doesn't fit, then it defaults to break each argument into separate line, but I am okay with fit everything as you can on a per line basis.

arc/src/arc_curie_pme.c
jimmy-huang: again, this is automatically done by clang, I couldn't find a rule to do this automatically for me, but I guess I can do it manually.

arc/src/arc_curie_pme.c
jimmy-huang: I'll take this out

arc/src/main.c
jimmy-huang: To avoid over 80chars for some of these functions, clang actually changes it to,

static void
ipm_msg_receive_callback(void *context, u32_t id, volatile void *data)

If that's something you prefer?

jimmy-huang:
@grgustaf patch updated

arc/src/arc_aio.h
grgustaf: Really arc_aio.h should include that then... that's what my `checkheaders` script tries to check for but I must not run it on arc headers.

grgustaf:
+1

PR #1324, opened 2017-06-29
[trlite] Add fs to k64f tests so we catch build regressions

grgustaf:


jimmy-huang:
+1

Issue #1257, opened 2017-06-16
[ashell] make ashell connect to WebIDE instead of terminal


qiaojingx:
### Description

This is an regression issue. `make ashell` should use the shell via the command line. But it connect to WebIDE, just like `make ide` did.

### Test Code

[ashell.md](https://github.com/01org/zephyr.js/blob/master/docs/ashell.md); 
 
### Steps to Reproduction

1. make ashell ROM=256
2. make dfu
3. Reset 

### Actual Result

A notification from Chrome will appear with an URL of the IDE. A101 can be connect to WebIDE.
 
### Expected Result

It should use serial console instead of browser IDE. And user can use the following command to connect to the ashell from a terminal:
`$ screen /dev/ttyACM0 115200`
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 44a5e6b | Arduino 101 | Jun 16, 2017 | Fail
master | a09bd64 | Arduino 101 | Jun 16, 2017 | Pass

brianjjones:
zjs_ashell.json needs to be broken into multiple files to do this, I'll set that up.

qiaojingx:
Verified with commit 418c871. This issue is fixed.

PR #1325, opened 2017-06-30
[ashell] Fix to allow command line ashell to be built.

brianjjones:
Fix for #1257

### DIFF
--- a/src/zjs_ashell_ide.json
+++ b/src/zjs_ashell_ide.json
@@ -1,42 +1,7 @@
 {
     "module": "ashell_ide",
-    "depends": ["aio", "arduino101_pins", "ble",
-                "common", "console", "fs", "gpio", "grove_lcd", "i2c",
-                "performance", "pwm", "sensor_accel", "sensor_gyro",
-                "sensor_light", "sensor_temp", "spi"],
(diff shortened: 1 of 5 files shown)


grgustaf:
Rather than duplicate most of the JSON in these two files, let's leave a zjs_ashell.json with all the core settings except for the one that changes. Then just make the _cli and _ide files only contain "depends" on "ashell" and specify the one setting that is different.

src/zjs_ashell_ide.json
```diff
@@ -0,0 +1,42 @@
+{
+    "module": "ashell_ide",
+    "depends": ["aio", "arduino101_pins", "ble",
+                "common", "console", "fs", "gpio", "grove_lcd", "i2c",
+                "performance", "pwm", "sensor_accel", "sensor_gyro",
+                "sensor_light", "sensor_temp", "spi"],
```
grgustaf: Based on my previous comment, in the _cli and _ide versions I think this can just contain "ashell" and nothing else.

brianjjones:
Moved all the common code to zjs_ashell.json

grgustaf:
+1 looks great, thanks...

PR #1337, opened 2017-07-07
DHCP and Network configuration

jprestwo:
New "net-config" module allows the JS to do DHCP, set a static IP, and set the BLE address. You can also register for net up and net down events.

Also fixed the issues with the "net" module after the zephyr update.


grgustaf:
+1, want to squash those fixups?

jprestwo:
@grgustaf, you can quash "forgot to depend on events" and "fixed conflict with OCF". The other three are kinda self contained.

grgustaf:
+1

Issue #1055, opened 2017-04-24
[sensor] Missing controller name on sensor and can not set name except default name


cuiyanx:
### Description

The sensor controller name is undefined always. And if setting controller as not default name, creating sensor will be failed. This is inconsistent with [/docs/sensors.md](https://github.com/01org/zephyr.js/blob/master/docs/sensors.md).

**Accelerometer Interface:**
```
dictionary AccelerometerOptions : SensorOptions  {
    string controller;       // controller name, default to "bmi160"
    boolean includeGravity;  // not supported currently
};
```
 **GyroscopeSensor Interface:**
```
dictionary GyroscopeOptions : SensorOptions  {
    string controller;  // controller name, default to "bmi160"
};
```
**AmbientLightSensor Interface:**
```
dictionary AmbientLightSensorOptions : SensorOptions  {
    string controller;  // controller name, default to "ADC_0"
    unsigned long pin;  // analog pin where the light is connected
};
```
**TemperatureSensor Interface:**
```
dictionary TemperatureSensorOptions : SensorOptions  {
    string controller;  // controller name, default to "bmi160"
};

```

### Test Code

Test default name:
```
var pins = require("arduino101_pins");

var Frequency = 20;

var Accelerometer = new Accelerometer({
    includeGravity: false,
    frequency: Frequency
});

console.log("Accelerometer controller: ", Accelerometer.controller);
console.log("Accelerometer frequency: ", Accelerometer.frequency);

var Gyroscope = new Gyroscope({
    frequency: Frequency
});

console.log("Gyroscope controller: ", Gyroscope.controller);
console.log("Gyroscope frequency: ", Gyroscope.frequency);

var Temperature = new TemperatureSensor({
    frequency: Frequency
});

console.log("Temperature controller: ", Temperature.controller);
console.log("Temperature frequency: ", Temperature.frequency);

var AmbientLight = new AmbientLightSensor({
    pin: pins.A2
});

console.log("AmbientLight controller: ", AmbientLight.controller);
console.log("AmbientLight pin: ", AmbientLight.pin);
```
![selection_001](https://cloud.githubusercontent.com/assets/22387607/25322955/37d1420a-28ee-11e7-845c-a458271de16b.jpg)

Test controller name with not default name:
```
var Accelerometer = new Accelerometer({
    controller: "AccelerometerSensor",
    includeGravity: false,
    frequency: 20
});

console.log("Accelerometer controller: ", Accelerometer.controller);
[...]

PR #1342, opened 2017-07-11
[tests] Add test case for magnetometer sensor on FRDM-K64F

cuiyanx:
**TOTAL:** 11 of 11 passed

### DIFF
--- a/modules/GenericSensor.js
+++ b/modules/GenericSensor.js
@@ -117,7 +117,8 @@ function GenericSensor() {
                     console.log("sensor.timestamp: " + sensor.timestamp);
                     console.log(sensorType + ": " + sensor.illuminance);
                 } else if (sensorType === "Accelerometer" ||
-                           sensorType === "Gyroscope") {
+                           sensorType === "Gyroscope" ||
+                           sensorType === "Magnetometer") {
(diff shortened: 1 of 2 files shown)


grgustaf:
@jimmy-huang please review

modules/GenericSensor.js
```diff
@@ -126,8 +127,13 @@ function GenericSensor() {
                            sensor.z !== null,
                            "sensor: reading value for '" + sensorType + "'");
 
-                    assert(sensor.controller === "bmi160",
-                           "sensor: controller name as default");
+                    if (sensorType === "Magnetometer") {
```
jimmy-huang: I think this was discussed in #1055, I don't think we should be checking the sensor.controller as it's not a property that will be part of the API.  The controller is only passed as an optional argument to the constructor, and once the object is created, it's no longer relevant.  @cuiyanx so can you maybe explain what is the need here to check the controller value after it has been set?  Maybe we need to update the API doc a little better to not confuse the developers.

cuiyanx: Li Hao and I have discussed this issue, so we support your point of view. I will delete this test point.

jimmy-huang:
+1

PR #1355, opened 2017-07-12
[build] Work-around fix for Zephyr SDK bug on AIO

jimmy-huang:
The 0.9.1 SDK still has upstream bug #2363 not fixed, so
this work-around patches our build to build with -O2 on ARC
images so it fixes the bug with AIO not booting.


grgustaf:
+1

PR #1356, opened 2017-07-12
[tools] TRACE=on will now print all memory left after cleanup

brianjjones:
This provides a method to print all the memory left after the
module cleanup methods have been called.  It will provide an easy
way to spot memory leaks.  You can use this method to print at any
time, but currently it will only be called after module cleanup.

### DIFF
--- a/src/zjs_util.c
+++ b/src/zjs_util.c
@@ -13,7 +13,7 @@
 #ifdef ZJS_TRACE_MALLOC
 #define MAX_LIST_SIZE 300
 mem_stats_t mem_array[MAX_LIST_SIZE];
-#endif //ZJS_TRACE_MALLOC
+#endif
 
 void *zjs_malloc_with_retry(size_t size)
 {
@@ -25,6 +25,7 @@ void *zjs_malloc_with_retry(size_t size)
     }
     return ptr;
 }
+
 #ifdef ZJS_TRACE_MALLOC
 void zjs_print_mem_stats()
 {
@@ -62,13 +63,13 @@ void zjs_pop_mem_stat(void *rm_ptr)
             if (mem_array[i].ptr == rm_ptr) {
                 mem_array[i].ptr = NULL;
                 mem_array[i].file = "";
-                //free(mem_array[i].func);
                 mem_array[i].line = 0;
             }
         }
     }
 }
-#endif // ZJS_TRACE_MALLOC
+#endif  // ZJS_TRACE_MALLOC
+
 void zjs_set_property(const jerry_value_t obj, const char *name,
                       const jerry_value_t prop)
 {
--- a/src/zjs_util.c
+++ b/src/zjs_util.c
@@ -10,9 +10,10 @@
 #ifndef ZJS_LINUX_BUILD
 #include "zjs_zephyr_port.h"
 #endif
-
+#ifdef ZJS_TRACE_MALLOC
 #define max_list_size 300
 mem_stats_t mem_list[max_list_size];
+#endif //ZJS_TRACE_MALLOC
 
 void *zjs_malloc_with_retry(size_t size)
 {
--- a/src/zjs_util.c
+++ b/src/zjs_util.c
@@ -11,6 +11,9 @@
 #include "zjs_zephyr_port.h"
 #endif
 
+#define max_list_size 300
+mem_stats_t mem_list[max_list_size];
+
 void *zjs_malloc_with_retry(size_t size)
 {
     void *ptr = malloc(size);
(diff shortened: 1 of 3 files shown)


grgustaf:
@brianjjones please rebase, should build now

src/zjs_util.c
```diff
@@ -10,6 +10,10 @@
 #ifndef ZJS_LINUX_BUILD
 #include "zjs_zephyr_port.h"
 #endif
+#ifdef ZJS_TRACE_MALLOC
+#define max_list_size 300
```
grgustaf: Use all caps for a macro like this so we know where to look for the definition.

src/zjs_util.c
```diff
@@ -21,7 +25,50 @@ void *zjs_malloc_with_retry(size_t size)
     }
     return ptr;
 }
+#ifdef ZJS_TRACE_MALLOC
+void zjs_print_mem_stats()
+{
+    for (int i = 0; i < max_list_size; i++) {
+        if (mem_list[i].ptr != NULL) {
+            ZJS_PRINT("index %i %s - %s:%d: %p\n", i, mem_list[i].file, mem_list[i].func, mem_list[i].line, mem_list[i].ptr);
```
grgustaf: Add a newline to wrap this at 80 chars.

src/zjs_util.c
```diff
@@ -21,7 +25,50 @@ void *zjs_malloc_with_retry(size_t size)
...
+
+void zjs_push_mem_stat(void *ptr, char *file, char *func, int line)
+{
+    int i = 0;
+    // Find an open spot
+    while (mem_list[i].ptr != NULL && i < max_list_size) {
+        i++;
+    }
+
+    if (i > max_list_size) {
```
grgustaf: This should be >= or you will write one past the array.

src/zjs_util.c
```diff
@@ -21,7 +25,50 @@ void *zjs_malloc_with_retry(size_t size)
...
+        ZJS_PRINT("zjs_push_mem_stat no memory stat slots available\n");
```
grgustaf: Not so important since this is debug, but we've stopped putting function names in strings, to save space. The message is plenty unique w/o that.

src/zjs_util.c
```diff
@@ -10,6 +10,10 @@
 #ifndef ZJS_LINUX_BUILD
 #include "zjs_zephyr_port.h"
 #endif
+#ifdef ZJS_TRACE_MALLOC
+#define max_list_size 300
+mem_stats_t mem_list[max_list_size];
```
grgustaf: Hmm, this isn't a list, it's an array. :)

src/zjs_util.h
```diff
@@ -15,6 +15,16 @@
 #include "zjs_common.h"
 #include "zjs_error.h"
 
+#ifdef ZJS_TRACE_MALLOC
+typedef struct mem_stats {
+    void *ptr;
+    char *file;
+    char *func;
+    int line;
+    struct mem_stats *next;
```
grgustaf: Looks like you started out with this as a list and later changed to an array, probably so you wouldn't need dynamic allocations? So you can drop the `next` field now and better to rename `mem_list` to `mem_array`, `mem_in_use`,  `mem_tracker`, or something.

src/zjs_util.h
```diff
@@ -39,12 +53,15 @@ void *zjs_malloc_with_retry(size_t size);
 #define zjs_malloc(sz)                                                      \
     ({                                                                      \
         void *zjs_ptr = zjs_malloc_with_retry(sz);                          \
-        ZJS_PRINT("%s:%d: allocating %lu bytes (%p)\n", __func__, __LINE__, \
+        ZJS_PRINT("%s:%d: allocating %u bytes (%p)\n", __func__, __LINE__,  \
                   (u32_t)sz, zjs_ptr);                                      \
+        char *func_copy = malloc(strlen(__func__) + 1);                     \
+        strcpy(func_copy, __func__);                                        \
+        zjs_push_mem_stat(zjs_ptr, __FILE__, func_copy, __LINE__);          \
```
grgustaf: I don't think you need to copy the function name like this, which is nice - it will save RAM while you're using this feature. When you use __func__, it will cause the compiler to store the function name as a static string in the program's data segment. So you can pass it around as a const char * and it will always be there; then you don't need to free it.

So basically here remove func_copy and replace it with __func__ in the push call. Then make sure to stop freeing it above, just NULL it out instead. And you probably need to make it const char *func in the header file.

src/zjs_util.c
brianjjones: heh, it started as a list and the name never changed :)

src/zjs_util.h
brianjjones: Indeed no need for next.

grgustaf:
+1

grgustaf:
cool patch, good idea

PR #1362, opened 2017-07-13
[travis] Fix broken build by using deprecated setting

grgustaf:
This is a temporary workaround until we figure out how to solve the
problem of "import yaml" no longer working in Python 3 scripts.

I'd prefer we not merge this and solve that instead; otherwise we'll
have masked over it and won't ever fix it until something worse goes
wrong.


grgustaf:
#1364 solves this in a better way. Closing.

PR #1365, opened 2017-07-14
Support enc28j60 with virtual deps

grgustaf:


src/zjs_ocf.json
```diff
@@ -14,34 +15,27 @@
...
         "all": [
             "CONFIG_NETWORKING=y",
             "CONFIG_NET_IPV6=y",
             "CONFIG_NET_UDP=y",
             "CONFIG_TEST_RANDOM_GENERATOR=y",
             "CONFIG_INIT_STACKS=y",
-            "CONFIG_NET_NBUF_RX_COUNT=14",
-            "CONFIG_NET_NBUF_TX_COUNT=14",
+            "CONFIG_NET_BUF=y",
+            "CONFIG_NET_PKT_RX_COUNT=5",
```
jimmy-huang: Can you tell me the reason you picked these numbers for OCF? 5, 2, 10, 6?  A working configuration?

grgustaf: These are the defaults from the echo server sample in Zephyr. I guess an alternative would be to leave them out entirely and take whatever defaults there are.

But the lines above are no longer valid so these were not actually being set at all.

jimmy-huang:
lgtm

PR #1366, opened 2017-07-14
[net-config] Fix BLE addresses given that top two bits are always set

grgustaf:
For some reason, perhaps a good one, the top two bits of the BLE address
are always set regardless of what we try to set it to. So this patch will
warn the user about that and update the address we print out in debug
mode.


jimmy-huang:
+1

PR #1370, opened 2017-07-18
[ws] Work around to fix 1369

cuiyanx:


jimmy-huang:
I rather we fix the WS instead of adding this work-around, maybe @brianjjones can take a look at the bug? If we are going to use this work-around for now, at least add a comment above the code and reference the bug# so we can fix it later.

cuiyanx:
OK, wait for fix bug, I will close this PR.

Issue #1345, opened 2017-07-11
[uart] only write the first four letters after update Zephyr 1.8.0


haoxli:
# Description
After update Zephyr to 1.8.0, there are only first 4 letters can output in console window when write a buffer. 

# Test Code
[UART.js](https://github.com/01org/zephyr.js/blob/master/samples/UART.js)
```
uart.write(new Buffer('UART write succeeded, echoing input.\r\n'));
```

# Steps to Reproduction
1. `$ make JS=samples/UART.js `
2. `$ make dfu`
3. Reset the board
4. In another Terminal: `minicom -D /dev/ttyACM0` 

# Expected Result
![uart-1](https://user-images.githubusercontent.com/3271257/28055103-1ad92278-664a-11e7-9722-816fe3288292.png)

# Actual Result
![uart-2](https://user-images.githubusercontent.com/3271257/28055123-2d126314-664a-11e7-9240-b3ea9c49b649.png)
Enter a string "abc", output is only the first letters of "Data recv: abc"
![uart-3](https://user-images.githubusercontent.com/3271257/28055176-6aca843e-664a-11e7-8669-829eb5250efb.png)

# Test Builds
Branch | Build ID | Target Device| Date | Result
---|---|---|---|---
master | 0334065| Arduino 101| July 11, 2017 | Fail
master | 4e44cc6| Arduino 101| July 11, 2017 | Pass

brianjjones:
Apparently the A101 would occationally repeat bytes if more than 4 bytes were sent at a time using CDC_ACM, so Zephyr made it so you could only send 4 bytes at once (ZEP-2074).  I filed a bug (jira.zephyrproject.org/browse/ZEP-2401) to track this.  In the meantime I created a fix for UART that causes it to keep sending 4 bytes at a time until the entire buffer is sent. See https://github.com/01org/zephyr.js/pull/1377

qiaojingx:
Verified with commit af7eadd + #1377. This issue is not reproducible.

PR #1377, opened 2017-07-18
[uart] Fix for issue #1345, only 4 bytes sent to UART

brianjjones:
Due to ZEP-2074, the Arduino 101 is limited to sending 4 bytes of data at a time via CDC_ACM.  This fix addresses that limitation.


grgustaf:
+1

Issue #1326, opened 2017-06-30
[SPI] There are only zero's in the reply buffer on K64F


qiaojingx:
### Description

 When running [SPI.js](https://github.com/01org/zephyr.js/blob/master/samples/SPI.js) on FRDM-K64F, only seeing zero's in the reply buffer.

### Test Code

[SPI.js](https://github.com/01org/zephyr.js/blob/master/samples/SPI.js);
 
### Steps to Reproduction

 1. `make JS=samples/SPI.js BOARD=frdm_k64f`
 2. `cp outdir/frdm_k64f/zephyr.bin /media/<username>/MBED/`

### Actual Result

 only seeing zero's in the reply buffer:
![selection_086](https://user-images.githubusercontent.com/19813546/27718584-6a4c4bba-5d7f-11e7-943a-eadc9a27083b.png)
### Expected Result

Reply content is correct. 

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | c2ca034 | FRDM-K64F | Jun 30, 2017 | Fail

brianjjones:
This is possibly a Zephyr issue, I've opened https://jira.zephyrproject.org/browse/ZEP-2353 and will work with them to find out if that's the case.

brianjjones:
This issue was due to the fact that K64f uses SPI_0 not SPI_1 for communication, I updated the example to check the board and use the correct SPI bus.

qiaojingx:
Verified with commit 5726f4e. This issue is fixed.

PR #1380, opened 2017-07-19
[SPI] Fix for #1326 - Only zeros on SPI with K64F

brianjjones:

### DIFF
--- a/samples/SPI.js
+++ b/samples/SPI.js
@@ -6,33 +6,37 @@ console.log("SPI test starting..");
 try {
     var spi = require("spi");
     var board = require('board');
-    // Arduino 101 uses SPI_1 while FRDM_K64F uses SPI_0
+    // Arduino 101 uses SPI_1 while others use SPI_0
     var busNum = board.name === "arduino_101" ? 1 : 0;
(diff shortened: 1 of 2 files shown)


samples/SPI.js
```diff
@@ -5,7 +5,10 @@
 console.log("SPI test starting..");
 try {
     var spi = require("spi");
-    var spiBus1 = spi.open({speed:20000, bus:1, polarity:0, phase:0, bits:16});
+    var board = require('board');
+    // Arduino 101 uses SPI_1 while FRDM_K64F uses SPI_0
+    var busNum = board.name === "arduino_101" ? 1 : 0;
+    var spiBus1 = spi.open({speed:20000, bus:busNum, polarity:0, phase:0, bits:16});
 
     var buffer = spiBus1.transceive(1, "Hello World\0");
```
cuiyanx: Please change "1" to "busNum". And the following is the same as this.

brianjjones: That 1 is not the bus number, it is the device number (slave select).  I've updated the sample to make this clearer.

grgustaf:
+1

PR #1388, opened 2017-07-21
[docs] Removed onup event from net, now in net-config

jprestwo:


jimmy-huang:
The one on net module is onnetup and onnetdown, can you update the docs to reflect that?  We'll need to update all the samples and tests too.

grgustaf:
+1

PR #1391, opened 2017-07-21
[websockets] Fix "cannot acquire send_pkt" on Arduino 101

### LINKED ISSUE #1386: [ws] Error thrown when running websocket on A101
### Description

When running [WebSocketServer.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer.js) on arduino 101,  `(ERROR) cannot acquire send_pkt` will be thrown.

### Test Code

[WebSocketServer.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer.js);
[NodeWebSocketClient.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/NodeWebSocketClient.js);
 
### Steps to Reproduction

1. Run [WebSocketServer.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer.js) on a101
2. Run [NodeWebSocketClient.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/NodeWebSocketClient.js) on linux

### Actual Result

Got `(ERROR) cannot acquire send_pkt` error on Arduino 101:
selection_125

 ### Expected Result

No error thrown. 

### Test Builds

---|---|---|---|---
master | 5c65f11 | Arduino101 | Jul 21, 2017 | Fail

jprestwo:
With limited memory, the A101 was running into two problems:

1. The number of net pkt's was set to 2, which, in the 'ping' case
   of the web socket server, was not enough.

2. The timer handle was not being cleared once the timeout expired.
   Normally this is ok but since it was being saved in the global
   scope it was never getting garbage collected.

Fixes #1386

### DIFF
--- a/samples/websockets/WebSocketServer.js
+++ b/samples/websockets/WebSocketServer.js
@@ -36,6 +36,7 @@ wss.on('connection', function(ws) {
                 ws.ping(new Buffer("PING"));
             }
             count++;
+            clearTimeout(t);
         }, 1000);
     });
     ws.on('ping', function(data) {
--- a/samples/websockets/WebSocketServer4.js
+++ b/samples/websockets/WebSocketServer4.js
@@ -39,6 +39,7 @@ net_cfg.dhcp(function(address, subnet, gateway) {
                     ws.ping(new Buffer("PING"));
                 }
                 count++;
+                clearTimeout(t);
             }, 1000);
         });
         ws.on('ping', function(data) {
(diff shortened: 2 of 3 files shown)


jimmy-huang:
+1

jprestwo:
This did not actually fix the problem, just caused memory to be freed explicitly so the test could run longer. The end result was the same though. Closing.

PR #1392, opened 2017-07-21
[promise] Use built-in JobQueue implementation

jimmy-huang:
Update JerryScript and use the new Promise jobqueue implementation
as it is now built-in to JerryScript, thus no longer need to
provide our own job queue, which hopefully fix some of the bugs.


grgustaf:
+1

PR #1394, opened 2017-07-24
[pme] Change 'JSON' to 'json' for building

### LINKED ISSUE #1393: Key 'JSON' is not in deps list
### Description

This is a regression issue. Building [PMESaveRestore.js](https://github.com/01org/zephyr.js/blob/master/samples/PMESaveRestore.js) and [test-pme-save-restore.js](https://github.com/01org/zephyr.js/blob/master/tests/test-pme-save-restore.js) failed for a101.

### Test Code

[PMESaveRestore.js](https://github.com/01org/zephyr.js/blob/master/samples/PMESaveRestore.js);
[test-pme-save-restore.js](https://github.com/01org/zephyr.js/blob/master/tests/test-pme-save-restore.js)
 
### Steps to Reproduction

 1. `make JS=samples/PMESaveRestore.js`

### Actual Result

selection_128
 
### Expected Result

 Building succeed.

### Test Builds

---|---|---|---|---
master | 5c65f11 | Arduino 101 | Jul 24, 2017 | Fail
master | af7eadd | Arduino 101 | Jul 18, 2017 | Pass

cuiyanx:
Fix #1393 

### DIFF
--- a/src/zjs_json.json
+++ b/src/zjs_json.json
@@ -1,6 +1,6 @@
 {
     "module": "json",
-    "object": "JSON",
+    "object": "json",
     "zjs_config": [],
     "src": [],
     "jrs_config": ["CONFIG_DISABLE_JSON_BUILTIN"]


jprestwo:
-1

This will break at runtime. The `object` property value is used to do a regex on the JS code to look if the `JSON` object is being built. Analyze needs to change in the new virtual deps part to not index by `dlist`, because the dependency name is not always the same name as whats in the tree.

@grgustaf, I think a quick and easy if is to check:
`
if i in tree:
`
before checking if `virtualdeps` is in `tree[i]`

grgustaf:
I believe #1395 instead is the right way to fix this.

PR #1405, opened 2017-07-25
[aio] Increase sampling_delay to 30

jimmy-huang:
This will significantly increase the latency of ADC read
from 100ms to 10ms.


jimmy-huang:
closing this

PR #1406, opened 2017-07-25
[net] Fix issue with multiple client connections

jprestwo:
If 2+ clients connected, if either disconnected they would both be
disconnected by the server.


jimmy-huang:
+1

PR #1426, opened 2017-08-03
[spi] New SPI screen module and SPI fixes

brianjjones:
This adds the ST7735.js module, which controls the SPI based
ST7735 LCD.  It also has a few fixes to SPI that were discovered
during development.  An example is also included, SPI_Screen.js

### DIFF
--- a/src/zjs_spi.c
+++ b/src/zjs_spi.c
@@ -92,6 +92,13 @@ static ZJS_DECL_FUNC(zjs_spi_transceive)
         }
     }
 
+    // If a buffer is provided, check that it is a valid type
+    if (argc > 1 && (!jerry_value_is_array(argv[1]) &&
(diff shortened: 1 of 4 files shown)


src/zjs_spi.c
```diff
@@ -215,7 +224,7 @@ static ZJS_DECL_FUNC(zjs_spi_open)
 
     // Default values
     u32_t bus = 0;
-    double speed = 10;
+    double speed = 2;
```
jimmy-huang: this is wrong as on the frdm_k64f, the value will be interpreted as baud rate, and a baud rate of 2 will not work.  In fact, the clock speed should be used here, see #1449

brianjjones: Likely because K64 uses the new SPI API while A101 uses legacy. I'll test out your changes, thanks.

PR #1449, opened 2017-08-09
[spi] Fix incorrect spi speed bus set

jimmy-huang:
The SPI interface in Zephyr seems to accept a different
input for the max_sys_freq when configuring the bus, on
the Arduino 101, it is expecting it to be the clock divider
and on the FRDM-K64F, it is expecting it to be the baud rate.

Also removes configs that are not needed as it is automatically
set by Zephyr when you turn on SPI.

### DIFF
--- a/docs/spi.md
+++ b/docs/spi.md
@@ -48,7 +48,7 @@ interface SPI {
 
 dictionary SPIOptions {
     octet bus;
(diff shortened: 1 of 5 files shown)


brianjjones:
LGTM +1

PR #1452, opened 2017-08-10
[spi] Fix spi unit tests

### LINKED ISSUE #1445: [spi] Read and write 8 bits data failed if set 16 bits for SPI
### Description

When running [test-spi.js](https://github.com/01org/zephyr.js/blob/master/tests/test-spi.js), read and write 8 bits data failed if set 16 bits for SPI.
 
### Test Code

[test-spi.js](https://github.com/01org/zephyr.js/blob/master/tests/test-spi.js);
 
### Steps to Reproduction

 1. `make JS=tests/test-spi.js`
 2. `make dfu`
### Actual Result

 selection_148

### Expected Result

 No error 

### Test Builds

---|---|---|---|---
master | ca7af311 | Arduino 101 | Aug 9 2017 | Fail
master | ca7af311 | FRDM-K64F | Aug 9 2017 | Fail

jimmy-huang:
Fixes #1445


jimmy-huang:
@cuiyanx We think  that this tests would actually be invalid, see comment on #1445, I think if you are doing 16bit words, then sending and reading 3 bytes over SPI would be unexpected behavior.

cuiyanx:
@jimmy-huang OK，I will remove this test point, but I think if someone do this, he will get some error.

jimmy-huang:
@cuiyanx maybe we should throw an exception or something if someone trying to do this, but I am not sure, I think it'll be nice if someone more familiar with SPI bus or the spec to determine this, maybe @brianjjones can follow up later.

jimmy-huang:
I am dropping this patch for now

PR #1460, opened 2017-08-11
[websockets] Increase net buf data size to work around issues

### LINKED ISSUE #1385: [ws] If disconnect ws connection, you will never create a new connection
### Description

I disconnect ws connection and will re-connect, but ws server and client look like be hang. So I close ws client and get an error.
 

### Test Code

[/samples/websockets/WebSocketServer.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer.js)
[/samples/websockets/NodeWebSocketClient.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/NodeWebSocketClient.js)

 

### Steps to Reproduction

1. Run WebSocketServer.js on a101
2. Run NodeWebSocketClient.js on linux
3. Disconnect this ws connection
4. Re-connect ws connection
5. Close ws client
6. Check log info.
 

### Actual Result

Error info:
```
 src/zjs_event.c:358 zjs_trigger_event():
(ERROR) callback_id not found
```

### Expected Result

 Work well and can be re-connected.

### Test Builds

---|---|---|---|---
master | 5c65f1102 | Arduino 101 | July 21, 2017 | Fail

grgustaf:
There appears to be a bug in Zephyr's net buf fragmentation code,
or else this shouldn't make a difference. Need to look into that
further.

Also, clean up some code and output in the node sample code.

Fixes an unreported issue where the websocket sample doesn't work at
all on A101 w/ Ethernet currently.

Fixes #1385


samples/websockets/NodeWebSocketClient.js
```diff
@@ -2,17 +2,19 @@
 
 var WebSocket = require('ws');
 
+console.log("Starting WebSocket Client...\n");
```
jimmy-huang: You don't need to use \n for new line

jimmy-huang:
+1

PR #1462, opened 2017-08-11
[sensor] Fixed polling frequency not updated

### LINKED ISSUE #1451: [sensor] incorrect update frequency / sampling rate of the Sensor API implementation
The [IDL of the ZJS Sensor interface](https://github.com/01org/zephyr.js/blob/master/docs/sensors.md#web-idl) states the sampling rate 'frequency' is in Hz, however if we changed the sampling rate of the [Accelerometer](https://github.com/01org/zephyr.js/blob/master/samples/BMI160Accelerometer.js), [Gyroscope](https://github.com/01org/zephyr.js/blob/master/samples/BMI160Gyroscope.js), or [Temperature](https://github.com/01org/zephyr.js/blob/master/samples/BMI160Temperature.js) sample to 1 Hz, the actual update frequency is faster than every 1 second.

jimmy-huang:
The poll frequency was declared as a static variable and
couldn't be set from a different src file, making it extern
so it could be set.

Fixes #1451

### DIFF
--- a/arc/src/arc_sensor.c
+++ b/arc/src/arc_sensor.c
@@ -20,6 +20,8 @@
 #define BMI160_NAME BMI160_DEVICE_NAME
 #endif
 
+u32_t sensor_poll_freq = 20;        // default polling frequency
+
 #ifdef BUILD_MODULE_SENSOR_LIGHT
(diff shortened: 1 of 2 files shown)


grgustaf:
+1

PR #1464, opened 2017-08-14
[fs] Fixed invalid FSstat object when file is deleted

### LINKED ISSUE #1205: [fs] Remove and overwritten file failed on K64F
### Description

`Remove file` and `file was overwritten with 'w'` got result 'FAIL' when running [test-fs.js](https://github.com/01org/zephyr.js/blob/master/tests/test-fs.js) on K64F.
 
### Test Code

[test-fs.js](https://github.com/01org/zephyr.js/blob/master/tests/test-fs.js)
 
### Steps to Reproduction

1. `make BOARD=frdm_k64f JS=tests/test-fs.js`
2. cp outdir/frdm_k64f/zephyr.bin /media/<USERNAME>/MBED/
  
### Actual Result

 Remove and overwritten file failed
selection_025

### Expected Result

 All test case pass.

### Test Builds

---|---|---|---|---
master | 6180fec | FRDM-K64F | May 31, 2017 | Fail
 
### Additional Information

Not reproduced on Arduino 101

jimmy-huang:
We are not currently checking if the file exists when
calling FS.statSync() to create the Stats object, in which
case if the file doesn't exists, we should return a undefined
value, and the unit tests should also check if the value
is undefined.  Added a ToDo to properly fix this with adding
a FS.exist() api that can be used to check if the file exists,
and then return a valid FSStat object when the file is invalid
according the NodeJS spec.

Fixes #1205

### DIFF
--- a/src/zjs_fs.c
+++ b/src/zjs_fs.c
@@ -326,6 +326,10 @@ static ZJS_DECL_FUNC_ARGS(zjs_fs_unlink, u8_t async)
     }
 
     ret = fs_unlink(path);
+    if (ret != 0) {
+        ERR_PRINT("failed to unlink %s with error [%d]\n", path, ret);
+        return ZJS_UNDEFINED;
+    }
 
 #ifdef ZJS_FS_ASYNC_APIS
     if (async) {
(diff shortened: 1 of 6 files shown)


src/zjs_fs.c
```diff
@@ -326,6 +326,10 @@ static ZJS_DECL_FUNC_ARGS(zjs_fs_unlink, u8_t async)
     }
 
     ret = fs_unlink(path);
+    if (ret != 0) {
+        ERR_PRINT("failed to unlink %s with error [%d]\n", path, ret);
```
grgustaf: Let's change this to DBG, that way we save the string ROM unless the user is specifically looking into weird behavior and turns debug on. This is a good policy for very rare errors I think.

src/zjs_fs.c
```diff
@@ -783,15 +787,26 @@ static ZJS_DECL_FUNC_ARGS(zjs_fs_stat, u8_t async)
     struct fs_dirent entry;
 
     ret = fs_stat(path, &entry);
+    if (ret != 0) {
+        // ToDo: Decide what to do with FStat if the file doesn't exists,
```
grgustaf: Change this to TODO just to match others in case we grep for them.

grgustaf:
+1

PR #1469, opened 2017-08-15
[gpio] Be more clear about holding a reference to pin object

grgustaf:


jimmy-huang:
+1

PR #1470, opened 2017-08-15
[docs] Move to a common Node.js-like format for documenting events

### LINKED ISSUE #1403: [net-config] Please optimize all API docs
### Description

In aio:
```
AIOPin.on
void on(string eventType, ReadCallback callback);
```
In ble:
```
BLE.on
void on(string eventType, EventCallback callback);
```
In gpio:
```
GPIOPin.onchange
attribute ChangeCallback onchange;
```
In net-config:
```
NetConfig.onup Event
void onup(void);

NetConfig.ondown Event
void ondown(void);
```
What is rule for event? If I only read api docs, I will can not use it. 
So, please optimize all API docs. 

### Test Builds

---|---|---|---|---
master | 5726f4e | Arduino 101 | July 25, 2017 | Fail

grgustaf:
Fixes #1403.

### DIFF
--- a/docs/ble.md
+++ b/docs/ble.md
@@ -37,7 +37,6 @@ interface BLE: EventEmitter {
     void startAdvertising(string name, string[] uuids, string url);
     void stopAdvertising();
     void setServices(PrimaryService services[]);
-    void updateRssi();
     PrimaryService PrimaryService(PrimaryServiceInit init);
     Characteristic Characteristic(CharacteristicInit init);
     Descriptor Descriptor(DescriptorInit init);
@@ -112,13 +111,6 @@ for success, otherwise for an error.
 Emitted when a BLE client has disconnected. `clientAddress` will be the same as
 one previously sent with the 'accept' event.
 
-### Event: 'rssiUpdate'
-
-* `int` `rssiValue`
-
-Emitted periodically with most recent RSSI (received signal strength indication)
-value measured in dBm. *NOTE*: currently rssiValue is the dummy value -50.
-
 ### Event: 'stateChange'
 
 * `string` `newState`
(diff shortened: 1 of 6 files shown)


docs/ble.md
```diff
@@ -92,6 +89,48 @@ interface Characteristic {
...
 API Documentation
 -----------------
+BLE is an [EventEmitter](./events.md) with the following events:
+
+### Event: 'accept'
+
+* `string` `clientAddress`
+
+Emitted when a BLE client has connected. *NOTE*: currently the address is the
+dummy string "AB:CD:DF:AB:CD:EF".
```
jimmy-huang: This is not true anymore, it now has a random fake client address generated from Zephyr.

docs/ble.md
```diff
@@ -92,6 +89,48 @@ interface Characteristic {
...
+* `int` `status`
+
+Emitted when BLE services have begun to be advertised. The `status` will be 0
+for success, otherwise for an error.
+
+### Event: 'disconnect'
+
+* `string` `clientAddress`
+
+Emitted when a BLE client has disconnected. *NOTE*: currently the address is the
```
jimmy-huang: same here, it will at least match the address of the connected client.

docs/ble.md
```diff
@@ -92,6 +89,48 @@ interface Characteristic {
...
+dummy string "AB:CD:DF:AB:CD:EF".
+
+### Event: 'rssiUpdate'
+
+* `int` `rssiValue`
+
+Emitted periodically with most recent RSSI (received signal strength indication)
+value measured in dBm. *NOTE*: currently rssiValue is the dummy value -50.
+
+FIXME: @jimmy-huang, while connect/disconnect are still sort of useful with a
```
jimmy-huang: real address it not possible until a client is initiating a pairing process, I think this would be something we can ask for Zephyr Bluetooth team to see if they can add this.

jimmy-huang: I am okay with removing the RSSI value, since we didn't support it and originally we just want to match the Johnly5 API.

docs/ble.md
grgustaf: But this is not the actual address that shows up on the wire?

docs/ble.md
grgustaf: OK I made those updates but leave it to you to actually remove the RSSI function from ble.c; please make a note to do that.

jimmy-huang:
+1

Issue #1187, opened 2017-05-22
[net] Throw out an error before the server responds


cuiyanx:
### Description

I test socket.connect function as client and find socket connection throw out error so fast, the server hasn't had time to respond yet. So, the socket connection will be failed when first connecting.
 

### Test Code

[/samples/TCPClient6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPClient6.js)

 

### Actual Result

If first connecting, socket connection will be failed.
 

### Expected Result

Socket connection wait respondent from server until time out, and throw out error.
 

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 82ef9f89c | Arduino 101 | May 22, 2017 | Fail

grgustaf:
I cannot seem to reproduce this problem. I tested after applying my patch from #1474.

qiaojingx:
Verified with commit 06cce05. It won't go into function `if` in error since error name is not `NotFoundError`. So this issue will not reproduce. But if modify `NotFoundError` to `Error`, still can see this issue.
![selection_138](https://user-images.githubusercontent.com/19813546/29451222-6bb84540-8434-11e7-94df-0cc7a461c2e0.png)

grgustaf:
OK, I finally see what you mean. First, to help reproduce this in the future, I needed to patch tests/tools/tests-tcp6-server.py to use port 4242 which TCPClient6.js is expected. Then start that on your host, and boot up this system.

The problem is that the first time it shows the "Server not found" message instead of working; it's only on retry that it starts working.

Now that I fully understand the issue we can lower the priority and don't need to fix this for 0.4.

qiaojingx:
Verified with commit b32969f on Arduino101, FRDM-K64F and Qemu.
This issue is reproduced contiunously on Arduino101 over Ethernet. But on FRDM-K64F and Qemu, it occurs alternately.

grgustaf:
Try to reserve the word `Verified` for when the fix works and the problem is solved. Here you `Tested` it but it still wasn't fixed.

PR #1474, opened 2017-08-15
[net] Clean up TCPClient6.js sample

grgustaf:
Move connection options to a struct near the top of the file, with
the IP addresses and ports used called out. Add a print statement at
the top to identify the sample.

Also, change the server address from the link local address of James's
machine to the normal 2001:db8::2 address we use.


grgustaf:
I made these cleanups while trying to test #1187, and I don't seem to reproduce the problem, so I think it has been fixed.

jimmy-huang:
+1

PR #1475, opened 2017-08-16
Fix messy files on Ashell

jimmy-huang:


src/zjs_fs.json
```diff
@@ -1,6 +1,7 @@
 {
     "module": "fs",
     "require": "fs",
+    "depends": ["buffer"],
```
brianjjones: Why is this being added?  I don't see buffer used in the sample.  Otherwise looks good.

jimmy-huang: The reason is the the C module actually uses the zjs_buffer apis, although the JS never uses the buffer object directly.  If you build a sample without explicitly importing buffer, then you would get a complication error.

brianjjones: Ah got it, thanks for the clarification.  +1

grgustaf: We ought to figure out a way to have the C portion get included without the JS API overhead in these cases.

PR #1477, opened 2017-08-17
[ashell] Added command to erase flash file system

### LINKED ISSUE #1193: [ashell] Messy filename display after pasting long code in ashell then reset device
### Description

If paste long code via `load test.js` command in ashell, it will stuck. Then, reset device and connect to terminal again, it should work fine. But files stored in ashell will display messily after execute `ls`. 
 
### Test Code

[ashell](https://github.com/01org/zephyr.js/blob/master/docs/ashell.md); 
 
### Steps to Reproduction

1. `make ashell ROM=256`
2. `make dfu`
3. Connect to terminal via `minicom -D /dev/ttyACM0`.
4. `acm> load test.js` and Enter
5. `RAW> /* Paste long code (e.g. `[HeartRateDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/HeartRateDemo.js)`) */`
6. reset device and reconnect again.
7. `acm> ls`
### Actual Result

After step 5. it will stuck and no response by CTRL+Z or CTRL+D. This issue is related to #783 
After step 7. Files stored in ashell display messily.
selection_018
I have no way to restore it.
 
### Expected Result

After step 5. Not stuck. CTRL+Z or CTRL+D will finish coding in RAW.
After step 7. files stored in ashell display as filename.

### Test Builds

---|---|---|---|---
master | 82ef9f8 | Arduino 101 | May 23, 2017 | Fail

jimmy-huang:
Added ability to erase the flash in Ashell/IDE, this
will help fix corrupted files created on the file system.

Fixes #1193

### DIFF
--- a/docs/ashell.md
+++ b/docs/ashell.md
@@ -86,7 +86,7 @@ Command list:
     rm     FILE    Remove file or directory
     mv     F1 F2   Move a file F1 to destination F2
     clear          Clear the terminal screen
-    erase          Erase flash file system
+    erase          Erase flash file system and reboot
     boot   FILE    Set the file that should run at boot
     reboot         Reboot the device
 ```
@@ -190,8 +190,8 @@ Clears the terminal screen.
 
 Erase flash file system.
 
-Note: this only works on the Arduino 101 and it will require rebooting the
-device to take affect.
+Note: this only works on the Arduino 101 and it will automatically rebooting the
+board after erase is finished.
 
 ### boot
 
--- a/docs/ashell.md
+++ b/docs/ashell.md
@@ -86,6 +86,7 @@ Command list:
     rm     FILE    Remove file or directory
     mv     F1 F2   Move a file F1 to destination F2
     clear          Clear the terminal screen
+    erase          Erase flash file system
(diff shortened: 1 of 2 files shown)


docs/ashell.md
```diff
@@ -183,6 +184,15 @@ Renames filename1 to filename2.
...
 Clears the terminal screen.
 
+### erase
+
+`erase`
+
+Erase flash file system.
+
+Note: this only works on the Arduino 101 and it will require rebooting the
+device to take affect.
```
grgustaf: affect -> effect

What do you mean by this? It looks like it happens right away; do you mean that you think their view of filesystem may be inconsistent after the operation if they don't reboot? If so, how about we reboot for them...

jimmy-huang: I believe their FS still has caches of the handles and everything, because I've ran a simple LS command and it's still listing all the files after I call erase.  However, after I reboot the board, it doesn't show up anymore, so maybe it's like stored in RAM or something, or maybe we need to call some FS init api again

brianjjones:
+1

grgustaf:
+1

Issue #1468, opened 2017-08-15
[tcp] 'no handle found' error happens when running TCPEchoServ4_DHCP.js


qiaojingx:
### Description

This is a regression issue. When running [TCPEchoServ4_DHCP.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ4_DHCP.js), got `no handle found ` error on FRDM-K64F.

### Test Code

[TCPEchoServ4_DHCP.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ4_DHCP.js);

### Steps to Reproduction

1. Build [TCPEchoServ4_DHCP.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ4_DHCP.js) on FRDM-K64F.
3. Connect to a router which assigns a DHCP address
4. `echo "hello" | nc <DHCP ip> 4242`

### Actual Result

no handle found error occurs when try to connect with net-tools. if try to connect many times, it also can reproduce bus fault error.
![selection_163](https://user-images.githubusercontent.com/19813546/29310051-e67f0172-81de-11e7-9937-3884bcaf88ec.png)

### Expected Result

No error
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | e3e4dde- 46d73b5 | FRDM-K64F | Aug 15, 2017 | Fail
master | 9f75ded | FRDM-K64F | Aug 15, 2017 | Pass

### Additional Information

Not sure if it is the same as #1458 ,but they are caused by the same commit 46d73b5.

jimmy-huang:
I am also seeing this when I run the BLE.js sample on the nRF52 board, but not on the Arduino 101:

BLE sample...

src/zjs_event.c:397 zjs_emit_event():
(ERROR) no handle found

qiaojingx:
Verified with commit 06cce05 on FRDM_K64F. `Cannot send data to peer(-9)` is print instead of `No handle found` error when send data by net-tools. But `BUS FAULT` error still occurs after connect with net-tools for many times.
![selection_144](https://user-images.githubusercontent.com/19813546/29595143-428b0204-87e8-11e7-9062-379fb8fa8034.png)
![selection_145](https://user-images.githubusercontent.com/19813546/29595161-5ae35108-87e8-11e7-94e3-fbf71cd3bff7.png)

qiaojingx:
[...]

PR #1480, opened 2017-08-18
Dhcp crash fix

grgustaf:


grgustaf:
This doesn't solve the intended problem yet, but I would like some review of what I've done here to rule out bugs in this part.

src/zjs_callbacks.c
```diff
@@ -149,6 +151,22 @@ static zjs_callback_id new_id(void)
...
+typedef struct deferred_work {
+    zjs_deferred_work callback;
+    u32_t length;  // length of user data
+    char data[0];  // user data
+} deferred_work_t;
+
+static void deferred_work_callback(void *handle, const void *args) {
+    const deferred_work_t *deferred = (const deferred_work_t *)args;
+
+    if (deferred->callback) {
```
jimmy-huang: deffered could be NULL

src/zjs_callbacks.c
```diff
@@ -567,3 +587,18 @@ u8_t zjs_service_callbacks(void)
     UNLOCK();
     return serviced;
 }
+
+void zjs_defer_work(zjs_deferred_work callback, const void *buffer, u32_t bytes)
+{
+    DBG_PRINT("deferring work: %d bytes\n", bytes);
+    int len = sizeof(callback) + bytes;
+    char buf[len];
```
jimmy-huang: if len is large number,  you could corrupt the stack

grgustaf: It won't be a large number, that's under our control.

grgustaf: But I can make a comment requiring that in the docs.

src/zjs_callbacks.c
grgustaf: Not really; the callback is only used by `zjs_defer_work`, and that always allocates at least the four bytes for the user callback function pointer. Someone could call it wrong later, but that's their bug and problem... and they'll find it when the run on the likes of K64F. :) In this constrained space, it doesn't make sense to have unnecessary NULL checks, though extra checking may be considered good practice in desktop/server programming. IMHO.

grgustaf: By this rationale I could potentially remove the NULL check we're looking at here. :) But that one can be caused by users of zjs_defer_work so I wanted to be a little more obvious. And it might get dropped in release build where DBG_PRINT does nothing.

grgustaf:
Fixed quite a few things here, along with some improvements I made to help me debug. It does pretty much fix #1468. However, there is at least one problem left that I know of. Eventually I do still get a "BUS FAULT" due to some corruption of a net_pkt object. Its first four bytes are supposed to point to the next buffer in a k_mem_slab "free list". But the last one in the slab of memory they are allocated from always gets this "next" pointer overwritten by some other thread in the system, eventually.

At one point I was able to prove to myself that it was happening in another thread because from the main thread I printed the value out 20x in a row, and in between a couple of the prints, it changed. It typically seems to get overwritten with a pointer into the ROM code region, almost as if a stack return pointer was getting written there for some reason.

It only seems to happen when using DHCP. We don't have a lot of code for DHCP, we pretty much just call Zephyr. So I suspect there's actually something wrong in Zephyr DHCP or net code. But I should still try to review carefully our DHCP code to rule that out.

Running the nc command against the DHCP sample I've seen it now connect and disconnect around 40 times before it gets the fault. Changing that sample to just call setStaticIP though I watched it connect up to 300x with no failure. And with a hack to Zephyr that makes it forget about the existence of the last net_pkt object it's allocated (which is always the one that gets corrupted) it was able to connect 600x w/o getting a fault.

src/zjs_net.c
```diff
@@ -593,33 +594,39 @@ static void add_socket_connection(jerry_value_t socket,
...
+} accept_connection_t;
+
+// a zjs_deferred_work callback
+static void accept_connection(const void *buffer, u32_t length)
 {
-    net_handle_t *handle = (net_handle_t *)user_data;
-    sock_handle_t *sock_handle = NULL;
+    if (length != sizeof(accept_connection_t)) {
+        // shouldn't happen so make it a debug print
+        DBG_PRINT("Error: Unexpected data!");
```
jimmy-huang: add \n

brianjjones:
LGTM +1

PR #1492, opened 2017-08-23
[net-config] Fix two bugs in setStaticIP

grgustaf:
The logic was wrong so that it wouldn't work for IPv4, and the
config flag had a capitalization typo.


jimmy-huang:
+1

PR #1494, opened 2017-08-23
[SPI] New SPI screen module and SPI fixes

brianjjones:
New module for the ST7735 screen which uses SPI for communication.  Also some minor SPI fixes.

### DIFF
--- a/src/zjs_spi.c
+++ b/src/zjs_spi.c
@@ -93,6 +93,13 @@ static ZJS_DECL_FUNC(zjs_spi_transceive)
         }
     }
 
+    // If a buffer is provided, check that it is a valid type
+    if (argc > 1 && (!jerry_value_is_array(argv[1]) &&
+                      !jerry_value_is_string(argv[1]) &&
+                      !zjs_value_is_buffer(argv[1]))) {
+        return SYNTAX_ERROR("SPI transceive buffer is an invalid type");
+    }
+
     jerry_size_t dir_len = 13;
     char dir_str[dir_len];
     // Set the direction default based on the topology.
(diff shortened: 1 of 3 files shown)


src/zjs_spi.c
```diff
@@ -171,15 +177,19 @@ static ZJS_DECL_FUNC(zjs_spi_transceive)
                                tx_buf->bufsize, rx_buf->buffer,
                                rx_buf->bufsize) != 0) {
                 jerry_release_value(rx_buf_obj);
+                jerry_release_value(tx_buf_obj);
```
grgustaf: This is a good case for using ZVAL_MUTABLE to declare tx_buf_obj above (and iniitalize it to 0). Then it will be automatically freed on all return paths and you can remove four copies of jerry_release_value down here.

The "MUTABLE" part is what allows you to overwrite it in those different paths above. We try to avoid using it because regular ZVAL is safer, but this is the kind of case where you need it. (The immutable version uses `const` underneath so you can't overwrite it.)

You should also be able to use it for rx_buf_obj and initialize that to 0 (represents a 0 integer) instead. (On one path it gets set to jerry_create_null() even though it already had been.)

src/zjs_spi.c
```diff
@@ -93,6 +93,13 @@ static ZJS_DECL_FUNC(zjs_spi_transceive)
         }
     }
 
+    // If a buffer is provided, check that it is a valid type
+    if (argc > 1 && (!jerry_value_is_array(argv[1]) &&
+                      !jerry_value_is_string(argv[1]) &&
+                      !zjs_value_is_buffer(argv[1]))) {
+        return SYNTAX_ERROR("SPI transceive buffer is an invalid type");
+    }
```
grgustaf: This should not be needed - because of the ZJS_VALIDATE_ARGS call above, if you have an argv[1] it will have to be one of these three types specified there.

grgustaf:
+1

grgustaf:
New modules didn't need to go into 0.4 but shouldn't hurt anything. Fixes are useful though.

PR #1497, opened 2017-08-24
[ocf] Fixed OCF encoding issues

### LINKED ISSUE #1489: [ocf] The data type does not support arrays and decimal and string array
### Description

I set properties of resource as various types of data to test data transmission. I can not get the correct data that data type are arrays and decimal and string array. I always get "undefined".
 

### Test Code

Sample:
[/samples/thermostat/th02-ocf.js](https://github.com/01org/zephyr.js/blob/master/samples/thermostat/th02-ocf.js)

Simple test as server:
```
var ocf = require("ocf");
var server = ocf.server;

console.log("Test OCF server");

var testProperties = {
    num: 1024,
    negative: -1024,
    decimal: 80.1234,
    array: [512, 1024],
    boolean: true,
    string: "test",
    stringArray: ["test", "string", "array"]
}

var resourceInit = {
    resourcePath: "/test/data",
    resourceTypes: ["test.data"],
    interfaces: ["/oic/if/rw"],
    discoverable: true,
    observable: true,
    secure: true,
    slow: false,
    properties: testProperties
}

function onRetrieve(request, observe) {
    request.respond(testProperties);
}

server.on("retrieve", onRetrieve);

server.register(resourceInit).then(function(resource) {
    console.log("Registered resource test data");
}).catch(function(error) {
});

ocf.start();

```
 
Simple test as client:
```
var ocf = require('ocf');
var client = ocf.client;

ocf.start();

client.findResources({ resourceType:"test.data" }).then(function(resource) {
    console.log("findResources 'test.data' was successful, deviceId=" + resource.deviceId);

    client.retrieve(resource.deviceId, {observable: false}).then(function(res) {
        console.log("Resource path:" + res.resourcePath);
        console.log("Resource data:");
        console.log("    num: " + res.properties.num);
        console.log("    negative: " + res.properties.negative);
        console.log("    decimal: " + res.properties.decimal);
        console.log("    array: " + res.properties.array);
        console.log("    boolean: " + res.properties.boolean);
        console.log("    string: " + res.properties.string);
        console.log("    stringArray: " + res.properties.stringArray);
    }, function(error) {
    });
}, function(error) {
    console.log("findResources() returned an error: " + error.name);
});

```

Log:
```
findResources 'test.data' was successful, deviceId=7b9a6109-cb9a-4109-5b9b-61096b9b6109
Resource path:/test/data
Resource data:
    num: 1024
    negative: -1024
    decimal: undefined
    array: undefined
    boolean: true
    string: test
    stringArray: undefined

```
### Actual Result

Get "undefined".
 

### Expected Result

Get the correct data.
 

### Test Builds

---|---|---|---|---
master | 06cce05 | Arduino 101 | Aug 23, 2017 | Fail

jimmy-huang:
Added support for decimal values and arrays for property values in
OCF responses, this fixes a issue where different array types will
have wrong value when encoded or missing proper encoding.  Also
fixed the client to properly re-create the JS object from decoded
reponses.

Fixes #1489

### DIFF
--- a/src/zjs_ocf_client.c
+++ b/src/zjs_ocf_client.c
@@ -136,6 +136,10 @@ static char *create_url(const char *uuid, const char *path)
 static jerry_value_t get_props_from_response(oc_client_response_t *data)
 {
     jerry_value_t prop_object = jerry_create_object();
+    int *i_array;
+    double *d_array;
+    bool *b_array;
+    u32_t size;
     int i;
     oc_rep_t *rep = data->payload;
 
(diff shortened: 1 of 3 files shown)


src/zjs_ocf_client.c
```diff
@@ -155,18 +159,74 @@ static jerry_value_t get_props_from_response(oc_client_response_t *data)
...
             zjs_obj_add_string(prop_object, oc_string(rep->value.string),
                                oc_string(rep->name));
             DBG_PRINT("%s\n", oc_string(rep->value.string));
             break;
+        case INT_ARRAY:
+            DBG_PRINT("[ ");
+            i_array = oc_int_array(rep->value.array);
+            size = oc_int_array_size(rep->value.array);
+            if (size > 0) {
+                ZVAL array = jerry_create_array(size);
```
jprestwo: I think you would still want to create/add the array if the `size == 0`. I guess the question is does OCF allow you to send a zero length array. If it does then you should add an empty array to JS.

src/zjs_ocf_common.c
```diff
@@ -158,11 +158,36 @@ void *zjs_ocf_props_setup(jerry_value_t props_object, CborEncoder *encoder,
...
+                    }
+                } else if (jerry_value_is_boolean(element)) {
+                    bool boolean = jerry_get_boolean_value(element);
+                    zjs_rep_encode_int(&child, boolean);
+                    DBG_PRINT("Encoding boolean: %d\n", boolean);
+                } else if (jerry_value_is_string(element)) {
+                    ZJS_GET_STRING(element, str, OCF_MAX_PROP_NAME_LEN);
+                    zjs_rep_encode_string(&child, str);
+                    DBG_PRINT("Encoding string %s\n", str);
+                }
```
jprestwo: It *should* be relatively simple to add an object case here. You can just recursively call this function. Maybe I am overlooking the complexity though.

jimmy-huang: For objects, yeah, probably, the problem I am solving here is here is this function all works based on name-value pairs, so it doesn't work for an array which doesn't have any name keys, for example
var testProperties = {
    num: 1024,
    negative: -1024,
    decimal: 80.1234,
    array: [512, 1024],
}

jimmy-huang: Yeah, i added a TODO in there to see if we can further optimize it later.

src/zjs_ocf_client.c
jimmy-huang: Good question, maybe we need to look at the OCF spec to see if an empty array is allowed.

grgustaf:
+1

PR #1499, opened 2017-08-25
[util] Add documentation for ZJS_GET_HANDLE

grgustaf:
Also, wrap CHECK_REF_COUNT in DEBUG_BUILD ifdef so it doesn't find its
way into production.


brianjjones:
LGTM +1

PR #1500, opened 2017-08-25
[SPI] Fixing RX obj so that it doesn't get freed upon return.

brianjjones:


grgustaf:
+1

PR #1501, opened 2017-08-25
[ocf] Much better way to do object encoding/decoding

jprestwo:
Using some code from #1497, this further improves the object
encoding/decoding. Instead of doing 2 passes on the object
to encode, this just does one. Also added support for arrays
of objects.

### DIFF
--- a/src/zjs_ocf_client.c
+++ b/src/zjs_ocf_client.c
@@ -117,175 +117,91 @@ static char *create_url(const char *uuid, const char *path)
     return url;
 }
 
-/*
- * Get properties in 'data' and convert them to a jerry_value_t
- */
-static jerry_value_t get_props_from_response(oc_client_response_t *data)
+#ifdef DEBUG_BUILD
+#define do_print \
+    do_prefix(p); \
+    ZJS_PRINT \
+
+static void do_prefix(int p)
 {
-    jerry_value_t prop_object = zjs_create_object();
-    int *i_array;
-    double *d_array;
-    bool *b_array;
-    u32_t size;
     int i;
-    oc_rep_t *rep = data->payload;
-
-    DBG_PRINT("properties:\n");
-
-    while (rep != NULL) {
-        DBG_PRINT("Type: %u, Key: %s, Value: ", rep->type,
-                  oc_string(rep->name));
-        switch (rep->type) {
-        case BOOL:
-            zjs_obj_add_boolean(prop_object, oc_string(rep->name),
-                                rep->value.boolean);
-            DBG_PRINT("%d\n", rep->value.boolean);
-            break;
-        case INT:
-            zjs_obj_add_number(prop_object, oc_string(rep->name),
-                               (double)rep->value.integer);
-            DBG_PRINT("%d\n", rep->value.integer);
-            break;
-        case DOUBLE:
-            zjs_obj_add_number(prop_object, oc_string(rep->name),
-                               (double)rep->value.double_p);
-            DBG_PRINT("%lf\n", rep->value.double_p);
-            break;
-        case BYTE_STRING:
-        case STRING:
-            zjs_obj_add_string(prop_object, oc_string(rep->name),
-                               oc_string(rep->value.string));
-            DBG_PRINT("%s\n", oc_string(rep->value.string));
-            break;
-        case INT_ARRAY:
-            DBG_PRINT("[ ");
-            i_array = oc_int_array(rep->value.array);
-            size = oc_int_array_size(rep->value.array);
-            if (size > 0) {
-                ZVAL array = jerry_create_array(size);
-                for (i = 0; i < size; i++) {
-                    ZVAL val = jerry_create_number((double)i_array[i]);
-                    jerry_set_property_by_index(array, i, val);
-                    DBG_PRINT("%d \n", i_array[i]);
-                }
-                zjs_obj_add_object(prop_object, oc_string(rep->name), array);
-            }
-            DBG_PRINT("]\n");
-            break;
-        case DOUBLE_ARRAY:
-            ZJS_PRINT("[\n");
-            d_array = oc_double_array(rep->value.array);
-            size = oc_double_array_size(rep->value.array);
-            if (size > 0) {
-                ZVAL array = jerry_create_array(size);
-                for (i = 0; i < size; i++) {
-                    ZVAL val = jerry_create_number(d_array[i]);
-                    jerry_set_property_by_index(array, i, val);
-                    DBG_PRINT("%lf \n", i, d_array[i]);
-                }
-                zjs_obj_add_object(prop_object, oc_string(rep->name), array);
-            }
-            DBG_PRINT("]\n");
-            break;
-        case BOOL_ARRAY:
-            DBG_PRINT("[\n");
-            b_array = oc_bool_array(rep->value.array);
-            size = oc_double_array_size(rep->value.array);
-            if (size > 0) {
-                ZVAL array = jerry_create_array(size);
-                for (i = 0; i < size; i++) {
-                    ZVAL val = jerry_create_boolean(b_array[i]);
-                    jerry_set_property_by_index(array, i, val);
-                    DBG_PRINT("%d \n", b_array[i]);
-                }
-                zjs_obj_add_object(prop_object, oc_string(rep->name), array);
-            }
-            DBG_PRINT("]\n");
-        case STRING_ARRAY:
-            DBG_PRINT("[ ");
-            u32_t size = oc_string_array_get_allocated_size(rep->value.array);
-            if (size > 0) {
-                ZVAL array = jerry_create_array(size);
-                for (i = 0; i < size; i++) {
-                    if (oc_string_array_get_item_size(rep->value.array, i) > 0) {
-                        ZVAL val = jerry_create_string(oc_string_array_get_item(rep->value.array, i));
-                        jerry_set_property_by_index(array, i, val);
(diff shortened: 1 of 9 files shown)


grgustaf:
I'll plan to leave this for consideration by @jimmy-huang

jimmy-huang:
@jprestwo can you rebase?

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                jerry_release_value(array);
             }
-            // Close the array
-            zjs_rep_close_array(enc, &child);
-            DBG_PRINT("]\n");
+            break;
+        case OBJECT:
+            {
+                jerry_value_t new_object = zjs_create_object();
+                decode_value_helper(new_object, rep->value.object);
```
jimmy-huang: decode_value_helper() returns a object?  Could this be leak here if you don't free it?

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
-                } else if (jerry_value_is_string(element)) {
-                    ZJS_GET_STRING(element, str, OCF_MAX_PROP_NAME_LEN);
-                    zjs_rep_encode_string(&child, str);
-                    DBG_PRINT("Encoding string %s\n", str);
+            break;
+        case BOOL_ARRAY:
+            {
+                sz = oc_bool_array_size(rep->value.array);
+                bool *b_array = oc_bool_array(rep->value.array);
+                jerry_value_t array = jerry_create_array(sz);
```
jimmy-huang: maybe use ZVAL so you don't have to free it?

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                for (i = 0; i < sz; i++) {
+                    jerry_value_t b = jerry_create_number(b_array[i]);
```
jimmy-huang: ZVAL

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                zjs_set_property(object, oc_string(rep->name), array);
+                jerry_release_value(array);
+            }
+            break;
+        case INT_ARRAY:
+            {
+                sz = oc_int_array_size(rep->value.array);
+                int *i_array = oc_int_array(rep->value.array);
+                jerry_value_t array = jerry_create_array(sz);
+                for (i = 0; i < sz; i++) {
```
jimmy-huang: ZVAL array

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                    jerry_value_t integer = jerry_create_number(i_array[i]);
```
jimmy-huang: ZVAL integer

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                    oc_string(rep->value.string));
+            break;
+        case DOUBLE:
+            zjs_obj_add_number(object, oc_string(rep->name),
+                    rep->value.double_p);
+            break;
+        case STRING_ARRAY:
+            {
+                sz = oc_string_array_get_allocated_size(rep->value.array);
+                jerry_value_t array = jerry_create_array(sz);
```
jimmy-huang: ZVAL array

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                for (i = 0; i < sz; i++) {
+                    jerry_value_t str = jerry_create_string(
```
jimmy-huang: ZVAL

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
     }
 }
 
-static bool ocf_foreach_prop(const jerry_value_t prop_name,
-                             const jerry_value_t prop_value,
-                             void *data)
+static void zjs_ocf_encode_value_helper(jerry_value_t object, const char *name,
+                                 CborEncoder *encoder);
```
jimmy-huang: align here

src/zjs_ocf_client.c
```diff
@@ -401,15 +316,16 @@ static void observe_callback(oc_client_response_t *data)
         struct client_resource *resource =
             (struct client_resource *)data->user_data;
         jerry_value_t resource_val = create_resource(resource);
-        ZVAL properties_val = get_props_from_response(data);
+        ZVAL properties_val = zjs_ocf_decode_value(data->payload);
+        //ZVAL properties_val = get_props_from_response(data);
```
jimmy-huang: remove the unused code if not needed

src/zjs_ocf_client.c
```diff
@@ -117,175 +117,90 @@ static char *create_url(const char *uuid, const char *path)
...
+            break;
+        case OBJECT:
+            do_print("%s: {\n", oc_string(rep->name));
+            print_props_data(p + 1, rep->value.object);
+            do_print("}\n");
+            break;
+        case STRING_ARRAY:
+            do_print("%s: [", oc_string(rep->name));
+            for (i = 0;
+                    i < oc_string_array_get_allocated_size(rep->value.array);
```
jimmy-huang: align here

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                }
+                zjs_set_property(object, oc_string(rep->name), array);
+                jerry_release_value(array);
+            }
+            break;
+        case DOUBLE_ARRAY:
+            {
+                sz = oc_double_array_size(rep->value.array);
+                double *d_array = oc_double_array(rep->value.array);
+                jerry_value_t array = jerry_create_array(sz);
```
jimmy-huang: ZVAL array

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                for (i = 0; i < sz; i++) {
+                    jerry_value_t d = jerry_create_number(d_array[i]);
```
jimmy-huang: ZVAL d

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                    return new_object;
+                }
+                zjs_set_property(object, oc_string(rep->name), new_object);
+                jerry_release_value(new_object);
+            }
+            break;
+        case OBJECT_ARRAY:
+            {
+                oc_rep_t *iter = rep->value.object_array;
+                jerry_value_t array = jerry_create_array(0);
```
jimmy-huang: ZVAL

src/zjs_ocf_common.c
```diff
@@ -41,178 +55,232 @@ int zjs_ocf_is_int(jerry_value_t val)
...
+                while (iter) {
+                    jerry_value_t element = decode_value_helper(ZJS_UNDEFINED, iter);
```
jimmy-huang: ZVAL

src/zjs_ocf_common.c
jprestwo: It returns ZJS_UNDEFINED *most* of the time, but in the case of an object nested inside the root object yes it will return a new object. This if statement checks if the object is not named, if it is indeed not named then its part of an array, meaning its iterating the OBJECT_ARRAY case below. In that case the return value gets freed.

Unfortunately having this special case was a bit messy, but the only way I could think of to implement the OBJECT_ARRAY type.

src/zjs_ocf_client.c
jprestwo: I think it is right? its indented 4 spaces past the `i = 0` on the next line. That `oc_string_array_get_allocated_size()` is just quite a long function name.

src/zjs_ocf_common.c
jprestwo: array gets released + re-assigned each iteration of the while loop so its got to be a jerry_value_t

src/zjs_ocf_client.c
jimmy-huang: ok, i would usually indent 4 spaces from the for, or align the i with i = 0, but it's fine

src/zjs_ocf_common.c
jimmy-huang: ok, since i just see that returned obj gets released down further, just checking

src/zjs_ocf_common.c
jimmy-huang: I would expect ZVAL to work also? maybe i am not totally sure how ZVAL works, so I guess we can just leave it as is since it works.

jimmy-huang:
lgtm, +1

src/zjs_ocf_client.c
grgustaf: Huh? No, this looks wonky. 
How about call get_allocated_size on the previous line w/ a tmp var and check against that var to fit on one line.

That will also save you calling that function over and over again, I assume it always gives the same answer?

Do that in the cases below, too.

jprestwo:
@grgustaf, I used a temp variable for the length in that case where the line length was too long, but in the cases below the "get_length" API is different in each case so I couldnt just store it once. Plus the other cases were below 80 characters.

PR #1503, opened 2017-08-29
Net fix server freed

grgustaf:


grgustaf:
I ran this bash command and connected > 450 times w/o a bus fault:

```bash
for i in $(seq 1 1000); do echo -n "$i "; timeout 1 nc $(grep "Got IP" output | tail -1 | cut -d' ' -f3-) 4242; sleep 0.5; done
```

Note, this is after running this command to capture the output of the K64F to the 'output' file, which is how the above command finds the DHCP-assigned address.
```bash
minicom -D /dev/ttyACM0 -C output
```

src/zjs_net.c
```diff
@@ -214,13 +214,11 @@ static void post_closed(void *handle)
             zjs_free(h);
         }
 
-        if (net) {
```
brianjjones: No need to check if server_h exists?

src/zjs_net.c
```diff
@@ -864,22 +862,22 @@ static ZJS_DECL_FUNC(net_create_server)
...
-    server_handle_t *handle = zjs_malloc(sizeof(server_handle_t));
-    if (!handle) {
+    server_handle_t *server_h = zjs_malloc(sizeof(server_handle_t));
+    if (!server_h) {
         jerry_release_value(server);
         return zjs_error("could not alloc server handle");
     }
 
-    handle->server = server;
-    handle->listening = 0;
```
brianjjones: Is this handle->listening = 0; line no longer needed?

brianjjones:
Other than the minor questions (which may be invalid), looks good to me.

src/zjs_net.c
grgustaf: No because I'm `memset`ting the whole struct to 0 on the next line now.

src/zjs_net.c
grgustaf: Yeah, I don't think so. It should always be set if a socket exists; so considered an 'invariant'.

brianjjones:
LGTM +1

Issue #1502, opened 2017-08-28
[ashell] Redundant message outputs


xiaoyu3x:
## Issue description
Redundant message is printed on terminal and it affect the next input.

## Test Code
[ashell](https://github.com/01org/zephyr.js/blob/master/docs/ashell.md)

## Steps to reproduce
1. `make ROM=256 ashell; make dfu`
2. Flash ashell into 101 board, reset it and open port:
    `$ screen /dev/ttyACM0 115200`

## Actual Result
Redundant message is printed.
![selection_001](https://user-images.githubusercontent.com/17694094/29757325-a10e017a-8bdd-11e7-805d-dbef7f322bad.png)

## Expected Result
No redundant message outputs.

## Test Build
Branch| Commit id | Target Device |Test Date | Result 
---| ---| ---| ---|--- 
master | [6c160b5](https://github.com/01org/zephyr.js/commit/6c160b5d77c32b89e0fd165a644b4cd243d01929) | Arduino 101 | Aug 28, 2017 | Fail

brianjjones:
I'm betting its the same thing that causes it to be printed for issue #1458

brianjjones:
This will be fixed by #1510

qiaojingx:
Verified with commit a0b1217 +#1510， this issue is not reproduced.

xiaoyu3x:
This issue will be closed again when PR #1510 is merged.

brianjjones:
Looks like #1576 will be the new fix.

xiaoyu3x:
Verified with commit [be3ebf7](https://github.com/01org/zephyr.js/pull/1576/commits/be3ebf7c43843a99eb7229d7c3700628de666291)+[dd371e7](https://github.com/01org/zephyr.js/pull/1576/commits/dd371e794c65b3ec61100a61493d29f3c5d512f0) +[#1576](https://github.com/01org/zephyr.js/pull/1576)， this issue is not reproduced.

PR #1510, opened 2017-08-30
[events] Don't emit poweredon signal unless Bluetooth has been initialized.

brianjjones:
Fix for #1502 
main.c emits the signal that BT is powered.  However,
In the case of ashell the BT module might not have been
initialized yet, causing the "error no handle found".
So I'm adding a check that the handle has been created.

### DIFF
--- a/src/zjs_ble.c
+++ b/src/zjs_ble.c
@@ -581,9 +581,12 @@ static struct bt_conn_auth_cb zjs_ble_auth_cb_display = {
 
 void zjs_ble_emit_powered_event()
 {
-    const char state[] = "poweredOn";
-    zjs_defer_emit_event(ble_handle->ble_obj, "stateChange", state,
-                         sizeof(state), string_arg, zjs_release_args);
+    // Don't emit if bluetooth hasn't had init called yet
+    if (ble_handle) {
+        const char state[] = "poweredOn";
+        zjs_defer_emit_event(ble_handle->ble_obj, "stateChange", state,
+                             sizeof(state), string_arg, zjs_release_args);
+    }
 }
 
 static ZJS_DECL_FUNC(zjs_ble_disconnect)


src/zjs_ble.c
```diff
@@ -581,9 +581,12 @@ static struct bt_conn_auth_cb zjs_ble_auth_cb_display = {
...
 void zjs_ble_emit_powered_event()
 {
-    const char state[] = "poweredOn";
-    zjs_defer_emit_event(ble_handle->ble_obj, "stateChange", state,
-                         sizeof(state), string_arg, zjs_release_args);
+    // Don't emit if bluetooth hasn't had init called yet
+    if (ble_handle) {
+        const char state[] = "poweredOn";
+        zjs_defer_emit_event(ble_handle->ble_obj, "stateChange", state,
+                             sizeof(state), string_arg, zjs_release_args);
```
grgustaf: Will it be emitted later if we don't do it now?

brianjjones: No it won't.   But it also doesn't work now either, so this at least fixes the error.  Should we emit when the bluetooth handle is created? Or have the bluetooth module keep track of whether this has been emitted yet?

grgustaf: I think this needs to be fixed more fully then so let's leave it open until @jimmy-huang gets back, or we investigate further.

grgustaf:
@brianjjones please update the patch title to be descriptive rather than "Fix for #1502" though. That can be mentioned below, but the title should describe the nature of the fix.

jimmy-huang:
+1

grgustaf:
@jimmy-huang please look at this again, shouldn't the poweredon event be sent after initialization later then, so we don't lose it?

jimmy-huang:
@grgustaf yeah, I guess this patch doesn't solve the problem you mentioned, so I submitted another patch here #1576.

grgustaf:
Superceded by #1576.

PR #1511, opened 2017-08-30
[util] Add strequal macro to replace strcmp for equality tests

grgustaf:
This is more readable and less vulnerable to logic inversion.


brianjjones:
LGTM +1

Issue #1458, opened 2017-08-11
[ws] BUS FAULT error when running dhcp websocket


qiaojingx:
### Description

This is a regression issue. When running websocket-dhcp, got `BUS FAULT` error on FRDM-K64F.

### Test Code

[WebSocketServer4.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer4.js);
[test-ws4-server-dhcp.js](https://github.com/01org/zephyr.js/blob/master/tests/test-ws4-server-dhcp.js);
[test-dhcp-ws-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-dhcp-ws-manual.js);

### Steps to Reproduction

1. Build [WebSocketServer4.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer4.js) on FRDM-K64F.
3. Connect to a router which assigns a DHCP address
4. `node node samples/websockets/NodeWebSocketClient4.js <ip address> <port>`

### Actual Result

BUS FAULT error:
![selection_159](https://user-images.githubusercontent.com/19813546/29206842-4e250d70-7eb6-11e7-8391-16c4139bb491.png)

### Expected Result

No error
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | e3e4dde | FRDM-K64F | Aug 11, 2017 | Fail
master | f3625a8 | FRDM-K64F | Aug 11, 2017 | Pass

grgustaf:
@qiaojingx, just wanted to comment that you guys do a really great job and in particular giving us a commit id where it passed and then failed for a regression is really helpful. :)

qiaojingx:
I found BUS FAULT error begins with commit 46d73b5. A little different is:  `(ERROR) no handle found
` will print when first connect, then bus fault error occurs when second connect.
![selection_160](https://user-images.githubusercontent.com/19813546/29256036-d5651e54-80d9-11e7-9965-ffca4b926b74.png)
Hope it is also helpful. :)

grgustaf:
The no handle found issue should be fixed by #1503; please retest.

grgustaf:
Oops that was net, this is ws - I'll need to check this out. It might well be the same type of problem that I just fixed in net.

brianjjones:
I tested it and see the same thing qiaojingx did.  I didn't get a chance to start trouble shooting it, but its easily reproducible.

[...]

PR #1513, opened 2017-08-31
Ws fix bus fault

grgustaf:
Still not working but I'm looking for some early review. This is working toward fixing #1458.

### DIFF
--- a/src/zjs_web_sockets.c
+++ b/src/zjs_web_sockets.c
@@ -595,41 +595,24 @@ static void create_ws_connection(void *h, jerry_value_t argv[], u32_t *argc,
     *argc = 1;
 }
 
-static void tcp_received(struct net_context *context,
-                         struct net_pkt *pkt,
-                         int status,
-                         void *user_data)
+typedef struct {
+    ws_connection_t *con;
+    struct net_pkt *pkt;
+} receive_packet_t;
+
+// a zjs_deferred_work callback
+static void receive_packet(const void *buffer, u32_t length)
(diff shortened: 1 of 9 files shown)


src/zjs_web_sockets.c
```diff
@@ -173,6 +170,21 @@ static void handle_error_arg(void *unused, jerry_value_t argv[], u32_t *argc,
     *argc = 1;
 }
 
+static void emit_error(jerry_value_t obj, const char *format, ...)
```
brianjjones: Doesn't need to be done now, but would this be something we should put in the utils file?  Would it be useful for other modules?  Just a though.

grgustaf:
This seems to really fix #1458 for all three test scripts now. If you increase max_payload in the scripts the client connection succeeds. However, on test-ws4-server-dhcp.js just now if I connected a few times I started seeing `(ERROR) cannot acquire send_pkt` and then an MPU fault, so you might want to see if you can reproduce that (new bug).

brianjjones:
LGTM +1

cuiyanx:
Verified with d04ae005e + #1513, #1458 fixed. But the test-ws4-server-dhcp.js case will throw out error after connected sometimes.
Log:
```
***** BUS FAULT *****
  Executing thread ID (thread): 0x20003ec8
  Faulting instruction address:  0x12ba6
  Precise data bus error
  Address: 0x676e6960
Fatal fault in essential thread! Spinning...

```

Issue #1507, opened 2017-08-30
[tcp/websocket/udp] Default ble address is missing in tcp/websocket/udp server.


qiaojingx:
### Description

This is a regression issue. When running tcp6 server, websocket6 server and udp6 server on A101. Bluetooth adapter cannot search for ble address of this server.
 
### Test Code

[TCPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ6.js);
[UDPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/UDPEchoServ6.js);
[WebSocketServer.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer.js);
[test-tcp6-server.js](https://github.com/01org/zephyr.js/blob/master/tests/test-tcp6-server.js);
[test-udp6-server.js](https://github.com/01org/zephyr.js/blob/master/tests/test-udp6-server.js);
[test-udp6-server-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-udp6-server-manual.js);
[test-ws6-server.js](https://github.com/01org/zephyr.js/blob/master/tests/test-ws6-server.js)
 
### Steps to Reproduction

1. `make JS=samples/TCPEchoServ6.js ROM=256 `
2. `make dfu`
3. `hcitool lescan` to search ble address of this server.

### Actual Result

Bluetooth adapter cannot search for ble address of this server.
![selection_003](https://user-images.githubusercontent.com/19813546/29859563-4b7c360a-8d95-11e7-92fa-ef99111ece0e.png)

### Expected Result

Should get server address like:
![selection_004](https://user-images.githubusercontent.com/19813546/29859696-e8713b5e-8d95-11e7-9a27-24543ac8722c.png)

 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | bf70e36 - 6c160b5 | Arduino 101 | Aug 30, 2017 | Fail
master | d3f0434 | Arduino 101 | Aug 30, 2017 | Pass
### Additional Information

If add the following code in these samples, this issue will not reproduce.
```
var net_cfg = require('net-config');
net_cfg.setBleAddress('FF:EE:DD:CC:BB:AA');
```
Is that by design that default ble address be removed? if so, please update sample to set ble address.

brianjjones:
[...]

PR #1514, opened 2017-09-01
[net] Fix for #1507 - Default ble address

brianjjones:
The default address was actually being set, but there
were other methods being called that have been moved
to the main file, and needed to be removed otherwise it caused
havoc.


src/zjs_net_config.c
```diff
@@ -33,9 +33,6 @@ void zjs_net_config_default(void)
 #ifdef CONFIG_NET_L2_BLUETOOTH
     if (!net_ble_enabled) {
         zjs_init_ble_address();
-        ipss_init();
```
brianjjones: These are now called in main.c:212.  Calling them here causes net to hang.

grgustaf:
+1

PR #1523, opened 2017-09-08
Net mutex

grgustaf:


src/zjs_common.h
```diff
@@ -33,39 +33,77 @@
...
+ *
+ * FTRACE takes format string and arguments like printf
+ * FTRACE_JSAPI takes not arguments and expects to run in a ZJS_DECL_FUNC API
+ */
+#ifdef USE_FTRACE
+#define FTRACE                                                   \
+    ZJS_PRINT("[%x] %s: %s: ", (u32_t)k_current_get() & 0xffff,  \
+              FTRACE_PREFIX, __func__);                          \
     ZJS_PRINT
+#define FTRACE_JSAPI                                             \
```
jprestwo: Could this just move into ZJS_DECL_FUNC? Then you it would be in all modules without having to call it everywhere.

grgustaf: Not the way ZJS_DECL_FUNC is done currently because it doesn't include the opening brace of the function, so you can't add statements to the function. It could perhaps go into ZJS_VALIDATE_ARGS though.

src/zjs_common.h
```diff
@@ -33,39 +33,40 @@
  */
 char *zjs_shorten_filepath(char *filepath);
 
-#ifdef DEBUG_BUILD
+// enable to use asserts even in release mode, e.g. for terse debugging
+#if 0
+#define USE_ASSERT
+#endif
```
brianjjones: Wouldn't this be better to add as a make flag rather than requiring editing the header file?

jprestwo: Yeah, it would be really nice if you could specify a comma separated list of all the different debug options, and/or 'all' for all of them.

grgustaf: You can feel free to submit such a patch. :)

I don't think it makes much of a difference actually, this is very easy. It's on by default in DEBUG anyway, usually you would not want this on in release build.

But FYI for FTRACE I very specifically designed it the way I did so that you can turn on function tracing per-file by making the define in that file, instead of enabling it for the whole project and get way too much output. Yes there are fancier ways to handle it, again - feel free. This is quick and good enough for what I need.

jprestwo:
+1

PR #1524, opened 2017-09-11
[websockets] Move max payload check to check data payload

jprestwo:
The check for max_payload was checking the total TCP packet
length rather than just the websocket data payload length.
This caused the initial accept packet to be rejected if
the maxPayload property was set to less than 276.


brianjjones:
LGTM +1

grgustaf:
+1

PR #1528, opened 2017-09-11
[SPI] Adding an acquire when using provided buffer

brianjjones:
If the passed arg is a Buffer object, I don't want to have
ZVAL free the original reference when its done, because I don't
own it. Instead, create a local reference that gets freed.

### DIFF
--- a/src/zjs_spi.c
+++ b/src/zjs_spi.c
@@ -160,7 +160,7 @@ static ZJS_DECL_FUNC(zjs_spi_transceive)
         } else {
             // If we were passed a buffer just use it as is
             tx_buf = zjs_buffer_find(argv[1]);
-            tx_buf_obj = buffer;
+            tx_buf_obj = jerry_acquire_value(buffer);   // We don't own the buffer
         }
         // If this is a read / write
         if (dir_arg == ZJS_TOPOLOGY_FULL_DUPLEX) {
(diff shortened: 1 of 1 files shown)


src/zjs_spi.c
```diff
@@ -160,7 +160,7 @@ static ZJS_DECL_FUNC(zjs_spi_transceive)
         } else {
             // If we were passed a buffer just use it as is
             tx_buf = zjs_buffer_find(argv[1]);
-            tx_buf_obj = buffer;
+            tx_buf_obj = jerry_acquire_value(buffer);   // We don't own the buffer
```
grgustaf: While this does make it work correctly, you never actually use this tx_buf_obj again... so even better is just remove this line entirely.

grgustaf: Also, this isn't really a case where ZVAL_MUTABLE is needed, and it's a little dangerous so it would be better to remove it. You could do that by declaring tx_buf_obj as ZVAL in each of the two other cases above, and removing the function-wide declaration above.

The rx_buf_obj case, where it's needed at the very end in the return statement, but can be created in various places, is the kind of case where we *do* need ZVAL_MUTABLE.

brianjjones: Sounds good I'll make the changes.

PR #1529, opened 2017-09-12
[net] Distinguish some identical debug messages; print rvals

grgustaf:

### DIFF
--- a/src/zjs_net.c
+++ b/src/zjs_net.c
@@ -1218,7 +1218,7 @@ static ZJS_DECL_FUNC(socket_connect)
                               &handle->tcp_sock));
     }
     if (!handle->tcp_sock) {
-        DBG_PRINT("connect failed\n");
+        DBG_PRINT("failed to get context\n");
         error_desc_t desc = create_error_desc(ERROR_CONNECT_SOCKET, this,
                                               function_obj);
         zjs_defer_emit_event(this, "error", &desc, sizeof(desc),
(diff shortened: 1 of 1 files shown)


src/zjs_net.c
```diff
@@ -1253,10 +1253,12 @@ static ZJS_DECL_FUNC(socket_connect)
...
-                sizeof(peer_addr6), tcp_connected, 1, handle) < 0) {
-            DBG_PRINT("connect failed\n");
+        int rval = net_context_connect(handle->tcp_sock,
+                                       (struct sockaddr *)&peer_addr6,
+                                       sizeof(peer_addr6), tcp_connected,
+                                       1, handle);
+        if (rval < 0) {
+            DBG_PRINT("IPv6 connect failed: %d\n", rval);
             zjs_obj_add_boolean(this, false, "connecting");
             error_desc_t desc = create_error_desc(ERROR_CONNECT_SOCKET, this,
```
jprestwo: you didn't edit this line but what is this doing? Its just returning so `desc` isn't being used. Shouldn't the `error` event get emitted like it does below?

grgustaf: Did you get confused because the patch cuts off here? It does call defer_emit in the full source.

jprestwo: yes, I only pressed the expand button once. Disregard.

brianjjones:
LGTM +1

PR #1532, opened 2017-09-14
[general] Use sizeof(args) as safer than manual calculation

grgustaf:
Also, restore style of args declaration in another place where I had
just fixed the same issue; after reviewing all of them I see we use
spaces around the braces consistently.


brianjjones:
+1

PR #1534, opened 2017-09-14
[utils] Switch to the use the new UTF8 JrS string functions

grgustaf:
I'm afraid we've been concerned only with ASCII to date but getting
strings as UTF-8 should get us a lot closer to correct behavior in the
face of extended characters.


brianjjones:
Grove LCD does seem like a reasonable one to drop off the list for ashell.  Its not used for any of the IDE provided examples.  +1

PR #1539, opened 2017-09-18
[zephyr] Update to Zephyr 1.9.0

grgustaf:
Also, update iotivity-constrained to build with Zephyr 1.9.

In analyze, add expanding of shell variables used in zephyr_conf
JSON settings to work around an issue with the MBEDTLS_CFG_FILE.

Add a workaround to checkheaders to create a stub file for
generated_dts_board.h that is expected.

Decrease ROM size in trlite builds to allow for slight expansion
in ARC image build size. Also, drop 'fs' module from ashell/ide
builds so that they will fit in the new 255 KB allotted.

Follow rename of CONFIG_BLUETOOTH_* to CONFIG_BT_* in Zephyr.

Slightly reorganize Bluetooth service registration to match new
convention.

Work around some network macros that have gone away.

Fix references to sections.h that moved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,8 @@ cache:
 
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install make gcc g++ python3-ply ncurses-dev uglifyjs libc6-dev-i386 python3-yaml -qq
+  - sudo apt-get install make gcc g++ python3-ply ncurses-dev uglifyjs libc6-dev-i386 python3-pip python3-yaml -qq
+  - sudo pip3 install pyelftools
 
 install: >
   if [ "$TARGET" != "linux" -a "$(cat $ZEPHYR_SDK_INSTALL_DIR/sdk_version)" != "$ZEPHYR_SDK_VERSION" ]; then
--- a/README.md
+++ b/README.md
@@ -67,6 +67,13 @@ change in Zephyr, so it could be left out currently if you don't use k64f.
 Before that, for a while python-yaml was needed when the script was using
 Python 2.
 
+You may also need to install pip3 to get pyelftools:
+
+```bash
+$ sudo apt-get install python-pip3
+$ sudo pip3 install pyelftools
+```
+
 #### Clone the ZJS repo
 Next, clone this git repo:
 ```bash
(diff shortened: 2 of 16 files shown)


grgustaf:
This is not fully working yet to pass Travis, but I think most issues are fixed so it's a starting point for anyone else.

brianjjones:
I did some testing and here is what I found to work / not work
Works:
I2C
SPI
GPIO
ashell and IDE (Can connect, run scripts, etc as expected)
Sensor
FS

Broken:
Websockets

I get the following error:
src/zjs_web_sockets.c:1071 ws_server():
(ERROR) Error in net_context_bind(handle->server_ctx, (struct sockaddr *)&addr, sizeof(struct sockaddr)): -125
Error: net_context_bind(handle->server_ctx, (struct sockaddr *)&addr, sizeof(struct sockaddr))

grgustaf:
Net module seems to be working fine, I think we should get this in. @jprestwo has two PRs that fix the websocket issues.

grgustaf:
@jimmy-huang, please review...

.travis.yml
```diff
@@ -31,7 +31,8 @@ cache:
 
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install make gcc g++ python3-ply ncurses-dev uglifyjs libc6-dev-i386 python3-yaml -qq
+  - sudo apt-get install make gcc g++ python3-ply ncurses-dev uglifyjs libc6-dev-i386 python3-pip python3-yaml -qq
+  - sudo pip3 install pyelftools
```
jimmy-huang: I had to use sudo -H pip3 install here, otherwise i get a error msg

README.md
```diff
@@ -67,6 +67,13 @@ change in Zephyr, so it could be left out currently if you don't use k64f.
 Before that, for a while python-yaml was needed when the script was using
 Python 2.
 
+You may also need to install pip3 to get pyelftools:
+
+```bash
+$ sudo apt-get install python-pip3
+$ sudo pip3 install pyelftools
```
jimmy-huang: same here

jimmy-huang:
+1

.travis.yml
grgustaf: It appears to work fine in Travis, so I don't know that it needs to change; probably a configuration issue.

PR #1542, opened 2017-09-19
[timers] Make timers completely async/event driven

jprestwo:
Timers currently used the main event loop to check if
the timer had expired. This change uses Zephyr's timer
callbacks to signal the JS function. The linux port
uses signals to do the same thing.


grgustaf:
+1

PR #1543, opened 2017-09-19
[build] Make will now error if JS= is given when building ide / ashell

brianjjones:


grgustaf:
+1

PR #1544, opened 2017-09-19
see what update is spitting out

grgustaf:


grgustaf:
This is just a dummy PR to try to debug what's breaking with Travis

PR #1557, opened 2017-09-25
[GFX] Initial commit of the GFX module

brianjjones:
This module is a generic graphics module that can be used by
any screen to draw simple shapes and text. I also updated the
ST7735 module to use it, as well as the SPI_Screen.js exmaple.


docs/gfx.md
```diff
@@ -0,0 +1,124 @@
...
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+The GFX module provides a generic way to create pixel buffers, which can then
+be displayed on a display of some kind.  A Javascript method for initializing
```
kenchris: JavaScript*

docs/gfx.md
```diff
@@ -0,0 +1,124 @@
...
+This IDL provides an overview of the interface; see below for documentation of
+specific API functions.
+
+```javascript
+// require returns a GFX object
+// var gfx = require('gfx');
+
+[NoInterfaceObject]
+interface GFX {
+    GFXObj init(screen width, screen height, init screen function,
```
kenchris: JS uses Object, so I would write it out. but maybe GfxContext makes more sense, especailly since canvas uses Context

docs/gfx.md
```diff
@@ -0,0 +1,124 @@
...
+                draw function, optional this);
+};
+
+[NoInterfaceObject]
+interface GFXObj {
```
kenchris: Why not mimic: 

https://html.spec.whatwg.org/multipage/canvas.html#the-canvas-element
https://html.spec.whatwg.org/multipage/canvas.html#2dcontext

docs/gfx.md
```diff
@@ -0,0 +1,124 @@
...
+    fillRect(number x coord, number y coord, number width, number height,
```
kenchris: void clearRect(unrestricted double x, unrestricted double y, unrestricted double w, unrestricted double h);
  void fillRect(unrestricted double x, unrestricted double y, unrestricted double w, unrestricted double h);

docs/gfx.md
```diff
@@ -0,0 +1,124 @@
...
+             char array color);
+    drawPixel(number x coord, number y coord, char array color);
+    drawLine(number x0 coord, number y0 coord, number x1 coord,
```
kenchris: Doing something like this, would probably make much more sense and it brings more familiarity:

```
interface CanvasDrawPath {
  // path API (see also CanvasPath)
  void beginPath();
  void fill(optional CanvasFillRule fillRule = "nonzero");
  void fill(Path2D path, optional CanvasFillRule fillRule = "nonzero");
...

interface CanvasPath {
  // shared path API methods
  void closePath();
  void moveTo(unrestricted double x, unrestricted double y);
  void lineTo(unrestricted double x, unrestricted double y);
...
Path2D implements CanvasPath;
```

docs/gfx.md
```diff
@@ -0,0 +1,124 @@
...
+             number y1 coord, char array color, optional number size);
+    drawVLine(number x coord, number y coord, number height, char array color,
+              optional number size);
+    drawHLine(number x coord, number y coord, number width, char array color,
+              optional number size);
+    drawRect(number x coord, number y coord, number width, number height,
+             char array color, optional number size);
+    drawChar(number x coord, number y coord, character char, char array color,
+             optional number size);
+    drawString(number x coord, number y coord, string str, char array color,
```
kenchris:
```
interface CanvasText {
  // text (see also the CanvasPathDrawingStyles and CanvasTextDrawingStyles interfaces)
  void fillText(DOMString text, unrestricted double x, unrestricted double y, optional unrestricted double maxWidth);
```

kenchris: I don't say that we need all that is in Canvas, but lets follow their style

This is currently very inconsistent:

Ie
```drawString(number x coord, number y coord, string str, char array color, optional number size);```
vs
```void fillText(DOMString text, unrestricted double x, unrestricted double y, optional unrestricted double maxWidth);```

Comment (kenchris) [reply docs/gfx.md]: Color you set via

```
interface CanvasFillStrokeStyles {
  attribute (DOMString or CanvasGradient or CanvasPattern) fillStyle; // (default black)
}
```

Canvas has save() and restore() for state:

```
interface CanvasState {
  // state
  void save(); // push state on state stack
  void restore(); // pop state stack and restore state
};
```

**Comment by kenchris** [comment]
I think you could get the same basic functionality by just implementing a few of the methods in the 2DCanvas API which is quite familiar to web developers.

It is state based, with a push and pop stack, but if you only allow setting fillColor that should be quite easy to implement.

File: docs/gfx.md (hunk 5 above)
Comment (brianjjones) [reply docs/gfx.md]: Perhaps in the future, but that's more advanced than I was trying to accomplish for right now.

**Comment by brianjjones** [comment]
@kenchris While I can get behind the idea of modeling this after canvas, it may be bit before I have time to make those changes.  But we'll see...

File: src/zjs_gfx.c [hunk 7]
```diff
@@ -0,0 +1,468 @@
...
+    // requires: handle is the native pointer we registered with
+    //             jerry_set_object_native_handle
+    //  effects: frees the gfx module
+
+    gfx_handle_t *handle = (gfx_handle_t *)native;
+    zjs_free(handle);
+}
+
+static const jerry_object_native_info_t gfx_type_info = {
+   .free_cb = zjs_gfx_callback_free
```
grgustaf: I forget, but it may just work to set this to zjs_free directly here rather than adding an extra function?

src/zjs_gfx.c
```diff
@@ -0,0 +1,468 @@
...
+        { zjs_gfx_draw_char, "drawChar" },
+        { zjs_gfx_draw_string, "drawString" },
+        { NULL, NULL }
+    };
+
+    zjs_gfx_prototype = zjs_create_object();
+    zjs_obj_add_functions(zjs_gfx_prototype, proto);
+
+    jerry_value_t gfx_obj = zjs_create_object();
+    //zjs_obj_add_function(gfx_obj, zjs_gfx_set_cb, "init");
```
grgustaf: there are a few of these stray things left commented out

grgustaf:
Yeah, I like the idea of following canvas API too, but let's not hold it up for that. I'll add a feature request.

src/zjs_gfx.c
brianjjones: Whoops thanks I'll clean that up

src/zjs_gfx.c
brianjjones: Not sure I follow, this is how we get JerryScript to call the cleanup when the object is freed.  Its the pattern used by the other modules.  Granted it will change slightly with the Jerry modules patch.

src/zjs_gfx.c
brianjjones: Fixed

src/zjs_gfx.c
brianjjones: Ah I get what you meant now, could we just execute the single line there.  As far as I know, no.  jerry_set_object_native_pointer expects a jerry_object_native_info_t for the last arg, I can't just pass it a function.

src/zjs_gfx.c
```diff
@@ -176,25 +115,28 @@ static jerry_value_t zjs_gfx_fill_rect_priv (u32_t x, u32_t y, u32_t w, u32_t h,
...
         }
     }
 
-    for(int i = 0; i < 5; i++) { // Char bitmap == 5 columns
-        char line = charBuf->buffer[i];
-        for(int j = 0; j < 8; j++) {    // Draw each bit
+    for(int i = 0; i < fontBytes; i+=2) {
+        u16_t line = (((u16_t)charBuf->buffer[i]) << 8) | charBuf->buffer[i+1];
+        for(int j = 0; j < 16; j++) {
             if((line >> j)  & 1) {
```
grgustaf: Take the extra space before the & and put it after the if. ;)

modules/ST7735.js
```diff
@@ -51,106 +52,32 @@ function ST7735() {
...
     }
 
+
     // Send data over SPI
     st7735API.writeData = function(data) {
         this.dcPin.write(1);    // Set DC pin low to indicate data
         this.csPin.write(0);
         this.spiBus.transceive(1, data, "write");
         this.csPin.write(1);
+        //delete (data);
```
grgustaf: ?

src/zjs_gfx.c
```diff
@@ -0,0 +1,417 @@
...
+}
+
+jerry_value_t zjs_gfx_init()
+{
+    zjs_native_func_t proto[] = {
+        { zjs_gfx_fill_rect, "fillRect" },
+        { zjs_gfx_draw_pixel, "drawPixel" },
+        { zjs_gfx_draw_line, "drawLine" },
+        { zjs_gfx_draw_v_line, "drawVLine" },
+        { zjs_gfx_draw_h_line, "drawHLine" },
```
grgustaf: How about just give the user drawLine and you can pick between implementations if you see that x1 == x2 or y1 == y2... I don't think they should really have to think about that.

src/zjs_gfx.c
```diff
@@ -0,0 +1,417 @@
...
+        }
+    }
+    return ret;
+}
+
+static ZJS_DECL_FUNC(zjs_gfx_set_cb)
+{
+    // requires: Requires 4 arguments
+    //           arg[0] - The width of the screen.
+    //           arg[1] - The hight of the screen.
```
grgustaf: height, because spelling is next to cleanliness

src/zjs_gfx.c
```diff
@@ -0,0 +1,417 @@
...
+    //  effects: Draws a character on the screen
+
+    ZJS_VALIDATE_ARGS(Z_NUMBER, Z_NUMBER, Z_STRING, Z_ARRAY, Z_OPTIONAL Z_NUMBER);
+    ZJS_GET_HANDLE(this, gfx_handle_t, handle, gfx_type_info);
+    gfx_data_t argData;
+    args_to_data(&argData, argc, argv);
+    jerry_value_t ret = ZJS_UNDEFINED;
+
+    for (u8_t i = 0; i < argData.textSize; i++) {
+        ret = zjs_gfx_draw_char_priv(argData.coords[0] + (i * 7 * argData.size), argData.coords[1], argData.text[i], argData.color, argData.size, handle);
```
grgustaf: Since you chose a proportional width font, you really should space the characters out based on character width rather than giving them all the same size; it will look a lot better.

I think you can do that pretty easily here by just returning the character width from draw_char_priv and add it to your running x count instead of this i * 7 * size.

modules/ST7735.js
brianjjones: commented code is commented :)  I'll delete it thanks.

src/zjs_gfx.c
brianjjones: I've updated this to be smarter.

src/zjs_gfx.c
brianjjones: Well drawVLine and drawHLine take one less argument, so less typing on the user's part :)
These will all change anyway once I move to the more canvas-like API anyway, so perhaps the change can wait till then?

grgustaf: Yeah, that's fine.

grgustaf:
+1

PR #1559, opened 2017-09-25
[docs] Updating ASHELL docs with note about SPI pins

### LINKED ISSUE #727: [ashell] IO10~IO13 are failed as output pins by zephyrJS IDE
When run test/test-a101-pins.js by zephyr.js IDE, IO10~IO13 are failed as output pins. but they are passed by zephyr JS on a101.

brianjjones:
Fixes #727 - Pins 10-13 can only be used as SPI on ashell
when using the A101.

### DIFF
--- a/docs/ashell.md
+++ b/docs/ashell.md
@@ -14,6 +14,10 @@ arbitrary JavaScript directly from the USB port of a Zephyr enabled device.
 Although the shell would look like a common Unix Bash, it is heavily limited
 due the constraints of the micro-controller.
 
+NOTE: Because the filesystem module is used by ashell, the SPI module is also
+included. This means that on the Arduino 101, pins 10 - 13 are set as SPI and
+can't be used as GPIO pins.
+
 Compilation
 ------------
 


docs/ashell.md
```diff
@@ -14,6 +14,10 @@ arbitrary JavaScript directly from the USB port of a Zephyr enabled device.
 Although the shell would look like a common Unix Bash, it is heavily limited
 due the constraints of the micro-controller.
 
+NOTE: Because the filesystem module is used by ashell, the SPI module is also
```
jimmy-huang: Shouldn't it be easy to add a case to throw an exception in JS and print something if pins 10-13 are set to GPIO when running in ashell?  I think that would be much better than just documenting it here.

brianjjones: Sorry for the slow reply.  It would require adding A101 specific code to GPIO, but yes it wouldn't be difficult.  @grgustaf is that something we'd want to add?

grgustaf: Yeah, stuff like that will probably eventually come out of the board API work so that it could be more general.

grgustaf:
+1

PR #1568, opened 2017-09-28
[webusb] Add new WebUSB module to support URL advertising

grgustaf:

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -29,12 +29,14 @@ I/O
 
 [UART](./uart.md)
 
+[WebUSB}(./webusb.md)
+
 Generic Sensors
 ---------------
 [Sensors](./sensors.md)
 
-Communications
---------------
+Networking
+----------
 [BLE](./ble.md)
 
 [Network Configuration](./net-config.md)
--- a/docs/API.md
+++ b/docs/API.md
@@ -29,7 +29,7 @@ I/O
 
 [UART](./uart.md)
 
-[WebUSB}(./webusb.md)
+[WebUSB](./webusb.md)
 
 Generic Sensors
 ---------------
--- a/docs/webusb.md
+++ b/docs/webusb.md
@@ -28,7 +28,7 @@ API Documentation
 
 The `url` string should begin with "https://" in order for Chrome to accept it
 and display a notification. Other URLs are valid in terms of the protocol but
-will have no user-visible effect currently in Chrome.
+will have no user-visible effect in Chrome.
 
 Sample Apps
 -----------
--- a/docs/webusb.md
+++ b/docs/webusb.md
@@ -1,0 +1,35 @@
+Zephyr.js API for WebUSB
+========================
+
+* [Introduction](#introduction)
+* [Class: WebUSB](#class-webusb)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+The WebUSB module supports advertising the device as a WebUSB device when the
+USB port is connected to a PC. Currently, this only works on Arduino 101 as it
+is the only board with a USB driver in the Zephyr tree.
+
+The API allows you to set the URL that will be suggested to the browser for
+connecting to your device.
+
+When you connect your device to a Linux PC or Mac with Chrome >= 60 running, it
+will give a notification that the device would like you to visit the URL you've
+set. (Windows currently prevents this from working, I believe.)
+
+
+API Documentation
+-----------------
+### WebUSB.setURL
+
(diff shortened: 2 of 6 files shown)


samples/WebUSB.js
```diff
@@ -0,0 +1,14 @@
...
+var webusb = require('webusb');
+
+// On a Linux PC with the Google Chrome >= 60 running, when you plug the device
+//   in via USB, Chrome should detect it and display a notification prompting
+//   you to visit the URL specified below.
+// Note: You should only use https URLs because Chrome will not display a
+//   notification for other URLs.
+
+// suggest that the browser visit this URL to connect to the device
+webusb.setURL("https://github.com/01org/zephyrjs");
```
qiaojingx: I think you mean "https://github.com/01org/zephyr.js".

brianjjones: Perhaps the comment is misleading, but to connect to the device you should go to https://01org.github.io/zephyrjs-ide/

But if what you meant was you want them to go to our github to learn about connecting, that makes sense too.

docs/API.md
```diff
@@ -29,12 +29,14 @@ I/O
 
 [UART](./uart.md)
 
+[WebUSB}(./webusb.md)
```
kenchris: } should be ]

Maybe we should just call it USB ?

docs/webusb.md
```diff
@@ -0,0 +1,35 @@
...
+========================
+
+* [Introduction](#introduction)
+* [Class: WebUSB](#class-webusb)
+* [API Documentation](#api-documentation)
+* [Sample Apps](#sample-apps)
+
+Introduction
+------------
+The WebUSB module supports advertising the device as a WebUSB device when the
```
kenchris: Web USB, not WebUSB (yeah web is inconsistent, but spaces is preferred)

docs/webusb.md
```diff
@@ -0,0 +1,35 @@
...
+USB port is connected to a PC. Currently, this only works on Arduino 101 as it
+is the only board with a USB driver in the Zephyr tree.
+
+The API allows you to set the URL that will be suggested to the browser for
+connecting to your device.
+
+When you connect your device to a Linux PC or Mac with Chrome >= 60 running, it
+will give a notification that the device would like you to visit the URL you've
+set. (Windows currently prevents this from working, I believe.)
```
kenchris: @larsgk will know the details

docs/webusb.md
```diff
@@ -0,0 +1,35 @@
...
+
+API Documentation
+-----------------
+### WebUSB.setURL
+
+`void setURL(string url);`
+
+The `url` string should begin with "https://" in order for Chrome to accept it
+and display a notification. Other URLs are valid in terms of the protocol but
+will have no user-visible effect currently in Chrome.
```
kenchris: not currently, ever :-) Basically no new hardware related API will ever be allowed on http

grgustaf: Yeah I got that message reading more of the spec today. :) I guess another browser / environment could make a different choice though. I'll update that text.

docs/webusb.md
grgustaf: But the spec is called [WebUSB](https://wicg.github.io/webusb/)...

grgustaf:
@brianjjones, the URL in the sample is not meant to go to the IDE, but rather demonstrate how this could be any URL.

This should be ready to go now...

src/zjs_webusb.json
```diff
@@ -0,0 +1,27 @@
+{
+    "module": "webusb",
+    "description": "Module to advertise as a WebUSB device (only A101 has a USB driver in Zephyr currently)",
+    "require": "webusb",
+    "targets": ["arduino_101"],
+    "zephyr_conf": {
+        "all": [
```
jimmy-huang: why is this "all"? the only target is arduino_101, there's no support for k64f, shouldn't these also be a101-specific configs?

grgustaf: The DW one is truly A101-specfic, in that it uses the DW USB hardware. But the others are all software dependencies and would be needed for other platforms; so this is just forward looking to when there is more than one platform with USB support.

jimmy-huang:
+1

docs/webusb.md
larsgk: M61 (stable, sans flag, Windows:no pop-up yet and Android: likewise but slightly different ). Btw - I'd love to get it on the FRDM-KL25Z target as it has a separate USB port that works very well in our ARMmbed WebUSB setups (now that A101 was mentioned)

PR #1570, opened 2017-10-04
[utils] Do not include __FILE__ macro in non debug builds

### LINKED ISSUE #1550: Failed to build IDE image on zjs-0.4
`make ide ROM=256` is failing to build IDE image on zjs-0.4 branch. 

Error:
```
/opt/zephyr-sdk/sysroots/x86_64-pokysdk-linux/usr/libexec/i586-zephyr-elfiamcu/gcc/i586-zephyr-elfiamcu/6.2.0/real-ld: zephyr.elf section `initlevel' will not fit in region `ROM'
/opt/zephyr-sdk/sysroots/x86_64-pokysdk-linux/usr/libexec/i586-zephyr-elfiamcu/gcc/i586-zephyr-elfiamcu/6.2.0/real-ld: region `ROM' overflowed by 140 bytes
collect2: error: ld returned 1 exit status
```

Note: If I increase the ROM size, the ARC image doesn't fit in available space.

nagineni:
Using __FILE__ macro in non debug code is increasing the
image size depending on where the sources are built, so
this patch removes that macro from the non debug code.

Fixes #1550

### DIFF
--- a/src/zjs_common.h
+++ b/src/zjs_common.h
@@ -72,8 +72,7 @@ int zjs_get_ms(void);
 #else  // !DEBUG_BUILD
 #define DBG_PRINT(fmt...) do {} while (0)
 #define ERR_PRINT                                                        \
-    ZJS_PRINT("\n%s:%d %s():\n(ERROR) ", zjs_shorten_filepath(__FILE__), \
-              __LINE__, __func__);                                       \
+    ZJS_PRINT("\n%d:(ERROR) ", __LINE__);                                \
     ZJS_PRINT
 #endif  // DEBUG_BUILD
 


src/zjs_common.h
```diff
@@ -72,8 +72,7 @@ int zjs_get_ms(void);
 #else  // !DEBUG_BUILD
 #define DBG_PRINT(fmt...) do {} while (0)
 #define ERR_PRINT                                                        \
-    ZJS_PRINT("\n%s:%d %s():\n(ERROR) ", zjs_shorten_filepath(__FILE__), \
-              __LINE__, __func__);                                       \
+    ZJS_PRINT("\n%s():%d:\n(ERROR) ", __func__, __LINE__);               \
```
gabrielschulhof: For a non-debug build this should maybe just be empty - just like `DBG_PRINT` above. The name of the function also adds unnecessary bytes.

gabrielschulhof: ... though at least the size is consistent across machines, and independent of where on the file system the zephyr.js repo is located.

nagineni: Error logs are useful for better understanding in case of any failure in non-debug builds, but I do not have a strong feeling whether to keep or strip those logs. The main issue was inconsistent image size based on the sources location and that is now fixed. 

@brianjjones @grgustaf suggestions?

grgustaf: The reason I have it in there is just because it lets the error messages themselves be minimal, like "out of memory" - and the function etc. provides context. But I guess you're right, it would be better to leave that out of non-debug build. Then if you see an error and want more info, rebuild w/ debug. So yeah this change might as well remove __func__ too. I suppose __LINE__ could be left although it will be pretty mysterious w/o func or file. :)

However, the rest of the change below I don't think is necessary - that stuff is all conditional under ZJS_TRACE_MALLOC which is when you build with TRACE=on, not production mode. It's nice to be able to build w/ TRACE but w/o DEBUG cruft.

grgustaf: Now, how could I test for this in CI to make sure we don't accidentally reintroduce the issue? I could build twice in two different directories and compare. ;) But I'd have to do that across all modules so not a very good way. Any ideas?

nagineni: > Then if you see an error and want more info, rebuild w/ debug. So yeah this change might as well remove __func__ too. 

Ok, removed `__func__` also now.

> However, the rest of the change below I don't think is necessary - that stuff is all conditional under ZJS_TRACE_MALLOC which is when you build with TRACE=on, not production mode.

You are right, changes in zjs_util.* are not needed in production mode.

> Now, how could I test for this in CI to make sure we don't accidentally reintroduce the issue? I could build twice in two different directories and compare. ;) But I'd have to do that across all modules so not a very good way. Any ideas?

I can't think of any clean way other than what you mentioned for testing this in the CI. :(

grgustaf:
+1

PR #1571, opened 2017-10-05
[callbacks] Fixing a memory leak for callbacks

brianjjones:
Currently when zjs_timers_cleanup gets called, not all of the memory
in the callbacks is getting freed.  The ring buffer isn't being
freed before the javascript ends.  This change forces all the
memory to be freed when zjs_remove_all_callbacks gets called.

### DIFF
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -369,6 +369,7 @@ void zjs_remove_all_callbacks()
         CB_LOCK();
         if (cb_map[i]) {
             zjs_remove_callback_priv(i, skip_flush);
+            zjs_free_callback(i);
         }
         CB_UNLOCK();
     }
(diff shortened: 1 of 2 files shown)


src/zjs_callbacks.c
```diff
@@ -369,6 +369,7 @@ void zjs_remove_all_callbacks()
         CB_LOCK();
         if (cb_map[i]) {
             zjs_remove_callback_priv(i, skip_flush);
+            zjs_free_callback(i);
```
grgustaf: No, this isn't right... if you look in zjs_remove_callback_prev it may free the callback itself under some circumstances. You don't want to do it twice.

If you need to try to force all cleanup synchronously and can't wait for things to flush out naturally, I think what you probably need to do, which came up in our discussion yesterday, is remove all timers / callbacks and then call zjs_service_callbacks, probably in your ashell cleanup routine I'm guessing? 

How's that for a sentence!

-1 for now

brianjjones: I tried zjs_service_callbacks in several places, but it doesn't solve the issue.  Something in the ring buffer remains at the end. This was the only method I found that successfully got rid of all the references from the callbacks. Afraid I don't understand the ring buffer well, otherwise I'd tell you why its happening.

brianjjones: I could change it so that zjs_remove_callback_priv is passed skip_flush = true, then we'd be sure it didn't get freed 2x

grgustaf:
+1

PR #1574, opened 2017-10-06
RFC: Use JerryScript modules

gabrielschulhof:
This illustrates the use of JerryScript native modules in ZJS by converting `pwm` and `net` to such modules.

PR #1575, opened 2017-10-09
[modules] Migrating native modules to Jerryscript

brianjjones:
This PR will move the handling of loading native modules to the
Jerryscript side of the code.  This will simplify our code, help
catch memory leaks, and make things a little more robust.


brianjjones:
This is a WIP and has several more commits, I'll update the comment when its complete.

gabrielschulhof:
I think you need to compile the modules with -DENABLE_INIT_FINI because that will cause the `JERRYX_NATIVE_MODULE()` macro to expand to a constructor and destructor, rather than to two public symbols. In my example PR I added the flag to the `<module name>.json` file. [`zjs_net.json`]( https://github.com/01org/zephyr.js/pull/1574/files#diff-7b3bb3490238a769f310fc1c031cfa58R29), for example.

You also need to turn on Zephyr's [C++](https://github.com/01org/zephyr.js/pull/1574/files#diff-b61ad63c1ebb3de3ad759decc1d19993R5) support, because that turns on support for constructors and destructors.

src/zjs_modules.c
```diff
@@ -32,28 +33,27 @@ struct routine_map {
...
-            if (mod->instance == 0) {
-                mod->instance = mod->init();
-            }
-            return jerry_acquire_value(mod->instance);
-        }
+    jerry_value_t js_module = jerryx_module_resolve (module, resolvers, 1);
+    // If we found our module, return it.
+    if (!jerry_value_has_error_flag (js_module)) {
+        return js_module;
     }
```
gabrielschulhof: If it *does* have the error flag we need to `jerry_release_value (js_module);` otherwise we leak the error in `js_module`.

src/zjs_pwm.json
```diff
@@ -7,7 +7,5 @@
     },
     "zjs_config": ["-DBUILD_MODULE_PWM"],
```
gabrielschulhof: We need to add "-DENABLE_INIT_FINI" to this list for each module, or to a central place where such flags are specified, otherwise `JERRYX_NATIVE_MODULE ()` will not expand to a constructor/destructor pair, but rather to a pair of global functions.

gabrielschulhof:
I think we should move JerryScript up at least one more commit, to `6d53931055747d54c78620629714ea2e72c3ff30`, because that brings what looks to be the final shape of the module API, and it will give us a leg up if/when we implement loading JS files from disk, via the optional canonical name resolution it introduces.

brianjjones:
@gabrielschulhof is correct, we need a newer version of JerryScript for this PR to work.

src/zjs_modules.c
```diff
@@ -32,28 +33,27 @@ struct routine_map {
 static u8_t num_routines = 0;
 struct routine_map svc_routine_map[NUM_SERVICE_ROUTINES];
 
+    static jerryx_module_resolver_t resolvers[1] =
```
gabrielschulhof: If we go with the final version of the module API, this needs to change to
```c
static const jerryx_module_resolver_t *resolvers[1] = {
  &jerryx_module_native_resolver
};
```
because in the new version of the API the symbol `jerryx_module_native_resolver` refers to a global structure, not a function pointer.

brianjjones: Thanks, updated

src/zjs_modules.c
brianjjones: Updated thanks.

brianjjones:
This PR is ready to be reviewed finally, everything should work as intended.

src/zjs_aio.c
```diff
@@ -302,10 +301,25 @@ static ZJS_DECL_FUNC(zjs_aio_open)
...
+    if (!handle)
+        return zjs_error("could not allocate handle");
+
+      jerry_set_object_native_pointer(pinobj, handle, &aio_type_info);
+
     return pinobj;
 }
 
-jerry_value_t zjs_aio_init()
+void zjs_aio_cleanup()
```
gabrielschulhof: These cleanup functions can now be made static.

src/zjs_ble.c
```diff
@@ -1283,6 +1282,25 @@ static ZJS_DECL_FUNC(zjs_ble_descriptor)
...
+    jerry_release_value(ble_handle->ble_obj);
+    zjs_free(ble_handle);
+    ble_handle = NULL;
+}
+
+static const jerry_object_native_info_t ble_module_type_info = {
+   .free_cb = zjs_ble_cleanup
+};
+
 jerry_value_t zjs_ble_init()
```
gabrielschulhof: These inits can also be made static. If `ENABLE_INIT_FINI` is unavailable `JERRYX_NATIVE_MODULE()` will expand to the definition of two global functions: `<module_name>_register()` and `<module_name>_unregister()`, which will add resp. remove the module from Jerry's linked list of native modules which it knows about. So, in the absence of `ENABLE_INIT_FINI` those can still be used to register the modules with Jerry.

src/zjs_aio.c
brianjjones: Good point, updating...

src/zjs_ble.c
brianjjones: I added it to zjs_common.json.

src/zjs_pwm.json
brianjjones: zjs_common.json is used by all of them, I added it there.

src/zjs_common.json
```diff
@@ -2,10 +2,12 @@
...
         "all": [
+            "CONFIG_CPLUSPLUS=y",
             "CONFIG_STDOUT_CONSOLE=y",
             "CONFIG_NEWLIB_LIBC=y",
             "CONFIG_FLOAT=y",
             "CONFIG_RING_BUFFER=y"
         ]
-    }
+    },
+    "zjs_config": ["-DENABLE_INIT_FINI"]
```
jimmy-huang: isn't this a linux only config?

scripts/analyze
```diff
@@ -363,24 +357,11 @@ def write_modules(list, tree):
...
     void (*cleanup)();
-} gbl_module_t;
-module_t zjs_modules_array[] = {\n""")
-        for i in inits:
-            file.write("    {\"%s\", %s" % (i, inits[i]["inits"][0]))
-            if 'cleanups' in inits[i]:
-                file.write(", %s" % inits[i]["cleanups"][0])
-            file.write("},\n")
-        file.write("};\n")
+} gbl_module_t;\n""")
```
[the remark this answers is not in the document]
brianjjones: hmmm probably.

src/zjs_common.json
brianjjones: No its needed in every version in order to enable JerryScript modules.

src/zjs_event.c
```diff
@@ -460,41 +443,74 @@ void *zjs_event_get_user_handle(jerry_value_t obj)
...
+        zjs_event_create_prototype();
+    }
+    jerry_value_t proto = zjs_event_emitter_prototype;
+    if (jerry_value_is_object(prototype)) {
+        jerry_set_prototype(prototype, proto);
+        proto = prototype;
+    }
+    jerry_set_prototype(obj, proto);
+
+    emitter_t *emitter = zjs_malloc(sizeof(emitter_t));
```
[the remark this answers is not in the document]
brianjjones: Yes. JerryScript calls the cleanup function provided by emitter_type_info when the JS is stopped.

Makefile
```diff
@@ -227,7 +228,7 @@ flash:  analyze generate $(JERRYLIB) $(ARC)
 
 # Build for zephyr, default target
 .PHONY: zephyr
-zephyr: analyze generate $(JERRYLIB) $(ARC)
+zephyr: analyze generate $(JERRYLIB) outdir/$(BOARD)/libjerry-ext.a $(ARC)
```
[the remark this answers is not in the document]
brianjjones: outdir/$(BOARD) is used in a lot of places in the Makefile, which is why I used it.  Should it be made into its own macro?  Unlike jerry-core, jerry-ext is always the same, it doesn't change if we use snapshot.

src/ashell/Makefile
```diff
@@ -6,7 +6,7 @@ endif
 
 # Adding path for jerry script APIs
 DEPS_BASE ?= $(ZJS_BASE)/deps
-ZEPHYRINCLUDE += -I$(JERRY_INCLUDE) -I$(DEPS_BASE)
+ZEPHYRINCLUDE += -I$(JERRY_INCLUDE) -I$(JERRY_BASE)/jerry-ext/include -I$(DEPS_BASE)
```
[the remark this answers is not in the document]
brianjjones: I added it to JERRY_INCLUDE, but the export JERRY_BASE has to remain otherwise ashell doesn't know where JERRY_BASE is.

jimmy-huang:
+1

PR #1576, opened 2017-10-09
[ble] Fixed ble in Ashell

### LINKED ISSUE #1502: [ashell] Redundant message outputs

jimmy-huang:
This fixes the issue where bluetooth doesn't work in Ashell,
since when running in Ashell, bt_enable() cannot be called until
the Bluetooth module is initialized.  This also fixes the error message
showing up when IDE is launched.

Fixes #1502

### DIFF
--- a/src/zjs_ble.c
+++ b/src/zjs_ble.c
@@ -1288,14 +1288,12 @@ static ZJS_DECL_FUNC(zjs_ble_descriptor)
     return jerry_acquire_value(argv[0]);
 }
 
-#ifdef ZJS_ASHELL
 // INTERRUPT SAFE FUNCTION: No JerryScript VM, allocs, or release prints!
-static void ble_bt_ready(int err)
+void ble_bt_ready(int err)
 {
     DBG_PRINT("bt_ready() is called [err %d]\n", err);
     zjs_ble_emit_powered_event();
 }
-#endif
 
 jerry_value_t zjs_ble_init()
 {
--- a/src/zjs_ble.c
+++ b/src/zjs_ble.c
@@ -582,6 +582,7 @@ static struct bt_conn_auth_cb zjs_ble_auth_cb_display = {
 void zjs_ble_emit_powered_event()
 {
     const char state[] = "poweredOn";
+    ZJS_ASSERT(ble_handle, "ble_handle not set");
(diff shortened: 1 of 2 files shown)


src/zjs_ble.c
```diff
@@ -1283,6 +1284,15 @@ static ZJS_DECL_FUNC(zjs_ble_descriptor)
...
 }
 
+#ifdef ZJS_ASHELL
+// INTERRUPT SAFE FUNCTION: No JerryScript VM, allocs, or release prints!
+static void ble_bt_ready(int err)
+{
+    DBG_PRINT("bt_ready() is called [err %d]\n", err);
+    zjs_ble_emit_powered_event();
+}
+#endif
```
grgustaf: I don't like having two copies of this function.

It looks like to do this right, we need to have a "late init" hook, where the bt_enable would be called. We also maybe need to split BLE into a "service" layer and the "API" layer. So the service layer can be there even if the API layer isn't, and the service layer is what takes care of calling bt_enable. One goal would be to remove BUILD_MODULE_BLE from main.c (and anywhere else); all BT code should be in the zjs_bluetooth.c (service) and zjs_ble.c (API) or something.

If you don't want to tackle that now, let's at least remove this function from main.c and put it in ble.h to reuse the same function from main.c.

grgustaf: Yeah, that's what I really meant by that hook - something that gets called after jerry_run() I think.

jimmy-huang:
@grgustaf, updated patch

grgustaf:
+1

PR #1577, opened 2017-10-10
[websockets] Fix crash if no subprotocols were included

jprestwo:
If the node client did not specify any websocket subprotocols
to use, the server would crash if the acceptHandler was
defined in the server init object. This was due to the protocol
array not being initialized if no Sec-WebSocket-Protocol
tag was present in the accept packet.

This initially sets the protocol array to undefined, and if its
still undefined when the acceptHandler is called it initializes
it to an empty array.


grgustaf:
+1

Issue #1581, opened 2017-10-11
[timers] #1542 cause test-callback failed on linux and test-pwm failed on A101


qiaojingx:
### Description

This is an regression issue and casused by #1542. [test-callbacks.js](https://github.com/01org/zephyr.js/blob/master/tests/test-callbacks.js) stuck on Linux and cannot print the last 3 results. 
`set period, pulseWidth, periodCycles and pulseWidthCycles `failed when running [test-pwm.js](https://github.com/01org/zephyr.js/blob/master/tests/test-pwm.js) on A101.

### Test Code

[test-callbacks.js](https://github.com/01org/zephyr.js/blob/master/tests/test-callbacks.js);
[test-pwm.js](https://github.com/01org/zephyr.js/blob/master/tests/test-pwm.js)
 
### Steps to Reproduction

[test-callbacks.js](https://github.com/01org/zephyr.js/blob/master/tests/test-callbacks.js):
1. `make BOARD=linux`
2. `./outdir/linux/release/jslinux tests/test-callbacks.js --noexit`

[test-pwm.js](https://github.com/01org/zephyr.js/blob/master/tests/test-pwm.js):
1. `make JS=tests/test-pwm.js`
2. `make dfu`

### Actual Result

[test-callbacks.js](https://github.com/01org/zephyr.js/blob/master/tests/test-callbacks.js):
![selection_026](https://user-images.githubusercontent.com/19813546/31428750-78982142-ae9e-11e7-8df4-1f2e1ac17d41.png)

[test-pwm.js](https://github.com/01org/zephyr.js/blob/master/tests/test-pwm.js):
![selection_028](https://user-images.githubusercontent.com/19813546/31428758-84310aaa-ae9e-11e7-9922-ddcb651f012c.png)

### Expected Result

No error.
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | a36b2c9 - cd3f44e | Arduino101 | Oct 11, 2017 | Fail
master | a36b2c9 - cd3f44e | Linux | Oct 11, 2017 | Fail
master | 4b77505 | Arduino101 | Oct 11, 2017 | Pass
master | 4b77505 | Linux | Oct 11, 2017 | Pass

qiaojingx:
Verified with ac89871 + #1583 + #1584. This issue is fixed.
Will close it after patch merged.

qiaojingx:
Verified with commit cecd6fe. This issue is fixed. Close

PR #1583, opened 2017-10-11
[linux] Fix linux port to allow for a zero timeout

jprestwo:
The new linux timer port uses signals as callbacks for
timers. The signal API expects a greater than zero timeout
so in the zero timeout case we pass 1.


qiaojingx:
Verified with commit ac89871 + #1583 . test-callbacks.js issue in #1581 is fixed.

grgustaf:
+1

PR #1584, opened 2017-10-11
[tests] Fix PWM test to use actual time values

jprestwo:
The recent timer patch made JS timers much more accurate.
The PWM test was relying on the old slow timers. Changing
the interval values to what they should actually be fixes
the test.


qiaojingx:
Verified with commit ac89871 + #1584 . test-pwm.js issue in #1581 is fixed.

grgustaf:
+1

PR #1587, opened 2017-10-12
Add APIs to pass data over WebUSB connection

grgustaf:


src/zjs_util.h
```diff
@@ -85,7 +85,7 @@ void zjs_pop_mem_stat(void *ptr);
     ({                                             \
         void *zjs_ptr = zjs_malloc_with_retry(sz); \
         if (!zjs_ptr) {                            \
-            ERR_PRINT("malloc failed\n");          \
+            ERR_PRINT("malloc(%d) failed\n", sz);  \
```
jimmy-huang: space after malloc?

src/zjs_webusb.c
```diff
@@ -217,28 +241,343 @@ static ZJS_DECL_FUNC(zjs_webusb_set_url)
...
+    //             bytes to write, length must not be enough to overflow buffer
+    //  returns: true if all bytes read, false if more remain
+    //  effects: consume bytes from UART dev up to length bytes into ring at
+    //             index, until the fifo runs out
+    int total = 0;
+    while (length > 0) {
+        int index = ring->write;
+        char tmp[100];
+        int bytes = uart_fifo_read(dev, tmp, length);
+        memcpy(&ring->buf[index], tmp, bytes);
```
jimmy-huang: can uart_fifo_read() return more than 100 bytes per read? possible overflow here

src/zjs_webusb.c
```diff
@@ -217,28 +241,343 @@ static ZJS_DECL_FUNC(zjs_webusb_set_url)
...
+    tx_ring->write = write;
+
+    if (copied > 0) {
+        struct device *dev = device_get_binding(WEBUSB_SERIAL_PORT_NAME);
+        uart_irq_tx_enable(dev);
+        // FIXME: supposedly this is bad to call uart from outside ISR
+        fill_uart_from_tx(dev);
+    }
+    if (size > copied) {
+        return zjs_error("output buffer full\n");
```
jimmy-huang: no need \n

src/zjs_webusb.c
```diff
@@ -217,28 +241,343 @@ static ZJS_DECL_FUNC(zjs_webusb_set_url)
...
+static void check_uart(struct k_timer *timer)
+{
+    static int count = 0;
+    int rval;
+    u32_t result = 0;
+
+    struct device *dev = device_get_binding(WEBUSB_SERIAL_PORT_NAME);
+    if (!dev) {
+        ERR_PRINT("serial port not found\n");
+        k_timer_stop(timer);
```
jimmy-huang: return here

src/zjs_webusb.c
```diff
@@ -217,28 +241,343 @@ static ZJS_DECL_FUNC(zjs_webusb_set_url)
...
+    webusb_api = zjs_create_object();
+    zjs_make_emitter(webusb_api, prototype, NULL, NULL);
 
     // set the custom and vendor request handlers
     webusb_register_request_handlers(&req_handlers);
 
+    k_timer_init(&uart_timer, check_uart, NULL);
+    k_timer_start(&uart_timer, 0, 100);
+
+    tx_ring = zjs_malloc(sizeof(ringbuf_t));
```
jimmy-huang: check tx_ring and rx_ring on malloc fail

src/zjs_webusb.c
```diff
@@ -217,28 +241,343 @@ static ZJS_DECL_FUNC(zjs_webusb_set_url)
...
+    rx_ring = zjs_malloc(sizeof(ringbuf_t));
+    memset(tx_ring, 0, sizeof(ringbuf_t));
+    memset(rx_ring, 0, sizeof(ringbuf_t));
+
     return jerry_acquire_value(webusb_api);
 }
 
 void zjs_webusb_cleanup()
 {
+    zjs_free(tx_ring);
```
jimmy-huang: add check if is NULL?

src/zjs_util.h
grgustaf: That's intended to show the function call, malloc(64) for 64 bytes.

src/zjs_webusb.c
grgustaf: Good catch, i think this was a partial attempt at an improvement... I don't really want to read in two chunks, so it would be better to read the whole thing and then write to end of ring buf, then wrap to beginning at once.

src/zjs_webusb.c
grgustaf: free(NULL) is safe

src/zjs_webusb.c
jimmy-huang: Ok, add a FIXME here so we don't lose track of this

jimmy-huang:
lgtm, +1

PR #1589, opened 2017-10-13
[travis] Add Mac OSX build to travis CI

jimmy-huang:


grgustaf:
+1, very nice - it doesn't seem to block the green checkmark from coming up so let's do it.

PR #1590, opened 2017-10-13
[buffer] Add utf8 encoding for toString (default as in Node)

grgustaf:


jimmy-huang:
+1

PR #1591, opened 2017-10-13
Buffer add copy

grgustaf:


grgustaf:
This includes #1590 so let's make sure that one gets merged first and this rebases properly.

jimmy-huang:
+1

PR #1597, opened 2017-10-16
Added AIO for K64f

jimmy-huang:
Make AIO a event emitter and replace the existing .on() implementation.

Add support for Analog A0-A3 pins on the FRDM-K64F board.


src/zjs_aio_a101.c
```diff
@@ -142,13 +142,15 @@ static void ipm_msg_receive_callback(void *context, u32_t id,
 #ifdef DEBUG_BUILD
         u32_t pin = msg->data.aio.pin;
 #endif
+        ZVAL num = jerry_create_number(pin_value);
 
         switch (msg->type) {
         case TYPE_AIO_PIN_READ:
+            DBG_PRINT("aio async read %d\n", pin_value);
+            break;
```
grgustaf: Read is going to do nothing now?

src/zjs_aio_k64f.c
```diff
@@ -0,0 +1,331 @@
+// Copyright (c) 2017, Intel Corporation.
+
```
grgustaf: Shouldn't this work for all single-core boards, not just k64f? Let's just name this aio maybe and the other one a101 or ipm?

src/zjs_aio_k64f.c
```diff
@@ -0,0 +1,331 @@
...
+    for (int i = 0; i < AIO_LEN; i++) {
+        if (pin_enabled[i]) {
+            // FIXME: We dont know when a object subscribes to onchange
+            // events, need a way to intercept the .on()
+            // currently, if the pin is opened, then it do a read
+            // and signal the event, this could slow down the system
+            // when there are multiple AIO pins opened.
+            if (uptime % (u32_t)(CONFIG_SYS_CLOCK_TICKS_PER_SEC /
+                AIO_POLL_FREQUENCY * 10) == 0) {
```
grgustaf: Does this get run every millisecond? What if we don't happen to run right at the right uptime? Seems like instead you should save the last update time in a static variable or something and see if we're beyond the next poll time now?

src/zjs_aio_k64f.c
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    zjs_loop_unblock();
+    return K_FOREVER;
+}
+
+jerry_value_t zjs_aio_init()
+{
+    // FIXME: need a way to unregister service routines
```
grgustaf: True, Brian should care about this for ashell; could you open an issue?

src/zjs_k64f_pins.c
```diff
@@ -81,14 +81,12 @@ jerry_value_t zjs_k64f_init()
...
-    zjs_obj_add_number(obj, "A2", PTB + 10);
-    zjs_obj_add_number(obj, "A3", PTB + 11);
-    zjs_obj_add_number(obj, "A4", PTC + 11);
-    zjs_obj_add_number(obj, "A5", PTC + 10);
+    // TODO: It appears that currently Zephyr only enables
+    // PTB2, PTB3, PTB10, PTB11, which are A0-A3 only, A4-A5 are not supported.
+    zjs_obj_add_number(obj, "A0", 12);
+    zjs_obj_add_number(obj, "A1", 13);
+    zjs_obj_add_number(obj, "A2", 14);
+    zjs_obj_add_number(obj, "A3", 15);
```
grgustaf: Where do these 12-15 numbers come from then?

src/zjs_aio_a101.c
jimmy-huang: This type of read msg from ARC is no longer needed for async reads, for simplicity of the API, both read() and readAsync() does a synchronous read over IPM, and asyncRead just call signal callback after it gets the result, so this type of message should never show up.

src/zjs_aio_k64f.c
jimmy-huang: i could rename them

src/zjs_aio_k64f.c
jimmy-huang: I'll open an issue

src/zjs_k64f_pins.c
jimmy-huang: these come from https://os.mbed.com/media/uploads/GregC/frdm-k64f_pinout_reve3.xls
in the J4 section, where pin A0-A4 (pins 55-59), is mapped to ADC channel 12-15 in Zephyr

jimmy-huang:
@grgustaf, updated patch

grgustaf:
+1

PR #1601, opened 2017-10-18
[ashell] Allow setting IDE_GPIO_PIN to enter IDE mode

jimmy-huang:
If you building the IDE by passing IDE_GPIO_PIN to configure
the board to only boot into IDE mode only if a button press is detected

For example:
make ide IDE_GPIO_PIN=8

This will build the IDE so that if a button is connected to IO8
on the board using GPIO and it's holded down when the Arduino 101
is booting, it will then boot into ASHELL/IDE mode, otherwise
it will look for the boot.cfg file and boots the file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,14 @@ endif
 ifeq ($(shell test $(ROM) -gt 296; echo $$?), 0)
 $(error ROM must be no higher than 296)
 endif
+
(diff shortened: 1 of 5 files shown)


jimmy-huang:
@grgustaf, updated patch

grgustaf:
+1

PR #1604, opened 2017-10-19
[util] Added Math.random() api

jimmy-huang:

### DIFF
--- a/docs/mathstubs.md
+++ b/docs/mathstubs.md
@@ -1,0 +1,37 @@
+ZJS API for MathStubs
+========================
+
+* [Introduction](#introduction)
+* [Web IDL](#web-idl)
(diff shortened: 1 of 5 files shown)


grgustaf:
Does this conflict with pulling in the full math library by editing fragments/jerry_feature.base?

If so, we might want to either solve that or name this something else.

jimmy-huang:
Yeah, it will conflict, i think a better way would be to detect if the jerry math feature is enabled or not before including this module.  I can probably look into it or you want this landed soon, then I can just rename it.

jimmy-huang:
@grgustaf changed to MathStubs to avoid conflict

jimmy-huang:
@grgustaf updated patch

Issue #1582, opened 2017-10-11
ringbuf put errors when running I2CBMP280.js and HeartRateDemo.js on A101


qiaojingx:
### Description

This is an regression issue and probably caused by #1542.` ringbuf put errors` prints when running [HeartRateDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/HeartRateDemo.js) and [I2CBMP280.js](https://github.com/01org/zephyr.js/blob/master/samples/I2CBMP280.js) on A101.

### Test Code

[HeartRateDemo.js](https://github.com/01org/zephyr.js/blob/master/samples/HeartRateDemo.js);
[I2CBMP280.js](https://github.com/01org/zephyr.js/blob/master/samples/I2CBMP280.js)
 
### Steps to Reproduction

1. `make JS=samples/I2CBMP280.js`
2. `make dfu`

### Actual Result

![selection_029](https://user-images.githubusercontent.com/19813546/31429549-ef2b20d2-aea0-11e7-82c9-540a12ab0e0d.png)

### Expected Result

No error.
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | a36b2c9 - cd3f44e | Arduino101 | Oct 11, 2017 | Fail
master | 4b77505 | Arduino101 | Oct 11, 2017 | Pass

brianjjones:
Looks like the setInterval isn't finishing before it gets called again.  I'll look to make sure there isn't a problem with callbacks.  It may also be that 10 is too low a timer number.

brianjjones:
Looks like callbacks are fine, the demo timer is just going faster than the callbacks can be handled - hence the error message.  #1614 will fix this.

qiaojingx:
Verified with commit cac4c3a + #1614. Error in I2CBMP280.js has been fixed. Could you change interval time in HeartRateDemo.js, too?

grgustaf:
This seems to indicate we have a performance regression because we used to do this work in 10ms apparently and now it takes 200ms? Rather than mask the problem (:D) let's look into why. @brianjjones, could you try testing this just before and after #1542 to see if that's what caused the problem.

grgustaf:
@brianjjones ping on this one, just note I don't think we're done here.

jimmy-huang:
@cuiyanx is this still reproducible?

cuiyanx:
@jimmy-huang 
[...]

PR #1614, opened 2017-10-25
[samples] Making the timer longer for I2CBMP280.js

brianjjones:
The current timer is too fast, and it gets called multiple times
before the first call gets serviced. Increasing the time allows
the CB to get serviced befor the next interval hits.

### DIFF
--- a/samples/I2CBMP280.js
+++ b/samples/I2CBMP280.js
@@ -130,4 +130,4 @@ glcdInit();
 
 /* Loop read forever */
 
-var timer = setInterval(readTemperature, 10);
+var timer = setInterval(readTemperature, 200);


brianjjones:
This is a fix for #1582

grgustaf:
This is fine in itself but I fear it's just masking the real problem in #1582, so maybe we should root cause that more.

jimmy-huang:
yeah, we should be able to hit 10ms with service callbacks, have you tried increasing the service callback count?

brianjjones:
@jimmy-huang FYI yes this is still an issue, I see the following output on master.

554:(ERROR) 101 ringbuf put errors (last rval=-122)

554:(ERROR) 264 ringbuf put errors (last rval=-122)

554:(ERROR) 660 ringbuf put errors (last rval=-122)

554:(ERROR) 1452 ringbuf put errors (last rval=-122)

Issue #1561, opened 2017-09-26
[ashell] command boot not working properly


qiaojingx:
### Description

1. After set [BMI160Accelerometer.js](https://github.com/01org/zephyr.js/blob/master/samples/BMI160Accelerometer.js) as boot file in ashell, then reboot device, `ringbuf put errors` will print at the beginning. And sensor start delay.

2. Not all the samples can start automatically after be set by boot. e.g. [BLE.js](https://github.com/01org/zephyr.js/blob/master/samples/BLE.js)

### Test Code

[Ashell](https://github.com/01org/zephyr.js/blob/master/docs/ashell.md);
 
### Steps to Reproduction

1. `make ashell ROM=256`
2. `make dfu`
3. load the code of samples to ashell with name as test.js
4. `acm>boot test.js`
5. `acm>reboot`

### Actual Result

1. ringbuf put errors print first if boot BMI160Accelerometer.js in ashell
![selection_017](https://user-images.githubusercontent.com/19813546/30838501-6ecd306e-a29e-11e7-8945-5d1583c41ccf.png)

2. BLE service not start if boot BLE.js in ashell although can see `BLE sample...` in `/dev/ttyUSB0`.
![selection_018](https://user-images.githubusercontent.com/19813546/30838929-1ba60642-a2a1-11e7-9d1e-164eff295497.png)

### Expected Result

No error and boot file can start automatically after reboot.

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | cbe7e8f | Arduino 101 | Sep 26, 2017 | Fail
master | b11dc55 | Arduino 101 | Sep 26, 2017 | Fail

xiaoyu3x:
This issue is reproduced when using commit [3ca5e23](https://github.com/01org/zephyrjs-ide/commit/3ca5e23effd1ef9dddcfa49a963c660fd225cf0a)

brianjjones:
This is fixed by #1615

qiaojingx:
Verified with commit e19ed4f. This issue is fixed.

PR #1615, opened 2017-10-25
[ashell] Fix for ashell boot

brianjjones:
Currently ashell boot doesn't work with javascript that has callbacks
or timers.  This is because once the boot.cfg JS gets called, it
goes into an infinite loop waiting for uart. Until it gets the
connection, it won't return.  This means the main loop doesn't get
control back, meaning callbacks are not processed.  This change
breaks the code up a bit so that it checks if the connection is ready
when uart_process gets called, if not try to connect.  If the
connection fails, just return and go through the main loop again.
Continue until a connection is made.

### DIFF
--- a/src/ashell/term-uart.c
+++ b/src/ashell/term-uart.c
@@ -90,6 +90,7 @@ u32_t free_count = 0;
 static struct uart_input *isr_data = NULL;
 static u32_t tail = 0;
 static char *buf;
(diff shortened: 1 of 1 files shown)


brianjjones:
This is a fix for #1561

grgustaf:
+1

PR #1632, opened 2017-11-03
[trlite] Add soft fail when Linux tests don't reach completion

grgustaf:
We should turn this into a hard fail as soon as we get everything
working again: callbacks, event, gpio, promise, and timers tests
were all broken by PR #1542.


jimmy-huang:
+1

PR #1633, opened 2017-11-03
[build] Disable LTO for JerryScript to prevent ar warnings

grgustaf:
Not sure if these are going to cause a problem, but apparently Zephyr
doesn't use LTO now so we're better off making things match I expect.


jimmy-huang:
+1

PR #1635, opened 2017-11-03
[modules] Adding zjs_unregister_service_routine function

brianjjones:

### DIFF
--- a/src/zjs_modules.c
+++ b/src/zjs_modules.c
@@ -235,6 +235,18 @@ void zjs_register_service_routine(void *handle, zjs_service_routine func)
     return;
 }
 
+void zjs_unregister_service_routine(zjs_service_routine func) {
+    for (int i = 0; i < num_routines; i++) {
+        if (svc_routine_map[i].func == func) {
+            svc_routine_map[i].handle = NULL;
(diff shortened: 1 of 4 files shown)


jimmy-huang:
+1

src/zjs_modules.c
```diff
@@ -235,6 +235,18 @@ void zjs_register_service_routine(void *handle, zjs_service_routine func)
     return;
 }
 
+void zjs_unregister_service_routine(zjs_service_routine func) {
+    for (int i = 0; i < num_routines; i++) {
+        if (svc_routine_map[i].func == func) {
+            svc_routine_map[i].handle = NULL;
+            zjs_free(&svc_routine_map[i].func);
```
jprestwo: This should not be freed. In the register API above the function is just being set not allocated.

Another thing that may bite you in the future is if you plan on adding/removing service routines during normal runtime. With this removal you could overwrite an existing service routine by just decrementing `num_routines`.

brianjjones: Good point. I'll also switch the services to be a linked list rather than array so that removing one doesn't possibly break something.

PR #1636, opened 2017-11-03
[gfx] Changes to make GFX match the new Jerry module model.

brianjjones:


jimmy-huang:
+1

PR #1642, opened 2017-11-07
[webusb] Fixing Events and WebUSB after regression

brianjjones:
Events and WebUSB were broken after the Jerry modules change,
this adds the code needed to make them work again.


src/zjs_event.c
```diff
@@ -478,9 +467,8 @@ static void zjs_event_create_prototype() {
 void zjs_make_emitter(jerry_value_t obj, jerry_value_t prototype,
                       void *user_data, zjs_event_free free_cb)
 {
-    if (!zjs_event_initialized) {
-        zjs_event_create_prototype();
-    }
+    zjs_event_create_prototype();
```
grgustaf: Since the function name is "create", it should create, not "create unless something" - so this was better before with the if here I think.

src/zjs_event.c
```diff
@@ -443,17 +443,6 @@ void *zjs_event_get_user_handle(jerry_value_t obj)
     return NULL;
 }
 
-static void zjs_event_cleanup(void *native)
-{
-    jerry_release_value(zjs_event_emitter_prototype);
-    zjs_event_initialized = false;
```
grgustaf: Instead of this global boolean set the _prototype = 0 initially; when it's set it's its own "existence boolean". And then when it gets freed you'll have to set that global back to 0. (This will mean you'll have to catch the prototype object getting freed, I guess by setting a different native pointer on it.)

src/zjs_event.c
```diff
@@ -443,17 +443,6 @@ void *zjs_event_get_user_handle(jerry_value_t obj)
     return NULL;
 }
 
-static void zjs_event_cleanup(void *native)
-{
-    jerry_release_value(zjs_event_emitter_prototype);
```
grgustaf: The prototype never gets freed now - so as discussed in IRC you should acquire prototype in make_emitter and release it once each time an emitter is freed. Be careful that when you *create* the prototype, which adds one reference, that first emitter doesn't end up with two references or it will never get freed. That will be a little ugly.

src/zjs_webusb.c
```diff
@@ -540,7 +539,22 @@ static void check_uart(struct k_timer *timer)
...
+    jerry_release_value(webusb_api);
+    webusb_api = 0;
+    zjs_free(webusb_origin_url);
+    webusb_origin_url = NULL;
+}
+/*
+static const jerry_object_native_info_t webusb_module_type_info = {
+   .free_cb = zjs_webusb_cleanup
+};
+*/
```
grgustaf: Why move cleanup function up here? I at least am expecting to find it at the bottom.
Also, why the commented out type_info?

brianjjones: Leftover code from a previous attempt to fix this,  webusb_module_type_info was used in init and had to be first. But since its an emitter it follows a different pattern, I'll move it back and remove the un-needed code.

src/zjs_event.c
brianjjones: zjs_event_create_prototype already checks if its initialized, and just returns if it is already done.  No reason to check twice. I can name it something else if you'd prefer, but I think the check should live in the prototype function, so that if more code is added that calls it, we don't accidentally create another copy.

grgustaf: I'm suggesting that you move the check back out where it was before. I doubt anyone else is going to call that. It doesn't make sense to call a function "create prototype" if you don't need to create the prototype and it's not actually going to create the prototype. It's wise to make the behavior match the English. :)

grgustaf: Or if you insist, then a better name for the function would be ensure_prototype_exists().

src/zjs_event.c
brianjjones: The latest commit contains code to free the prototype, give it a look.

src/zjs_event.c
brianjjones: Fixed.

grgustaf:
+1

PR #1643, opened 2017-11-07
[net] Jerry module fix for net

brianjjones:
Jerry modules don't support '-' in the name.  Renaming net-config
to net_config.

### DIFF
--- a/src/zjs_net_config.json
+++ b/src/zjs_net_config.json
@@ -1,6 +1,6 @@
 {
-    "module": "net-config",
-    "require": "net-config",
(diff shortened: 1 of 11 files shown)


src/zjs_net_config.json
```diff
@@ -1,6 +1,6 @@
 {
-    "module": "net-config",
-    "require": "net-config",
+    "module": "net_config",
```
jimmy-huang: I would recommend to change this into NetConfig for Javascript.

grgustaf:
+1

PR #1654, opened 2017-11-13
[build] Allow outdir to be specified with O=

grgustaf:
If not specified, 'outdir' remains the default.


jimmy-huang:
+1

PR #1655, opened 2017-11-14
[build] Move generated config files to outdir

grgustaf:


grgustaf:
I'll take another stab at this if needed after Jimmy gets us ported to cmake.

PR #1658, opened 2017-11-16
[modules] Adding JerryScript module support for JS modules

brianjjones:

### DIFF
--- a/src/ashell/file-utils.c
+++ b/src/ashell/file-utils.c
@@ -170,3 +170,38 @@ int fs_get_boot_cfg_filename(const char *timestamp, char *filename)
     fs_close_alloc(file);
     return -1;
 }
+
+char *read_file_alloc(const char *file_name, ssize_t *size)
+{
+    char *file_buf = NULL;
+    fs_file_t *fp = fs_open_alloc(file_name, "r");
+
+    if (fp == NULL)
+        return 0;
+
+    *size = fs_size(fp);
+    if (*size == 0) {
+        ERR_PRINT("[ERR] Empty file (%s)\n", file_name);
+        fs_close_alloc(fp);
+        return NULL;
+    }
+
+    file_buf = (char *)zjs_malloc(*size);
+    if (file_buf == NULL) {
+        ERR_PRINT("[ERR] Not enough memory for (%s)\n", file_name);
+        fs_close_alloc(fp);
+        return NULL;
+    }
+
+    ssize_t brw = fs_read(fp, file_buf, *size);
+
+    if (brw != *size) {
+        ERR_PRINT("[ERR] Not enough memory for (%s)\n", file_name);
+        fs_close_alloc(fp);
+        zjs_free(file_buf);
+        return NULL;
+    }
+
+    fs_close_alloc(fp);
+    return file_buf;
+}
--- a/src/ashell/file-utils.c
+++ b/src/ashell/file-utils.c
@@ -27,6 +27,8 @@
 #include "file-utils.h"
 #include "../zjs_util.h"
 
+#define FILE_ERR ERR_PRINT("Read file failed\n")
+
 int fs_exist(const char *path)
 {
     int res;
@@ -177,18 +179,20 @@ char *read_file_alloc(const char *file_name, ssize_t *size)
     fs_file_t *fp = fs_open_alloc(file_name, "r");
 
     if (fp == NULL)
-        return 0;
+        return NULL;
 
     *size = fs_size(fp);
     if (*size == 0) {
-        ERR_PRINT("[ERR] Empty file (%s)\n", file_name);
+        DBG_PRINT("Empty file (%s)\n", file_name);
+        FILE_ERR;
         fs_close_alloc(fp);
         return NULL;
     }
 
     file_buf = (char *)zjs_malloc(*size);
     if (file_buf == NULL) {
-        ERR_PRINT("[ERR] Not enough memory for (%s)\n", file_name);
+        DBG_PRINT("Not enough memory for (%s)\n", file_name);
+        FILE_ERR;
         fs_close_alloc(fp);
         return NULL;
     }
@@ -196,7 +200,8 @@ char *read_file_alloc(const char *file_name, ssize_t *size)
     ssize_t brw = fs_read(fp, file_buf, *size);
 
     if (brw != *size) {
-        ERR_PRINT("[ERR] Not enough memory for (%s)\n", file_name);
+        DBG_PRINT("Read failed for (%s)\n", file_name);
+        FILE_ERR;
         fs_close_alloc(fp);
         zjs_free(file_buf);
         return NULL;
--- a/src/ashell/jerry-code.c
+++ b/src/ashell/jerry-code.c
@@ -32,74 +32,12 @@
 
 static jerry_value_t parsed_code = 0;
 
-static void javascript_print_error(jerry_value_t error_value)
-{
-    if (!jerry_value_has_error_flag(error_value))
-        return;
-
-    jerry_value_clear_error_flag(&error_value);
-    ZVAL err_str_val = jerry_value_to_string(error_value);
-
-    jerry_size_t size = 0;
-    char *msg = zjs_alloc_from_jstring(err_str_val, &size);
-    const char *err_str = msg;
-    if (!msg) {
-        err_str = "[Error message too long]";
-    }
-
-    jerry_port_log(JERRY_LOG_LEVEL_ERROR, "%s\n", err_str);
-    zjs_free(msg);
-}
-
-static void javascript_print_value(const jerry_value_t value)
-{
-    if (jerry_value_is_undefined(value)) {
-        jerry_port_log(JERRY_LOG_LEVEL_TRACE, "undefined");
-    } else if (jerry_value_is_null(value)) {
-        jerry_port_log(JERRY_LOG_LEVEL_TRACE, "null");
-    } else if (jerry_value_is_boolean(value)) {
-        if (jerry_get_boolean_value(value)) {
-            jerry_port_log(JERRY_LOG_LEVEL_TRACE, "true");
-        } else {
-            jerry_port_log(JERRY_LOG_LEVEL_TRACE, "false");
-        }
-    }
-    /* Float value */
-    else if (jerry_value_is_number(value)) {
-        double val = jerry_get_number_value(value);
-        // %lf prints an empty value :?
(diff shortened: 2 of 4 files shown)


src/ashell/file-utils.c
```diff
@@ -170,3 +170,38 @@ int fs_get_boot_cfg_filename(const char *timestamp, char *filename)
...
     return -1;
 }
+
+char *read_file_alloc(const char *file_name, ssize_t *size)
+{
+    char *file_buf = NULL;
+    fs_file_t *fp = fs_open_alloc(file_name, "r");
+
+    if (fp == NULL)
+        return 0;
```
grgustaf: return NULL for consistency

src/ashell/file-utils.c
```diff
@@ -170,3 +170,38 @@ int fs_get_boot_cfg_filename(const char *timestamp, char *filename)
...
+
+    *size = fs_size(fp);
+    if (*size == 0) {
+        ERR_PRINT("[ERR] Empty file (%s)\n", file_name);
```
grgustaf: Drop the [ERR] from all ERR_PRINT calls, ERR_PRINT already puts (ERROR) in the output.

However, we want to be careful about use of ERR_PRINT since these strings all take space in our image. The string "out of memory" is used elsewhere so please use that exact string for any out of memory cases... then there will be only one copy in the image. (Bonus points: we should #define that string as a macro somewhere for spelling consistency and use the macro everywhere.) We don't really need more detail because ERR_PRINT will tell them the filename and line number where it occurred.

If you think more detail is warranted, like the filename here, you can put that in DBG_PRINT.

src/ashell/file-utils.c
```diff
@@ -170,3 +170,38 @@ int fs_get_boot_cfg_filename(const char *timestamp, char *filename)
...
+    if (file_buf == NULL) {
+        ERR_PRINT("[ERR] Not enough memory for (%s)\n", file_name);
+        fs_close_alloc(fp);
+        return NULL;
+    }
+
+    ssize_t brw = fs_read(fp, file_buf, *size);
+
+    if (brw != *size) {
+        ERR_PRINT("[ERR] Not enough memory for (%s)\n", file_name);
```
grgustaf: This condition doesn't mean out of memory, it means the read didn't find enough bytes, or there was a read error (negative return value). The former shouldn't ever occur since you just got the file size. Still not bad to check but don't say out of memory. Maybe "read failed". Should be very uncommon so might make sense to put in DBG_PRINT only, but otoh kind of important if it did happen, so I'm of two minds about it.

src/ashell/jerry-code.c
```diff
@@ -32,69 +32,23 @@
...
-        if (!cur->reqs_checked) {
-            return cur;
-        } else {
-            cur = cur->next;
-        }
+    jerry_size_t size = 0;
+    char *msg = zjs_alloc_from_jstring(err_str_val, &size);
+    const char *err_str = msg;
+    if (!msg) {
+        err_str = "[Error message too long]";
```
grgustaf: This code is duplicate of zjs_util.c's `zjs_print_error_message`; however the string above has [] instead of () so that will be an extra string stored, better to use the same thing for just one copy.

But also we ought to make this common code somewhere and reuse it instead of duplicating.

src/ashell/jerry-code.c
```diff
@@ -414,17 +139,14 @@ int javascript_parse_code(const char *file_name)
...
-                javascript_print_error(parsed_code);
-                jerry_release_value(parsed_code);
-            } else {
-                ret = 0;
-            }
+        /* Setup Global scope code */
+        parsed_code = jerry_parse((const jerry_char_t *)buf, size, false);
+        if (jerry_value_has_error_flag(parsed_code)) {
+            DBG_PRINT("Error parsing JS\n");
+            zjs_print_error_message(parsed_code, ZJS_UNDEFINED);
```
grgustaf: Or if you're able to use this function directly, maybe you can replace the calls to `javascript_print_error` with this? I haven't looked in detail.

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+// Eval the JavaScript, and return the module.
+static bool javascript_eval_code(const char *source_buffer, ssize_t size, jerry_value_t *ret_val)
 {
-    &jerryx_module_native_resolver
-};
-
-static ZJS_DECL_FUNC(native_require_handler)
+    (*ret_val) = jerry_eval((jerry_char_t *)source_buffer, size, false);
+    if (jerry_value_has_error_flag(*ret_val)) {
+        ERR_PRINT("[ERR] failed to evaluate JS\n");
```
grgustaf: drop the [ERR] here too

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+        return false;
+    }
+    return true;
+}
+#endif
+// Find the module on the filestystem
```
grgustaf: Add newline before comment for readability

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+static bool load_js_module_fs (const jerry_value_t module_name, jerry_value_t *result)
```
grgustaf: Remove space before ( for consistency.

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
-    // args: module name
-    ZJS_VALIDATE_ARGS(Z_STRING);
-
-    jerry_size_t size = MAX_MODULE_STR_LEN;
-    char module[size];
-    zjs_copy_jstring(argv[0], module, &size);
-    if (!size) {
-        return RANGE_ERROR("argument too long");
+    // Currently searching the filesystem is only supported on ashell
+    #ifdef ZJS_ASHELL
```
grgustaf: Preprocessor commands should be at far left column, i.e. remove the four spaces here and below on the #else and #endif.

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
-        return native_module;
-    }
-    DBG_PRINT("Native module not found, searching for JavaScript module %s\n",
-              module);
-#ifdef ZJS_LINUX_BUILD
+    // Eval the code from the file.
+    bool ret = javascript_eval_code(req_file_buffer, file_size, result);
+    zjs_free(req_file_buffer);
+    return ret;
+    #else
```
grgustaf: no indent

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+    return false;
+    #endif // ZJS_ASHELL
```
grgustaf: no indent

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+#else // ZJS_LINUX_BUILD
+/****************************************
+*   Linux JavaScript module resolver
+*****************************************/
+// Find the module on the filestystem
+static bool load_js_module_fs (const jerry_value_t module_name, jerry_value_t *result)
+{
     // Linux can pass in the script at runtime, so we have to read in/parse any
     // JS modules now rather than at compile time
+    #ifdef ZJS_SNAPSHOT_BUILD
```
grgustaf: no indent

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+    DBG_PRINT("Parser disabled, can't check FS for module\n");
+    return false;
+    #endif
```
grgustaf: no indent

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+    return false;
+    #endif // ZJS_ASHELL
+
+}
+#else // ZJS_LINUX_BUILD
+/****************************************
+*   Linux JavaScript module resolver
+*****************************************/
+// Find the module on the filestystem
+static bool load_js_module_fs (const jerry_value_t module_name, jerry_value_t *result)
```
grgustaf: no space

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
     }
 
     zjs_free_script(str);
-#endif
+    return ret;
+}
+#endif // !ZJS_LINUX_BUILD
+
+// Try to find the module in the main JS file
+static bool load_js_module_obj (const jerry_value_t module_name, jerry_value_t *result)
```
grgustaf: no space before (

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+{
+    jerry_size_t module_size = jerry_get_utf8_string_size (module_name) + 1;
```
grgustaf: no space before (

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+    &load_fs_resolver                   // Check for a JS module on the FS
+};
+
+// native require handler
+static ZJS_DECL_FUNC(native_require_handler)
+{
+    ZJS_VALIDATE_ARGS(Z_STRING);
+
+    // Get the module name
+    jerry_size_t module_size = jerry_get_utf8_string_size (argv[0]) + 1;
```
grgustaf: no space

src/zjs_modules.c
```diff
@@ -33,87 +36,169 @@ struct routine_map {
...
+    char module[module_size];
+    zjs_copy_jstring(argv[0], module, &module_size);
 
-    DBG_PRINT("JavaScript module %s loaded\n", module);
-    return jerry_acquire_value(found_obj);
+    // Try each of the resolvers to see if we can find the requested module
+    jerry_value_t result = jerryx_module_resolve (argv[0], resolvers, 3);
```
grgustaf: no space

src/ashell/file-utils.c
brianjjones: Created a macro here for the error and simplified it to just "Read file failed".  I've added DBG_PRINT as well in case someone needs to know more, it can be handy to know which file is failing and why.

src/ashell/file-utils.c
brianjjones: Created a macro here for the error and simplified it to just "Read file failed". I've added DBG_PRINT as well in case someone needs to know more, it can be handy to know which file is failing and why.

src/ashell/jerry-code.c
brianjjones: These javascript_print functions are old and I don't think we need them anymore.  I'll remove them and use the zjs_print_error_message for our needs.

grgustaf:
+1

PR #1659, opened 2017-11-16
[board] Fix support for boards without friendly pin mappings in ZJS

grgustaf:
Tested with the Nucleo F411RE.


jimmy-huang:
+1

PR #1665, opened 2017-11-17
[build] Build Zephyr with CMake

jimmy-huang:


Makefile
```diff
@@ -447,27 +419,26 @@ ARC_RESTRICT="zjs_ipm_arc.json,\
 
 # Builds ARC binary
 .PHONY: arc
+# Restrict ARC build to only certain "arc specific" modules
 arc: analyze
-	# Restrict ARC build to only certain "arc specific" modules
```
grgustaf: maybe you can restore the position with the @ trick?

Makefile
```diff
@@ -480,7 +451,8 @@ endif
 # Run debug server over JTAG
 .PHONY: adebug
 adebug:
-	make -f Makefile.zephyr BOARD=arduino_101 ARCH=x86 O=$(OUT)/arduino_101 debugserver
+	# fixme
+	# make -f Makefile.zephyr BOARD=arduino_101 ARCH=x86 O=$(OUT)/arduino_101 debugserver
```
grgustaf: maybe fail with an informative error instead?

README.md
```diff
@@ -62,6 +62,21 @@ sudo apt-get update
 sudo apt-get install cmake dfu-util git python3-yaml screen uglifyjs
 ```
 
+Note: cmake 3.8.2 or later is required, and your system might install an older
+version, if this is the case, you'll have to manually install the latest
+version of cmake:
+```bash
+sudo apt-get purge cmake
+mkdir .tmp
+cd .tmp
```
grgustaf: how about drop the .tmp stuff since it already creates a cmake subdir anyway

README.md
```diff
@@ -93,7 +108,8 @@ file in the root of the source tree to see what version you're on. If you're on
 the master development branch it will just say 'devel'.
 
 #### Install the Zephyr SDK
-Download the [latest Zephyr SDK](https://www.zephyrproject.org/downloads), then:
+Download the [latest Zephyr SDK](https://www.zephyrproject.org/downloads) (0.9.2
+or later), then:
```
grgustaf: don't really need this change... the "latest" is already "0.9.2 or later" :)

arc/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2017, Intel Corporation.
+
+include($ENV{ZEPHYR_BASE}/cmake/app/boilerplate.cmake NO_POLICY_SCOPE)
+project(NONE)
+
+if(NOT DEFINED ENV{ZEPHYR_BASE})
```
grgustaf: Again, already used this above.

arc/src/zjs_i2c_arc.json
```diff
@@ -8,6 +8,9 @@
             "CONFIG_I2C=y"
         ]
     },
-    "src": ["arc_i2c.c", "../../src/zjs_ipm.c", "../../src/zjs_i2c_handler.c"],
+    "src": [
+        "src/arc_i2c.c",
+        "../../src/zjs_ipm.c",
+        "../../src/zjs_i2c_handler.c"],
```
grgustaf: Are you sure this shouldn't be just one ../ ? Above it seemed like you're building from one dir higher now?

cmake/zjs.cmake
```diff
@@ -0,0 +1,54 @@
...
+if(PRINT_FLOAT)
+  add_definitions(-DPRINT_FLOAT=${PRINT_FLOAT})
+endif()
+
+if(VARIANT)
+  add_definitions(-DVARIANT=${VARIANT})
+endif()
+
+if(ZJS_FLAGS)
+  add_definitions(${ZJS_FLAGS})
```
grgustaf: Does this work if there are multiple flags?

scripts/analyze
```diff
@@ -305,19 +305,22 @@ def write_jrsconfig(list):
...
 
-def write_makefile(src, zconf, file):
-    """Write out a src Makefile"""
-    with open(args['MAKEBASE'], "r") as b:
+def write_cmakefile(src, zconf, file):
+    """Write out a src CMake file"""
+    with open(args['CMAKEBASE'], "r") as b:
         base = b.read()
     with open(file, "w") as f:
-        f.write(base)
```
grgustaf: This doesn't seem right, you're no longer using base variable you just made at all.

scripts/analyze
```diff
@@ -305,19 +305,22 @@ def write_jrsconfig(list):
...
         if src:
-            f.write("obj-y += \\\n")
             for i in src:
-                f.write("\t" + i.replace(".c", ".o") + " \\\n")
-            f.write("\nccflags-y += \\\n")
+                f.write("target_sources(app PRIVATE " + i + ")\n")
+            f.write("\n")
             for i in zconf:
-                f.write("\t" + i + " \\\n")
+                """fixme: include paths should be a separate config"""
```
grgustaf: Do you mean to fix this before merge?

src/zjs_qemu_x86.json
```diff
@@ -4,8 +4,8 @@
     "zephyr_conf": {
         "qemu_x86": [
             "CONFIG_MAIN_STACK_SIZE=2048",
-            "CONFIG_RAM_SIZE=80",
-            "CONFIG_ROM_SIZE=256"
+            "CONFIG_RAM_SIZE=4096",
+            "CONFIG_ROM_SIZE=4092"
```
grgustaf: 4092 seems like an odd choice. :)

I don't know if it makes sense to go so far above real board specs though? Maybe 512 / 1024 like a higher end board?

jimmy-huang: this is what they used in the latest Zephyr qemu sample, I found it weird too they use 4092, but it's what's working for them, so i just use that as is.

arc/src/zjs_i2c_arc.json
jimmy-huang: actually, i don't get any compile errors if i remove them, seems they are not neccessary.

scripts/analyze
jimmy-huang: no, it's intended to be a todo in another patch, i'll file a issue to keep track of it

nagineni:
+1 . Checked with cmake and no build errors noticed.

grgustaf:
+1

PR #1666, opened 2017-11-18
[build] Fixed Linux outdir include path on MacOS

jimmy-huang:
This fixes build fails on MacOS Linux builds


grgustaf:
+1

PR #1676, opened 2017-11-22
[gpio] Check for bad gpio device

grgustaf:
With the full pin naming method, the first part is used as a device name
so we should check that it's actually a legitimate device name.


jimmy-huang:
+1

PR #1677, opened 2017-11-22
[gpio] Expand GPIO tests for k64f/f411re to cover all external pins

grgustaf:
For F411RE, this demonstrates how there are a total of 41 external pins
that actually work as GPIOs. For K64F, basically none of the extra pins
actually work w/ the default Zephyr configuration, but may in the
future if we update pinmux settings.


jimmy-huang:
+1

PR #1688, opened 2017-12-05
[build] Update Makefile to support olimex_stm32_e407 and 96b_carbon

cuiyanx:
Update Makefile to support olimex_stm32_e407 and 96b_carbon boards using dfu mode.

a101: `make dfu` or `make dfu BOARD=arduino_101`
olimex_stm32_e407: `make dfu BOARD=olimex_stm32_e407`
96b_carbon: `make dfu BOARD=96b_carbon`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,11 +222,19 @@ ide: zephyr
 .PHONY: ashell
 ashell: zephyr
 
-# Flash Arduino 101 x86 and arc images
+# Flash for Arduino 101 x86 and arc images, olimex_stm32_e407, 96b_carbon
 .PHONY: dfu
 dfu:
+ifeq ($(BOARD), arduino_101)
 	dfu-util -a x86_app -D $(A101BIN).dfu
 	dfu-util -a sensor_core -D $(A101SSBIN).dfu
+endif
+ifeq ($(BOARD), olimex_stm32_e407)
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/olimex_stm32_e407/zephyr.bin --dfuse-address 0x08000000
+endif
+ifeq ($(BOARD), 96b_carbon)
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/96b_carbon/zephyr.bin --dfuse-address 0x08000000
+endif
 
 # Give an error if we're asked to create the JS file
 $(JS):
--- a/Makefile
+++ b/Makefile
@@ -230,10 +230,10 @@ ifeq ($(BOARD), arduino_101)
 	dfu-util -a sensor_core -D $(A101SSBIN).dfu
 endif
 ifeq ($(BOARD), olimex_stm32_e407)
-	dfu-util -a 0 -d 0483:df11 -D $(OUT)/olimex_stm32_e407/zephyr.bin --dfuse-address 0x08000000
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/olimex_stm32_e407/zephyr/zephyr.bin --dfuse-address 0x08000000
 endif
 ifeq ($(BOARD), 96b_carbon)
-	dfu-util -a 0 -d 0483:df11 -D $(OUT)/96b_carbon/zephyr.bin --dfuse-address 0x08000000
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/96b_carbon/zephyr/zephyr.bin --dfuse-address 0x08000000
 endif
 
 # Give an error if we're asked to create the JS file
--- a/Makefile
+++ b/Makefile
@@ -222,7 +222,7 @@ ide: zephyr
 .PHONY: ashell
 ashell: zephyr
 
-# Flash for Arduino 101 x86 and arc images, olimex_stm32_e407, 96b_carbon
+# Flash images
 .PHONY: dfu
 dfu:
 ifeq ($(BOARD), arduino_101)


cuiyanx:
Of course, you can use `make flash BOARD=<board name> JS=<js case>` to build and flash.

nagineni:
Does 'make dfu' works for olimex_stm32_e407 board? I am using an external ST-Link USB dongle and  openocd tool to debug or flash the board. 

https://github.com/zephyrproject-rtos/zephyr/blob/master/boards/arm/olimex_stm32_e407/support/openocd.cfg

http://docs.zephyrproject.org/boards/arm/olimex_stm32_e407/doc/olimex_stm32_e407.html#programming-and-debugging

cuiyanx:
@nagineni 
For olimex_stm32_e407 board:
1. USB connects to USB_OTG1
2. PWR_SEL: pin7 and pin8 are connected
3. DFU mode: B0->1  and  B1->0
4. `dfu-util -a 0 -d 0483:df11 -D <ZJS path>/outdir/olimex_stm32_e407/zephyr.bin --dfuse-address 0x08000000`

cuiyanx:
For 96b_carbon board:
1. BOOT0 -> 1
2. Reset

nagineni:
@cuiyanx Nice! Looked into the board manual for those pins now and looks perfect.  https://www.olimex.com/Products/ARM/ST/STM32-E407/resources/STM32-E407.pdf

nagineni:
@grgustaf

Makefile
```diff
@@ -247,11 +247,19 @@ ide: zephyr
 .PHONY: ashell
 ashell: zephyr
 
-# Flash Arduino 101 x86 and arc images
+# Flash for Arduino 101 x86 and arc images, olimex_stm32_e407, 96b_carbon
```
jimmy-huang: maybe replace with a simpler line like "Flash images", instead of listing every board in the same line, since we'll grow to support more boards, and put any board-specific comments below before each board if neccessary.

Makefile
```diff
@@ -247,11 +247,19 @@ ide: zephyr
...
+ifeq ($(BOARD), arduino_101)
 	dfu-util -a x86_app -D $(A101BIN).dfu
 	dfu-util -a sensor_core -D $(A101SSBIN).dfu
+endif
+ifeq ($(BOARD), olimex_stm32_e407)
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/olimex_stm32_e407/zephyr.bin --dfuse-address 0x08000000
+endif
+ifeq ($(BOARD), 96b_carbon)
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/96b_carbon/zephyr.bin --dfuse-address 0x08000000
+endif
```
nagineni: I figured it out that `make BOARD=<BOARD> flash` is working very well with any supported boards. So, wondering why we need to handle these board specific flashing commands (except A101) here. I tested with olimex_stm32_e407, stm32f4_disco and 96b_carbon, and no issues flashing those boards using `make BOARD=<BOARD> flash`.

Makefile
```diff
@@ -247,11 +247,19 @@ ide: zephyr
...
 
-# Flash Arduino 101 x86 and arc images
+# Flash for Arduino 101 x86 and arc images, olimex_stm32_e407, 96b_carbon
 .PHONY: dfu
 dfu:
+ifeq ($(BOARD), arduino_101)
 	dfu-util -a x86_app -D $(A101BIN).dfu
 	dfu-util -a sensor_core -D $(A101SSBIN).dfu
+endif
+ifeq ($(BOARD), olimex_stm32_e407)
```
nagineni: `make flash` invokes openocd tool by default for olimex_stm32_e407, but I think it makes sense to push this `dfu-util` option to Zephyr. A101 already supports flashing using JTAG or dfu-util in upstream Zephyr.

Makefile
jimmy-huang: Yes, make flash will work, like I've pointed out to @grgustaf that we might want to switch from make dfu to make flash for a101 as well, since it would wait for you to press reset instead of you having to flash within 5 secs. Since a101 flashes two images all at once, might be little tricky to get it working

Makefile
```diff
@@ -247,11 +247,19 @@ ide: zephyr
...
-# Flash Arduino 101 x86 and arc images
+# Flash images
 .PHONY: dfu
 dfu:
+ifeq ($(BOARD), arduino_101)
 	dfu-util -a x86_app -D $(A101BIN).dfu
 	dfu-util -a sensor_core -D $(A101SSBIN).dfu
+endif
+ifeq ($(BOARD), olimex_stm32_e407)
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/olimex_stm32_e407/zephyr.bin --dfuse-address 0x08000000
```
jimmy-huang: this should be $(OUT)/olimex_stm32_e407/zephyr/zephyr.bin

Makefile
```diff
@@ -247,11 +247,19 @@ ide: zephyr
...
+endif
+ifeq ($(BOARD), 96b_carbon)
+	dfu-util -a 0 -d 0483:df11 -D $(OUT)/96b_carbon/zephyr.bin --dfuse-address 0x08000000
```
jimmy-huang: this should be $(OUT)/96b_carbon/zephyr/zephyr.bin

grgustaf:
FYI @nagineni the purpose of "make dfu" is to handle A101 case where we have two images x86/arc to flash at once. Doing them one at a time doesn't make sense because of the base address trick where we ignore the default partition boundary.

jimmy-huang:
it might make sense to get rid of make dfu, and we just check when calling make flash that if BOARD is set to a101, then do the sitching trick, and else, pass it along to Zephyr's make flash, at least then the build workflow is the same and we get rid of make dfu.

grgustaf:
Another reason I created `make dfu` is so that I would never again accidentally flash over my board with JTAG... since I can potentially brick my board that way. I still like that reason for preserving it. So I guess I like this patch!

grgustaf:
+1

PR #1693, opened 2017-12-07
[pinmux] Fix D8 pin and support other GPIOs on k64f

grgustaf:


jimmy-huang:
+1

PR #1694, opened 2017-12-07
[build] Set the Intel's VID and PID for A101 device

nagineni:
SSIA


nagineni:
@jimmy-huang This is needed for IDE to connect to the device after updating Zephyr deps to v1.10.x.

jimmy-huang:
@nagineni thanks, works for me, +1

PR #1700, opened 2017-12-08
[build] Build linux with CMake

jimmy-huang:


README.md
```diff
@@ -445,10 +445,9 @@ command:
 make BOARD=linux
 ```
 
-The executable will be outputted to `outdir/linux/<variant>/jslinux`. Where
-`<variant>` is either `debug` or `release`. This is specified the same as the
-Zephyr target by passing in `VARIANT=` when running `make`. The default is
-release.
+The executable will be outputted to `outdir/linux/jslinux`.  If `variant` is
+set to `debug`. The debug executable will be stored in `outdir/linux/debug/jslinux`.
```
grgustaf: This 'If' sentence is incomplete and hence makes no sense...

grgustaf: Why only have one build? That's what we do for other boards, but it's kind of nice to have both and switch between them when debugging something.

jimmy-huang: they do build and the output is in different directories, and they don't conflict, so you could switch between them. The release sits just 1 directory higher now, so it's outdir/linux/jslinux instead of outdir/linux/release/jslinux

jimmy-huang:
@pfalcon, do you have any insights on why my patch is getting this weird linker error with the jerry libs in Travis?  It builds fine locally on my linux box and mac, but only fails on Travis CI.

pfalcon:
@jimmy-huang : I can only confirm that this PR builds on top of the latest master  without problems for me. From the build log, in Travis it can't find (when, doesn't use) jerryscript's libraries and fails on undefined symbols from them. But why? No idea. I'd start with checking which cmake version is used in Travis, to excludeany discrepancies due to that. I'm using 3.8.2. Perhaps, you could add cmake --version to Travis?

jimmy-huang:
@pfalcon so yeah, it turns out that it's a problem of gcc 4.8.2 on travis, which is too old and when it's linking, it failed because the link command order was wrong, if I upgrade it to gcc 5, the problem goes away.

pfalcon:
Ah, good investigative work!

jimmy-huang:
Yeah, thanks to a coworker @lpereira who helped root caused it.

grgustaf:
+1

PR #1704, opened 2017-12-11
[readme] Update support modules for more boards

haoxli:
Support more x86 and arm boards:
- tinyTILE
- ST STM32F4DISCOVERY
- OLIMEX STM32 E407
- 96Boards Carbon


grgustaf:
Maybe we should use shorter Zephyr names like 96b_carbon to keep the table thinner. Or perhaps break it into two tables, maybe on a separate page. But no reason to hold the patch back.

+1

Issue #1706, opened 2017-12-13
CMake version race


pfalcon:
Trying to build the latest master:

~~~
CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
  CMake 3.9 or higher is required.  You are running version 3.8.2
~~~

Really? Can't you guys stick with the version which Zephyr requires, which is 3.8.2, and which is not available as a native package for many Linux distros, so require additional pain on the user to install?

qiaojingx:
Verified with commit b32969f. This issue is not reproduced.

PR #1708, opened 2017-12-13
[build] Downgrade cmake version check to 3.8.2

pfalcon:
Zephyr 1.10 requires CMake 3.8.2. For some actively used Linux distros
(e.g. Ubuntu 16.04), that version is not in repositories, and users
have to do extra legwork to install it. Then Zephyr.js pushes it even
higher. Try to use the value consisten with what Zephyr uses instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Copyright (c) 2017, Intel Corporation.
 
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.8.2)
 
 if(NOT DEFINED ENV{ZEPHYR_BASE})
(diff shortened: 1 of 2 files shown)


pfalcon:
Addresses #1706. Building for BOARD=qemu_x86, I see no ill effect in this. 393af84a1860225c36886f2b6760efbc85f5bf4b which introduced 3.9 check, doesn't say why exactly 3.9 is used, so I'm not sure if there would be any ill effects.

jimmy-huang:
+1, i think I was a little too aggressive assuming user will manually need to install cmake and they would install the latest version 3.9.5+, should've stick with the minimum Zephyr required.

grgustaf:
+1

PR #1716, opened 2017-12-14
Pins removal

grgustaf:


Makefile.linux
```diff
@@ -11,7 +11,7 @@ BUILD_DIR = $(O)/$(VARIANT)
 
 UNAME := $(shell uname)
 
-CORE_SRC +=	src/zjs_a101_pins.c \
+CORE_SRC +=	\
```
brianjjones: Delete this extra \

docs/boards/frdm_k64f.md
```diff
@@ -0,0 +1,85 @@
...
+
+### GPIO Pins
+The FRDM-K64F has 16 general purpose I/O pins, D0 - D15.
+
+If you look at the diagrams above, the pin D0 in the bottom right corner is
+also labeled PTC16. This means it's pin 16 on the 'C' GPIO controller. The full
+pin specification for that pin is "GPIO_2.16" (2 because the controllers A-E are
+0-indexed, 0 to 4). All the other pins listed with a blue box of the format
+PTC16 can also be used as GPIOs. For example, the pins that can also be used as
+analog inputs (A0 - A5) also have GPIO names. (These work as GPIOs unless you
```
brianjjones: Should we consider adding a DBG_PRINT for these kind of multi use pin cases?  So the user could get some feedback on why the pin wasn't behaving as expected?  Doesn't have to be done in this patch, but might be something to consider for the future.

docs/boards/frdm_k64f.md
```diff
@@ -0,0 +1,85 @@
...
+The FRDM-K64F has twelve pins that can be used as PWM output. For ZJS purposes,
+we have named them PWM0 - PWM11. These correspond to the purple labels on the
+diagrams mentioned above. We've numbered them sequentially starting with the
+bottom right (D3/PTA1) upward, and then with A4 and A5 as PWM10 and PWM11.
+
+When support is added, you will be able to refer to these with names like
+"PWM0" but also "D3" and "A4" to identify them more clearly based on the
+diagram.
+
+However, PWM support for K64F is not working currently.
```
brianjjones: I might suggest putting this line first in case someone impatient like myself didn't read all the way to the and and wondered why PWM didn't work :)

docs/boards/frdm_k64f.md
grgustaf: I guess you mean if it's configured as an AIO but they try to use it as a GPIO... Yeah, I guess that would be possible but we'd want to be careful to not increase code size for release case. So this is not too bad right now because it will fail and tell them "invalid pin" which might bring them to the docs to investigate. I'll add it to my infinite TODO list. :)

docs/boards/frdm_k64f.md
grgustaf: Good point.

grgustaf:
Addressed your feedback, Mr. Jones.

src/zjs_aio.c
```diff
@@ -193,66 +131,59 @@ static ZJS_DECL_FUNC(zjs_aio_pin_read_async)
...
-    if (pin < AIO_MIN || pin > AIO_MAX) {
-        DBG_PRINT("PIN: #%u\n", pin);
-        return zjs_error("pin out of range");
+    char devname[20];
+    int pin = zjs_board_find_aio(pin_val, devname, 20);
+    if (pin == FIND_PIN_INVALID) {
+        return TYPE_ERROR("bad pin argument");
+    }
+    else if (pin == FIND_DEVICE_FAILURE) {
+        return zjs_error("device not found");
```
brianjjones: This string appears 3 times, should probably be a #define or something similar.

brianjjones:
Aside from the string that is repeated several times, this looks good.

src/zjs_aio.c
grgustaf: Yeah, a system for doing that consistently would be nice, but as long as it's an exact match for the string it's not really a problem... the compiler notices they're identical and uses the same static string.

PR #1718, opened 2017-12-15
docs: Fix readme to reflect zephyr docs for cmake

jprestwo:
The ZJS cmake instructions were wrong (and did not seem to work)
for ZJS. Updated the the same steps that zephyr uses. These
steps are also much easier and do not require actually building
cmake.


brianjjones:
Hmmm the old instructions worked for me.

grgustaf:
Yeah, me too but this is what Zephyr says to do now and is a lot faster.

+1

PR #1719, opened 2017-12-15
[WiP] Add experimental support for IDE protocol over WebUSB over UART.

zolkis:
This is for preliminary testing of the feature, expect changes.
A corresponding change has been done to zephyrjs-ide [here](https://github.com/intel/zephyrjs-ide/pull/211).

The IDE protocol is similar to the current console, with a few differences:
- Fixed format, no user input (User input is by pressing buttons and selecting options in ZephyJS IDE).
- Client sends commands that look much like command lines, the device responds with JSON replies.
- Streams from device to client are delimited by a start JSON reply and an ending JSON reply.
- From client to device, commands have a preamble, command, separator, [argument(s)], postamble.
- Currently the markers are  '{', '}', ' ', '$' and '#' (replaceable later with control characters, but it's much easier to test with these). 
- Example commands from client to device:  
    * `{ls}` (list files), 
    * `{rm <file>}`,
    * `{save test.js $console.log("Hello");#}`
- Streams can contain any characters and can span over multiple buffers sent on WebUSB. Streams are ended by (and files are written when) a combination of stream end character and postamble are found at the end of a buffer sent on WebUSB. 
- Currently WebUSB works on UART transport, but the code is prepared for switching to alternative raw USB transport.
- For testing purposes the current console mode can be compiled on the device side by resetting the `ASHELL_IDE_PROTOCOL` define in `ashell.h`. With this IDE option off, the device should work exactly as now in console mode, so this PR could be merged with IDE protocol turned off since it is orthogonal to the current functionality. When makefiles were still in use, this was a make command line option, now it's a conditional define for time being. On the client side, see the corresponding option [here](https://github.com/intel/zephyrjs-ide/pull/211).
- Even when IDE protocol is enabled, the user/tester can type in commands in zephyrjs-ide; input is buffered on the device side until entered. This code (the `ide_receive()` function in the `ide-comms.c` file) will be removed when raw WebUSB will be enabled.

Signed-off-by: Zoltan Kis <zoltan.kis@intel.com>


src/ashell/ashell.c
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2016-2017, Intel Corporation.
```
grgustaf: If it's newly written, drop the 2016.

src/zjs_ashell.json
```diff
@@ -18,12 +18,15 @@
...
+        "src/ashell/ide-comms.c",
+        "src/ashell/ide-cmd.c",
         "src/ashell/file-utils.c",
         "src/ashell/jerry-code.c",
         "src/ashell/term-cmd.c",
         "src/ashell/term-ihex.c",
         "src/ashell/term-uart.c",
-        "src/ashell/term-webusb.c",
+        "src/ashell/webusb-uart.c",
         "${ZEPHYR_BASE}/samples/subsys/usb/webusb/src/webusb_serial.c",
```
grgustaf: Note that Brian just moved this line to zjs_ashell_ide.json which probaby causes you a merge conflict.

src/ashell/ashell.c
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @file
+ * @brief Ashell main.
+ */
+
+#include "ashell.h"
+
+void ide_init()
+{
+    extern void parser_init();
```
brianjjones: I personally would prefer to have header files to #include vs using extern.  Is there a benefit to this method that I'm not aware of?

src/ashell/ashell.h
```diff
@@ -3,6 +3,11 @@
 #ifndef __ashell_h__
 #define __ashell_h__
 
+// Switches between  console and IDE protocol mode.
+#define ASHELL_IDE_PROTOCOL   1
```
brianjjones: I'm sure this was just for testing, but ASHELL_IDE_PROTOCOL should be set in the Makefile based on whether we are building the IDE or CLI version of ashell.  The same goes for ASHELL_IDE_DBG, though it should likely be a build argument to turn it on.

src/ashell/ashell.c
```diff
@@ -0,0 +1,29 @@
...
+    parser_init();
+}
+
+void zjs_ashell_init()
+{
+    extern void uart_init();
+    uart_init();
+#if (ASHELL_IDE_PROTOCOL == 1)
```
brianjjones: #if ASHELL_IDE_PROTOCOL, instead for consistency.

src/zjs_ashell.json
```diff
@@ -18,12 +18,15 @@
     ],
     "targets": ["arduino_101"],
     "src": [
+        "src/ashell/ashell.c",
+        "src/ashell/ide-comms.c",
```
brianjjones: The IDE specific files should be moved to zjs_ashell_ide.json

src/zjs_ashell.json
```diff
@@ -18,12 +18,15 @@
...
+        "src/ashell/ide-cmd.c",
         "src/ashell/file-utils.c",
         "src/ashell/jerry-code.c",
         "src/ashell/term-cmd.c",
```
brianjjones: The CLI specific files should be moved to zjs_ashell_cli.json

brianjjones: The CLI specific files should be moved to zjs_ashell_cli.json

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,108 @@
+// Copyright (c) 2016-2017, Intel Corporation.
```
brianjjones: As Geoff mentioned, this should just be 2017

brianjjones:
Just some early comments as I dive deeper.  What is the plan to reduce the ROM size for ashell?  Currently this adds 900+ lines of code and only removes 15.  Either the unused code for CLI needs to be #ifdef out, or we need to better separate the CLI and IDE code.  I'd personally prefer the latter as its easier to read, maintain, and debug.  Also are we keeping ihex?

Thanks for this I'll keep reviewing and trying it out tomorrow.

zolkis:
The plan was to have CLI mode with UART+WebUSB, or IDE mode with raw WebUSB. 

This IDE + UART + WebUSB is a temporary mode. However, we can get rid of iHex in this mode. Also, the code implementing the commands could be made common (or exclusive) with conditional defines. So yes, there are ways to reduce ROM size.

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,108 @@
...
+{
+    static char rx_buffer[P_SPOOL_SIZE + 1];
+    static size_t rx_cursor = 0;
+    extern void ide_parse(char *buf, size_t len);
+    char *buf = (char *) buffer;
+
+    if (len == 1) {  // character mode; buffer a command
+        ide_send_buffer(buf, 1);  // echo
+        if (rx_cursor >= P_SPOOL_SIZE || *buf == '\n' || *buf == '\r') {
+            //IDE_DBG("\r\nReceived %u chars.\r\n", rx_cursor);
```
brianjjones: Please remove all commented IDE_DBG prints

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,108 @@
...
+#include <string.h>
+
+// ZJS includes
+#include "ashell.h"
+#include "ide-comms.h"
+
+// Process a buffer (part of a message) in WebUSB driver context.
+void ide_receive(u8_t *buffer, size_t len)
+{
+    static char rx_buffer[P_SPOOL_SIZE + 1];
```
brianjjones: I don't like the idea of having static buffers in ashell as it means that memory is always held up, even when we aren't sending / receiving data from webUSB.  @grgustaf any opinions on the matter?

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,651 @@
...
+    } else if (buf[len] != '\0') {
+        IDE_DBG("\r\nParser received invalid buffer.");
+        return;
+    }
+    // IDE_DBG("\r\nReceived %u bytes.", len);
+
+    int ret = 0;
+    // Parser states allow resuming parsing with successive input buffers.
+    // Since all commands (except save) fit in one input buffer, it is enough to
+    //   handle idle and stream states as possible entry points.
```
brianjjones: Remove extra spaces in the comment

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,108 @@
...
+#else
+int ide_send_buffer(char *buf, size_t len)
+{
+    return webusb_write((u8_t *) buf, len);
+}
+#endif  // ASHELL_IDE_UART
+
+// Spooled write for webusb-raw is needed, but already there for webusb-uart.
+// Temporarily keeping here for testing webusb-raw as well.
+static char spool[P_SPOOL_SIZE + 1];
```
brianjjones: Same concern as above, this is allocating memory the JS app can't have access too once it starts running.

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,108 @@
...
+}
+
+int ide_spool_space()
+{
+    return P_SPOOL_SIZE - spool_cursor;
+}
+
+void ide_spool_adjust(size_t size)
+{
+    spool_cursor += size;
```
brianjjones: Needs a check that adding the size doesn't make spool_cursor > P_SPOOL_SIZE

src/ashell/ide-comms.h
```diff
@@ -0,0 +1,113 @@
...
+void ide_receive(u8_t *buf, size_t len);
+
+// Init WebUSB with a receiver callback.
+void ide_init();
+
+// For periodic jobs called from mainloop. Not used.
+void ide_process();
+
+// Send arbitrary data over WebUSB.
+int ide_send_buffer(char *buffer, size_t size);
```
brianjjones: Unless ide_send_buffer, ide_spool, ide_spool_flush,and ide_spool_space are going to possibly return error flags, they should return unsigned int instead of signed int.

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,108 @@
...
+static u32_t spool_cursor = 0;
+
+char *ide_spool_ptr()
+{
+    return spool + spool_cursor;
+}
+
+int ide_spool_space()
+{
+    return P_SPOOL_SIZE - spool_cursor;
```
brianjjones: Check that this number isn't negative.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,651 @@
...
+#ifdef CONFIG_BOARD_ARDUINO_101
+  #include <flash.h>
+#endif
+
+// JerryScript includes
+#include "jerry-code.h"
+
+// ZJS includes
+#include "../zjs_util.h"
+// #include "../zjs_util.h"
```
brianjjones: Remove commented code

src/ashell/ide-comms.c
zolkis: Of course I will, but I still need them for the raw webusb testing. This PR was made as a fallback (on WebUSB+UART), to give opportunity to early tests/comments etc. When the WiP flag is removed, all these will be cleaned up.

src/ashell/ide-comms.c
zolkis: Throughout zephyr there are a lot of places static buffers are used and held up and I agree this is wrong. In this particular case this is here only to make the _current_ zephyrjs-ide work with the IDE protocol, since the ide is sending each character separately at the moment. This will go away when WiP flag is removed.

src/ashell/ide-comms.c
zolkis: UART is also allocating memory the JS app cannot use, but this is part of the driver...
In  raw WebUSB, I can put this in the driver or here, but it must be somewhere. It can also be dynamically allocated (as UART does), but that doesn't change the fact JS apps will have less memory. Now on the other hand, whenever one or more JS apps prints something, will go through this single buffer.

src/ashell/ide-comms.c
zolkis: Which would need an 'unlikely' statement since it's only used in driver context with size being one or two, and the boundaries are checked when spool_cursor is used.

src/ashell/ide-comms.h
zolkis: They should return negative when there is an error, and positive (size) on success. Also, the underlying USB routines return int.  So `int` is appropriate.

src/ashell/ide-comms.c
zolkis: In the context this function is used this value can only be positive. But in general, you are right.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,651 @@
...
+}
+
+int ide_reply(int status, char *message)
+{
+    ide_spool("{\"reply\": %s, \"status\":%d, \"data\": %s }",
+              cmd_arg_map[parser.cmd_id].cmd, status, message);
+    ide_spool_flush();
+    if (!(status == NO_ERROR && parser.state == PARSE_ARG_STREAM))
+        parser_reset();
+    return -status;
```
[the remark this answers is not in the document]
zolkis: Yes. As mentioned in the PR, error numbers are declared (and transmitted) as positive integers, but in the code their negative value is used.

src/ashell/ide-comms.c
zolkis: Corrected, to be committed soon.

zolkis:
Based on the discussion on yesterday's call, if I can't get raw WebUSB work reliably this week, we will merge the WebUSB+UART transport first. I decided to do that right away since that also takes some work. The plan is the following:
- I will gather a minimal UART functionality needed to run the IDE protocol. Will leave this in the ashell json file, and move the rest to ashell-cli json. The ashell ide json will have the current ide functionality.
- Strip the new impl from raw WebUSB specific code and all other DBG garbage.
- Strip code that makes it work with the current character-based mode of zephyrjs-ide.
- Modify zephyrjs-ide to work with the IDE protocol by sending buffers and with the terminal by sending characters. This will be only half finished during this week. The UI needs redesign. For testing/intermediary solution we'll need a text box for entering an IDE protocol messages for testing.   @brianjjones if you want to contribute on that it would be really nice.

Later on I will continue work on raw WebUSB as well, but since time is short, this is a safer way to get to a more or less usable (interim) functionality.

src/ashell/ashell.c
zolkis: Done

src/ashell/ashell.c
zolkis: Will do this a bit later, but before merging (when removing the WiP flag). Then we'll include what is strictly needed.

src/ashell/ide-comms.c
zolkis: Done

src/ashell/ide-comms.c
zolkis: Done

zolkis:
I have added a make target to build the IDE with terminal (as it was):
- `make ide_term ROM=256` builds the IDE with terminal (WebUSB + UART)
- `make ide ROM=256` builds with the IDE protocol support (now with WebUSB + UART)
- `make ashell` builds for the command line.

I have kept the support for current character-mode zephyrjs-ide. This can be removed when the zephyrjs-ide is updated for the IDE protocol.

I have kept spooling (with a smaller buffer), needed for supporting file reads and assembling IDE messages. On UART transport it could be replaced with `comms_printf()`, but the change is too big and then I need to use another buffer for file reading (`cat` command). What we could do is make the spool allocation on-demand - but since this is quite frequently used (with every message), the benefit is to be investigated.

xiaoyu3x:
@zolkis 
I justed this PR.
1.`make ide_term ROM=256` can build and use normally in console of IDE.
2. `make ide ROM=256` can build, but not use in console of IDE.
selection_024

3.`make ashell ROM=256` can build, but there isn't any output by 'minicom -D /dev/ttyACM0'
It should be invalid.

zolkis:
For 1., at least that is good :).

For 2., it should work. Just click in the console and start to type. You can still type commands and see the echo, but there is no prompt. This is a temporary way to test the functionality: zephyrjs-ide needs to be updated to use the IDE messages, then the console will be only for showing output from the device.

For 3., unfortunately I cannot test since I am on holiday and don't have that cable.

xiaoyu3x:
@zolkis 
You are right, `make ide ROM=256` can build and it can be used in console of IDE.

src/ashell/ide-comms.h
brianjjones: Sounds good, just making sure.

cuiyanx:
@zolkis I submit PR to fix ashell cli mode output on ACM0. PR: [zolkis/zephyr.js/pull/1](https://github.com/zolkis/zephyr.js/pull/1).
@xiaoyu3x Please check it again, thx.

xiaoyu3x:
hi, @cuiyanx 
[zolkis/zephyr.js/pull/1](https://github.com/zolkis/zephyr.js/pull/1) is OK

zolkis:
The latest changes work with the following make targets:
1. `make cli` 
2. `make ide_term` will make the original command line based IDE
3. `make ide` will make the IDE protocol version, with the new raw webusb driver (not in this PR)
4. `make ide_serial` will make the IDE protocol version, with the UART transport.

We can test the latter two (3,4) against each other and against 2.
Since there will be one WebUSB driver in zephyr, once 3. is tested, it will replace the current WebUSB driver over UART.

Here is the .zip of the new driver. Unzip it in `deps/zephyr/samples/subsys/usb` and it will work with `make ide`.
[webusb-ide.zip](https://github.com/intel/zephyr.js/files/1665207/webusb-ide.zip)

xiaoyu3x:
hello, @zolkis 
Sorry that I reply you so late

About the 3 point `make ide`, I can't connect device successfully and the following is error.
selection_057

The following is test step:
```
1. git clone https://github.com/zolkis/zephyr.js.git
2. git checkout ashell-ide-protocol  && make update
3. cp -rf webusb-ide deps/zephyr/samples/subsys/usb
4. make ROM=256 ide & make dfu
```
```
5. git clone https://github.com/zolkis/zephyrjs-ide
6. git checkout  ashell-ide-protocol
7. npm install & npm start
```
Why happen this issue? Your advice?

xiaoyu3x:
@zolkis 
I run `make cli`, `make ide_term` and `make ide_serial`. Result is as follows.

> Precondition put webusb-ide to deps/zephyr/samples/subsys/usb

1.`make cli`
selection_059

2. `make ide_term` 
selection_058

3. `make ide_serial`
Use the IDE protocol version with the UART transport successfully.

I have two questions:
1. What is `make cli` function and why fail it is.
2. What's meaning 
>  Since there will be one WebUSB driver in zephyr, once 3. is tested, it will replace the current WebUSB driver over UART.

I put webusb-ide to deps/zephyr/samples/subsys/usb and use 'make ide_term' normally.

brianjjones:
@zolkis Can you look at fixing the file conflicts when you get a chance? Thanks.

brianjjones:
@zolkis  OK Think we are near the end of the tunnel!  I added a few minor changes I needed to get things working, as well as moved from using file-utils.h to src/zjs_file_utils.h (it was renamed).  Please rebase one last time to solve the conflicts and hopefully we can merge this tomorrow.

xiaoyu3x:
hi, @zolkis 
I just re-test this PR and found two issues.
First, I can't merge PR into zjs master by manual way.
selection_088

Second, `make cli` and `make ide `still had issues and result are at the same last time.
`make cli`
selection_085

`make ide`
selection_087

Glad to see that `make ide_serial` and `make ide_term` also  are well.

zolkis:
The problem is we have 2 webusb drivers now, and some of the targets are using the new driver which is not in zephyr yet.

We agreed with @poussa and @nagineni that we remove all but 2 targets: cli (using ashell uart + acm) and ide (using the IDE protocol and the new webusb driver).

The changes done by @brianjjones used the IDE protocol over UART (using the old webusb driver). So the zephyrjs-ide needs to be changed to use that, and also ashell needs to be cleaned up. 

The only question mark is how to do the transition between the webusb drivers. I suggested using both of them for a while (plus merge this PR now - after fixing the errors), and then deprecate the old webusb driver. However, @poussa and @nagineni are on the opinion we should just move to the new one after testing it extensively, i.e. we keep this PR open until that is done.

zolkis:
Attached the new webusb-ide driver. Copy it in `deps/zephyr/samples/subsys/usb` and extract. Then the `ide` target should work as well.

[webusb-ide.zip](https://github.com/intel/zephyr.js/files/1833106/webusb-ide.zip)

brianjjones:
@zolkis Where is this driver coming from? We can't expect people to download a zip and extract it.  Can we add this driver to the PR? Is it open source?

zolkis:
We need to add the driver to zephyr, but only after we have extensively tested it. Catch 22? :)

brianjjones:
Ah, is there an open PR for it on the Zephyr github?  Also I got everything working today, but still need to clean it all up. Once I do that I'll add my needed changes here and on the IDE.

zolkis:
OK, I (temporarily) added the webusb driver to ashell, used by the `make ide` target. After tested, we can move it out to Zephyr to replace the current webusb-uart driver.

brianjjones:
@zolkis  Also just a side note, with the new web-usb driver I don't see console.log on the IDE side. It prints out through serial. In case you know of a quick fix for that :)

zolkis:
I wrote the driver, actually it's a modified version of Babu's webusb driver cleaned up from CDC ACM stuff. I added new buffer management (API), tx, rx, and application code (ide_webusb.c). I need to update it to the latest syntactic changes wrt descriptor declaration, but it is functional.

However, you need to use a different interface (0) instead of 2. Also, there are no console printouts with this, only IDE messages. The console printouts come from reply messages, but support is needed in the zephyrjs-ide. Currently that is hardcoded to use webusb over UART and needs to be changed.

I was about to do that but for clarity (and for avoiding merge hell since I am modding a lot of code), are you to commit some changes first?

xiaoyu3x:
hi, @zolkis @brianjjones 

I just retested this PR, meanwhile used [#241](https://github.com/intel/zephyrjs-ide/pull/241) at the side of IDE.
It is failed to build `make cli`, `make ide_term` and `make ide_serial` 
selection_091
selection_092

To `make ide`, it can build and flash into A101, but can't operate normally after connected WebIDE.
selection_090

Console window can show output information, but can't input.

Makefile
```diff
@@ -330,10 +330,10 @@ ${JERRY_BASE}/CMakeLists.txt: update
 -.PHONY: setup
 setup: ${JERRY_BASE}/CMakeLists.txt
 ifeq ($(ASHELL), ashell)
-ifeq ($(filter ide,$(MAKECMDGOALS)),ide)
-	@echo CONFIG_USB_CDC_ACM=n >> prj.conf
-else
+ifeq ($(filter ide,$(MAKECMDGOALS)),cli)
 	@echo CONFIG_USB_CDC_ACM=y >> prj.conf
+else
```
jimmy-huang: One of these cases is not needed, the default is either y or n, then depending on the condition, you enable it or disable it

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
+// Copyright (c) 2017-2018, Intel Corporation.
+
+// C includes
+#include <ctype.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <string.h>
+#include <malloc.h>
```
jimmy-huang: what is is malloc.h  used for? Isn't this running on Zephyr?

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+
+// Zephyr includes
+#include <atomic.h>
+#include <misc/printk.h>
+#include <misc/reboot.h>
+#include <zephyr/types.h>
+
+#ifdef CONFIG_BOARD_ARDUINO_101
+  #include <flash.h>
```
jimmy-huang: no indent

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+#endif
+
+// JerryScript includes
+#include "jerry-code.h"
+
+// ZJS includes
+#include "../zjs_util.h"
```
jimmy-huang: this is not right,  should just set the include paths when you build instead of doing ../

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+
+// Local includes
+#include "ashell.h"
+#include "../zjs_file_utils.h"
```
jimmy-huang: same here

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+#include "jerryscript-port.h"
+#include "ide-comms.h"
```
jimmy-huang: maybe sort these.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+
+size_t ide_spool_space()
+{
+    int space = P_SPOOL_SIZE - spool_cursor;
+    return space > 0 ? space : 0;
+}
+
+void ide_spool_adjust(size_t size)
+{
+    if (spool_cursor + size < P_SPOOL_SIZE)
```
jimmy-huang: add curly braces

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+}
+
+int ide_spool_flush()
+{
+    if (spool_cursor == 0) {
+        return 0;
+    }
+    int size = ide_send_buffer(spool, MIN(spool_cursor, P_SPOOL_SIZE));
+    spool_cursor = 0;
+    // memset(spool, 0, P_SPOOL_SIZE + 1);
```
jimmy-huang: is this not needed? remove if so

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+}
+
+int ide_end_message(int status)
+{
+    ide_spool(", \"status\": %d }\r\n", status);
+    return ide_spool_flush();
+}
+
+#ifdef ASHELL_IDE_DBG
+#define IDE_DBG(...) do { \
```
jimmy-huang: align back slashes

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+// Reentrant: may be called multiple times until a full message is received.
+void ide_parse(char *buf, size_t len) {
+    IDE_DBG("\r\nEntering ide_parse...");
+
+    if (len == 0 || buf == NULL) {
+        IDE_DBG("\r\nParser received empty buffer.");
+        return;
+    } else if (buf[len] != '\0') {
+        IDE_DBG("\r\nParser received invalid buffer.");
+        // return;
```
jimmy-huang: no return here?

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+            break;
+        }
+        if (entry.type != FS_DIR_ENTRY_DIR) {
+            p = entry.name;
+            for (; *p; ++p) {
+                *p = tolower((int)*p);
+            }
+            if (first) {
+                ide_spool("\r\n\t{ \"name\": \"%s\", \"size\": %d }",
+                    entry.name, entry.size);
```
jimmy-huang: align

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+                first = false;
+            } else {
+                ide_spool(",\r\n\t{ \"name\": \"%s\", \"size\": %d }",
+                    entry.name, entry.size);
```
jimmy-huang: align

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+        fs_close_alloc(file);
+        ide_reply(ERROR_FILE, "\"Cannot write boot.cfg.\"");
+    }
+
+    fs_close_alloc(file);
+    ide_reply(NO_ERROR, "ok");
+}
+
+#ifdef CONFIG_BOARD_ARDUINO_101
+  #include <flash.h>
```
jimmy-huang: no indentation

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+#endif
+
+#ifdef CONFIG_REBOOT
+// TODO Waiting for patch https://gerrit.zephyrproject.org/r/#/c/3161/
+  #ifdef CONFIG_BOARD_ARDUINO_101
```
jimmy-huang: no indentation

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+      #include <qm_init.h>
+  #endif
+#endif
+
+// reset automatic run, then reboot
+static void ide_cmd_reboot(char *buf, size_t len)
+{
+    IDE_DBG("\r\nInvoking reboot...");
+    ide_reply(NO_ERROR, "ok");
+    #ifdef CONFIG_REBOOT
```
jimmy-huang: same here, no indentation

src/ashell/ide-comms.c
```diff
@@ -0,0 +1,78 @@
...
+
+// Zephyr includes
+#include <atomic.h>
+#include <device.h>
+#include <init.h>
+
+// ZJS includes
+#include "ashell.h"
+#include "ide-comms.h"
+#include "../zjs_util.h"
```
jimmy-huang: remove ../ and add include path

src/ashell/ide-comms.h
```diff
@@ -0,0 +1,67 @@
...
+// Init WebUSB with a receiver callback.
+void ide_init();
+
+// For periodic jobs called from mainloop.
+void ide_process();
+
+// Send arbitrary data over WebUSB.
+int ide_send_buffer(char *buffer, size_t size);
+
+#if 0
```
jimmy-huang: remove this section?

src/ashell/ide-comms.h
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2017-2018, Intel Corporation.
+
+#ifndef __ashell_comms_h__
+#define __ashell_comms_h__
+
+#include <arch/cpu.h>
```
jimmy-huang: split into to C includes and Zephyr includes

src/ashell/ide-webusb.c
```diff
@@ -0,0 +1,193 @@
...
+* @brief WebUSB specific configuration file
+*
+* Provides WebUSB and Microsoft OS Descriptor 2.0 Platform Capability
+* descriptors. Also, it handles the WebUSB Device Requests and provides
+* the landing page and allowed origin descriptors.
+*/
+
+#include <stdio.h>
+#include <string.h>
+#include <device.h>
```
jimmy-huang: split into to C includes and Zephyr includes

src/ashell/ide-webusb.c
```diff
@@ -0,0 +1,193 @@
...
+ *        WebUSB Platform Capability Descriptor.
+ *
+ * @param pSetup    Information about the request to execute.
+ * @param len       Size of the buffer.
+ * @param data      Buffer containing the request result.
+ *
+ * @return  0 on success, negative errno code on fail
+ */
+int custom_handle_req(struct usb_setup_packet *pSetup,
+    s32_t *len, u8_t **data)
```
jimmy-huang: align here

src/ashell/ide-webusb.c
```diff
@@ -0,0 +1,193 @@
...
+ * @brief Handler called for WebUSB vendor specific commands.
+ *
+ * @param pSetup    Information about the request to execute.
+ * @param len       Size of the buffer.
+ * @param data      Buffer containing the request result.
+ *
+ * @return  0 on success, negative errno code on fail.
+ */
+int vendor_handle_req(struct usb_setup_packet *pSetup,
+    s32_t *len, u8_t **data)
```
jimmy-huang: align here

src/ashell/ide-webusb.c
```diff
@@ -0,0 +1,193 @@
...
+{
+  /* Get URL request */
```
jimmy-huang: this file should uses 4 spaces for indentation like others.

src/ashell/ide-webusb.c
```diff
@@ -0,0 +1,193 @@
...
+static void webusb_receive(u8_t *data, size_t len)
+{
+    webusb_rx_buf_t *buf = buffer_from_data(data);
+    buf->length = len;
+    k_fifo_put(&rx_queue, buf);
+}
+
+// Initialize Custom and Vendor request handlers, and buffer handling callbacks.
+static struct webusb_req_handlers req_handlers = {
+  .custom_handler = custom_handle_req,
```
jimmy-huang: 4 spaces

src/ashell/term-uart.c
```diff
@@ -311,7 +316,8 @@ static int uart_out(int c)
     char ch = (char)c;
     buf[size++] = ch;
     if (ch == '\n' || size == 80) {
-        terminal->send(buf, size);
+        //terminal->send(buf, size);
```
jimmy-huang: remove commented line?

src/ashell/term-uart.c
```diff
@@ -339,24 +345,17 @@ static void uart_ready()
     uart_irq_tx_disable(dev_upload);
 
     uart_irq_callback_set(dev_upload, uart_interrupt_handler);
-    terminal->send(banner, sizeof(banner));
+    // terminal->send(banner, sizeof(banner));
```
jimmy-huang: same here

src/ashell/webusb_driver.h
```diff
@@ -0,0 +1,146 @@
...
+#define WEBUSB_VENDOR_ID   0x8086
+
+/* Product Id, random value */
+#define WEBUSB_PRODUCT_ID  0xF8A1
+
+/* BOS descriptor type */
+#define DESCRIPTOR_TYPE_BOS     0x0f
+
+/* Number of configurations for the USB Device */
+#define WEBUSB_NUM_CONF    0x01
```
jimmy-huang: align these

src/ashell/webusb_driver.h
```diff
@@ -0,0 +1,146 @@
...
+#define WEBUSB_TX_BUFFER_SIZE  WEBUSB_BULK_EP_MPS
+#define WEBUSB_RX_BUFFER_SIZE (2 * WEBUSB_BULK_EP_MPS)
+
+/* Size in bytes of the configuration sent to
+ * the Host on GetConfiguration() request
+ * For Communication Device: CONF + (3 x ITF) +
+ * (5 x EP) + HF + CMF + ACMF + UF -> 67 bytes
+ */
+#define WEBUSB_SERIAL_CONF_SIZE   (USB_CONFIGURATION_DESC_SIZE + \
+	(1 * USB_INTERFACE_DESC_SIZE) + (2 * USB_ENDPOINT_DESC_SIZE))
```
jimmy-huang: no tabs

src/ashell/webusb_driver.h
```diff
@@ -0,0 +1,146 @@
...
+ * Callback function to request a buffer of size WEBUSB_RX_BUFFER_SIZE
+ * from the application. Returns NULL if no buffer is available.
+ */
+typedef u8_t* (*webusb_buffer_provider) ();
+
+/**
+ * WebUSB request handlers, run in interrupt context, so keep them simple.
+ */
+struct webusb_req_handlers {
+	/* Handler for WebUSB Vendor specific commands */
```
jimmy-huang: no tabs

Makefile
brianjjones: Turns out neither is needed since they are provided by the .json files, so removing the whole logic block.

src/ashell/ide-cmd.c
brianjjones: Apparently nothing, removed...

xiaoyu3x:
hi, @zolkis 
Thanks to your reply。
but now my ide doesn't work normally. Console window can show output information, but can't input.
What have precondition about ide? Please help me.

zolkis:
The console window does not take any input any more, it only shows output. You can only press buttons in the IDE and they will result in sending IDE messages to the device and handle the replies.

src/ashell/ide-webusb.c
grgustaf: Yeah, I like to first include C lib stuff, then include Zephyr stuff, then finally include ZJS stuff... i.e. most standard to least standard. FWIW.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,741 @@
...
+// Check if the number of arguments left to be parsed matches with the input.
+static int check_argc(size_t argc, char *buf, size_t len)
+{
+    IDE_DBG("\r\nChecking %d argument(s).", argc);
+    if (argc == 0 && !match_postamble(buf)) {
+        return ide_reply(ERROR_UNEXPECTED_CHAR, "\"Postamble expected.\"");
+    } else if (argc > 0) {
+        if (match_postamble(buf)) {
+          return ide_reply(ERROR_MISSING_ARG, "\"Missing argument.\"");
+        } else if (!match_separator(buf)) {
```
grgustaf: Again I think the postamble check is superfluous - this separator check will also catch it.

Maybe I'm missing something - are these messages to the human IDE user? Is "Missing argument" more helpful to them than "Separator expected"? Do they need to distinguish the cases?

src/ashell/jerry-code.c
```diff
@@ -15,18 +15,18 @@
...
-#include "../zjs_sensor.h"
-#include "../zjs_timers.h"
-#include "../zjs_util.h"
+#include "zjs_buffer.h"
+#include "zjs_callbacks.h"
+#include "zjs_ipm.h"
+#include "zjs_modules.h"
+#include "zjs_sensor.h"
+#include "zjs_timers.h"
+#include "zjs_util.h"
```
grgustaf: Nice fix, I take it you added it to the path - why didn't we do this before. :)

src/ashell/webusb_driver.c
```diff
@@ -0,0 +1,430 @@
...
+     */
+    USB_STRING_DESC_SIZE,           /* Descriptor size */
+    USB_STRING_DESC,                /* Descriptor type */
+    0x09,
+    0x04,
+
+    /* Manufacturer String Descriptor "Intel" */
+    0x0C,
+    USB_STRING_DESC,
+    'I', 0, 'n', 0, 't', 0, 'e', 0, 'l', 0,
```
grgustaf: It would sure be nice as you look to upstreaming this, if you could make these strings configurable... I know it's not that way in the current driver either.

src/ashell/jerry-code.c
grgustaf: Nice fix, I take it you added it to the path - why didn't we do this before. :)

src/ashell/webusb_driver.c
grgustaf: It would sure be nice as you look to upstreaming this, if you could make these strings configurable... I know it's not that way in the current driver either.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,741 @@
...
+            return ide_reply(ERROR_UNEXPECTED_CHAR, "\"Separator expected.\"");
+        }
```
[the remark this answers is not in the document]
brianjjones: They're visible if you use the debugger, so useful to know why they failed. But yeah maybe combine them and just say "Malformed message"?

src/ashell/jerry-code.c
jimmy-huang: because i told him so :)

brianjjones:
@grgustaf Hopefully I fixed all the issues you found, let me know if there's anything else I should address.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,743 @@
...
+    if (parser.stream_fp != NULL) {
+        fs_close_alloc(parser.stream_fp);
+    }
+    parser.stream_fp = NULL;
+}
+
+void parser_init()
+{
+    parser_reset();
+}
```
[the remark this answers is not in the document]
zolkis: `parser_init()` is the global name as used elsewhere to initialize parsing. In this particular case we arrived at only invoking `parser_reset()` for init - which is also being used elsewhere in this file on its own. Of course we could compact it by scraping `parser_init()` and making `parser_reset()` non-static.

src/ashell/webusb_driver.c
zolkis: Thanks, point taken.

src/ashell/ide-comms.h
brianjjones: Done.

src/ashell/ide-comms.h
brianjjones: Done.

src/ashell/ide-webusb.c
brianjjones: Done.

src/ashell/ide-webusb.c
brianjjones: Done.

src/ashell/ide-webusb.c
brianjjones: Done.

src/ashell/ide-webusb.c
brianjjones: Done.

src/ashell/ide-webusb.c
brianjjones: Done.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,741 @@
...
+    IDE_DBG("\r\nInvoking cat...");
+    char *filename = buf;
+    fs_file_t *file;
+    int count, ret;
+
+    if ((ret = parse_filename(buf, len, 1)) > 0) {
+        skip_size(&buf, &len, ret);
+        IDE_DBG("\r\nFilename: %s\r\n", filename);
+
+        if (!fs_exist(filename) || !(file = fs_open_alloc(filename, "r"))) {
```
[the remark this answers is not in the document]
brianjjones: Yes it sure does, I'll simplify this thanks.

src/ashell/ide-cmd.c
```diff
@@ -0,0 +1,741 @@
...
+    int size = parse_filename(buf, len, 1);
+    if (size <= 0)
+        return;  // error has already been handled and parser is reset
+    skip_size(&buf, &len, size);
+    IDE_DBG("File: %s", name);
+
+    if (!fs_exist(name)) {
+        ide_reply(ERROR_FILE, "\"File not found.\"");
+        return;
+    }
```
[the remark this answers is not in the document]
brianjjones: I'll take it out. Since now the IDE doesn't let users use the console, in theory the chance to try and delete a non-existent file is basically nil. Not to mention yes the unlink should just return an error.

src/ashell/term-uart.c
```diff
@@ -339,24 +344,16 @@ static void uart_ready()
     uart_irq_tx_disable(dev_upload);
 
     uart_irq_callback_set(dev_upload, uart_interrupt_handler);
-    terminal->send(banner, sizeof(banner));
+    uart_write_buf(banner, sizeof(banner));
 
     /* Enable rx interrupts */
     uart_irq_rx_enable(dev_upload);
-    DBG("[Listening]\n");
```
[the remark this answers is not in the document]
brianjjones: Whoops

src/ashell/term-uart.c
```diff
@@ -34,23 +34,17 @@
 #include <fs.h>
 #include <misc/printk.h>
 
-// JerryScript includes
-#include "jerry-code.h"
-
 // ZJS includes
-#include "term-cmd.h"
+#include "ashell.h"
```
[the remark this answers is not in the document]
brianjjones: Done

src/ashell/term-uart.c
```diff
@@ -297,6 +291,17 @@ void uart_write_buf(const char *buf, int len)
     uart_irq_tx_disable(dev_upload);
 }
 
+/**
```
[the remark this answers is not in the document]
brianjjones: I'll try and make it a bit more consistent.

grgustaf:
Note: commit b3b06d1 was the last one before a giant rebase.

grgustaf:
Brian created a new PR #1870 with the rebase instead of wiping out this branch. So I'm closing this...

PR #1721, opened 2017-12-16
[ashell] Fix broken ashell CLI

brianjjones:
Including this webusb_serial.c file with the ashell CLI causes
it to never get events, which means it hangs and waits forever
after the first print.  Moving the file to where it belongs.


grgustaf:
+1

Issue #1722, opened 2017-12-18
[k64f] Use onboard LED lead to reset


qiaojingx:
### Description

This is a regression issue. When using onboard LED, FRDM-K64F will keep resetting.
 
### Test Code

[ButtonLEDs-k64f.js](https://github.com/01org/zephyr.js/blob/master/samples/ButtonLEDs-k64f.js);
[Blink-k64f.js](https://github.com/01org/zephyr.js/blob/master/samples/arduino/basics/Blink-k64f.js); 
[test-gpio.js](https://github.com/01org/zephyr.js/blob/master/tests/test-gpio.js); 

### Steps to Reproduction

1. `make BOARD=frdm_k64f JS=samples/ButtonLEDs-k64f.js`
2. `cp outdir/frdm_k64f/zephyr/zephyr.bin /media/<username>/MBED/`

### Actual Result

K64F keeps resetting.
 
### Expected Result

Onboard LED is blinking. The color of this led changes from pink to blue when pressing SW2
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 616c13f | FRDM-K64F | Dec 18, 2017 | Fail
master | 1b970c8 | FRDM-K64F | Dec 11, 2017 | Pass

qiaojingx:
Verified with commit 8c9d2b4. This issue is fixed.

PR #1723, opened 2017-12-18
[pinmux] Add check pin name as string

cuiyanx:
Fixed： #1722 


grgustaf:
Whoops, thanks! +1

grgustaf:
Actually I think this fix is not right, but you're right that it's buggy. I'll submit a PR.

grgustaf:
I submitted #1726 for this; closing.

PR #1724, opened 2017-12-18
[tests] Update a101 pins test case and add test points

cuiyanx:
**TOTAL:** 88 of 88 passed


grgustaf:
Thanks @cuiyanx, sorry I meant to email you that both this and test-k64f-pins.js would need to be updated.

grgustaf:
+1

PR #1727, opened 2017-12-18
[board] Move board support defines into src/boards files

grgustaf:
Also, add src/boards/README.md to clarify the intention


jimmy-huang:
+1

PR #1734, opened 2017-12-28
[build] Making changes needed to rebuild when ROM / RAM changes

brianjjones:
Currently if you want to change the ROM / RAM size for an Arduino 101 build you have to do either make clean or make pristine.  This change fixes it so that if either value changes since the last build, it will rebuild to update the values.


jimmy-huang:
+1

Issue #1710, opened 2017-12-13
[webusb] No notification pops up for WebUSB.


qiaojingx:
### Description

This issue a regression issue. There is no pop up notification when running WebUSB.

### Test Code

[WebUSB.js](https://github.com/01org/zephyr.js/blob/master/samples/WebUSB.js);
[WebUSBConfig.js](https://github.com/01org/zephyr.js/tree/master/samples/WebUSBConfig.js);
[WebUSBConfigGroveLCD.js](https://github.com/01org/zephyr.js/blob/master/samples/WebUSBConfigGroveLCD.js);
[test-webusb-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-webusb-manual.js);
 
### Steps to Reproduction

1. `make JS=samples/WebUSB.js`
2. `make dfu`
 
### Actual Result

There is no pop up notification when running WebUSB.

 
### Expected Result

A notification pops up.
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 1b970c8 | Arduino 101 | Dec 13, 2017 | Fail
master | cbec79d | Arduino 101 | Dec 5, 2017 | Fail

brianjjones:
Fix is ready for review.  CONFIG_USB_DEVICE_VID and CONFIG_USB_DEVICE_PID have to be added to the .json file for Chrome to display the popup.  I also updated the sample URL to the new one.

qiaojingx:
Verified with commit b32969f + #1738 . This issue is not reproduced.
But URL in [WebUSBConfig.js](https://github.com/01org/zephyr.js/tree/master/samples/WebUSBConfig.js), [WebUSBConfigGroveLCD.js](https://github.com/01org/zephyr.js/blob/master/samples/WebUSBConfigGroveLCD.js), [test-webusb-manual.js](https://github.com/01org/zephyr.js/blob/master/tests/test-webusb-manual.js) need to update, too.

brianjjones:
+1 LGTM

qiaojingx:
Verified with commit b17a496， this issue is fixed.

PR #1738, opened 2017-12-28
[webusb] Fixing webusb popup and sample URL

brianjjones:
Fix for #1710


jimmy-huang:
+1

PR #1751, opened 2018-01-11
[readme] Add info about libpcap-dev dependency for QEMU networking

grgustaf:
Also, clarify that net-tools scripts should be run from inside that
directory since loop-slip-tap.sh looks for tunslip6 there.

Also, fix that sudo is not needed for loop-socat.sh but is for
loop-slip-tap.sh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -738,10 +738,15 @@ cd net-tools
 make
 ```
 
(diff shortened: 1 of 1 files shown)


jimmy-huang:
+1

PR #1752, opened 2018-01-11
[build] Fixed build parameter PRINT_FLOAT not working

jimmy-huang:
This fixes the issue that Newlib is not being built with float printing
support when PRINT_FLOAT=on is passed to the command line.


grgustaf:
+1

PR #1757, opened 2018-01-12
01org update

grgustaf:


README.md
```diff
@@ -30,17 +30,29 @@ everything else.
 
 ## Getting Started
 
+ZJS should work on most of the boards supported by the Zephyr OS, though it
```
jimmy-huang: This is only with non-ide apps, the web-ide will have issue with most if not all of these board except a101, so maybe you should point it out.

Also, I think we should probably require user to pass BOARD=arduino_101, so there's no default BOARD set to a101 automatically similiar to what Zephyr does.   One of the reason is that "make flash BOARD=xxx" will work right for these boards if you pass it the right board.    But just typing "make flash" alone will try to flash using a101 instructions.   So this will force us to always build/flash with the correct board and not confused.

grgustaf: I do say that in the next paragraph. :)

I agree on BOARD=, I've been saying that too...

jimmy-huang: yeah, i missed that part, you did over it.

jimmy-huang:
+1

PR #1760, opened 2018-01-12
[build] Fix incorrect GCC version set for Mac builds on Travis

jimmy-huang:
GCC5 should only be needed on Ubuntu 14.04 builds, not MacOS
since it uses clang.


grgustaf:
+1

PR #1762, opened 2018-01-16
[aio] Increase analog sampling rate

### LINKED ISSUE #1661: [aio] 'couldn't read from pin 11' on A101
### Description

When use A1 on Arduino101, it will print `ipm_console0: '52:(ERROR) couldn't read from pin 11' `sometimes.
 
### Test Code

[AIO.js](https://github.com/01org/zephyr.js/blob/master/samples/AIO.js);
 
### Steps to Reproduction

1. `make JS=samples/AIO.js`
2. `make dfu`
### Actual Result

selection_005
 
### Expected Result

No error
 
### Test Builds

---|---|---|---|---
master | dcf56a9 | Arduino 101 | Nov 16, 2017 | Fail

jimmy-huang:
This will fix the couldn't read from pin error when AIO
is reading too fast.

Fixes #1661


arc/src/arc_aio.c
```diff
@@ -32,7 +32,7 @@ void arc_aio_cleanup()
 u32_t arc_pin_read(u8_t pin)
 {
     struct adc_seq_entry entry = {
-        .sampling_delay = 26, // anything smaller than 26 have error
+        .sampling_delay = 30,
```
grgustaf: Should we really remove the comment? Seems like useful information... is it no longer true?

jimmy-huang: @grgustaf we could leave the comment, but in reality, I don't know if anyone is going to be tweaking this value, or if they would gain any benefit by tweaking this.  All the Zephyr samples default to 30, which we had before, but I remember we lowered it to get rid of the the overhead that AIO read was causing, I remember you looked into this bug before, but I just tested the latest Zephyr, it seemed they might have fixed it.  So I think it's okay to leave it at 30 just as like other Zephyr samples.   But if you would prefer to leave the comment, I am okay with it.

grgustaf:
+1

PR #1764, opened 2018-01-17
[gpio] Fix memory leak through unnecessary acquire

### LINKED ISSUE #1363: GPIO memory leak
The pin object created at line 301 isn't being freed when the javascript stops.  This is likely due to another reference of pin_object being created somewhere, so jerryscript never calls zjs_gpio_free_cb.

grgustaf:
Since the native pointer struct held a reference to the pin object,
neither would ever get freed.

Fixes #1363


grgustaf:
@brianjjones, if you know how you performed the test to find this bug, would you mind repeating it?

jimmy-huang:
+1

PR #1765, opened 2018-01-17
[linux] Fix segfault when running jslinux with no args

grgustaf:


jimmy-huang:
+1

PR #1768, opened 2018-01-17
[build] Add support for 96b_carbon and olimex_stm32_e407

jimmy-huang:
Add additional target configs for ARM boards so that apps that
require BLE or networking support.


grgustaf:
+1

Issue #1670, opened 2017-11-20
[ws] BUS FAULT occurs when connect and disconnect websocket server for several times.


qiaojingx:
### Description

Try to connect and disconnect with websocket server for several times, `BUS FAULT` error will happen.

### Test Code

[WebSocketServer.js](https://github.com/01org/zephyr.js/blob/master/samples/websockets/WebSocketServer.js);
 
### Steps to Reproduction

 1. `make JS=samples/websockets/WebSocketServer.js BOARD=frdm_k64f`
 2. `cp outdir/frdm_k64f/zephyr.bin /media/<username>/MBED/`
 3. `node samples/websockets/NodeWebSocketClient.js`
 4. CTRL+C to disconnect 
 5. repeat steps 3 and 4(error happens after about 5 times).

### Actual Result

![selection_006](https://user-images.githubusercontent.com/19813546/33008929-0082013c-ce10-11e7-99c5-3c4df3fc28fe.png)

 ### Expected Result

 No error occurs.

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | dcf56a9 | FRDM-K64F | Nov 20, 2017 | Fail

jimmy-huang:
@qiaojingx Is this reproducible with master, it seemed the original check was with Zephyr 1.9, but I can't reproduce this on Zephyr 1.10.  However, I do get ip route dropped on the linux time periodically, and I have to add route again.

qiaojingx:
Verifed with commit 4b23a47. This issue happens with a much lower reappearance rate now(disconnect for 10+times).
![selection_021](https://user-images.githubusercontent.com/19813546/35131725-c27336dc-fd02-11e7-9114-6a504cdf5404.png)
And if keep connection from the beginning, issue #1725 will occurs.

jimmy-huang:
@brianjjones and @jprestwo were able to reproduce this yesterday and root-caused this bug, so a patch is on the way, don't close this until that's in.

qiaojingx:
Verified with commit 0c0af9b, not reproduced.

PR #1770, opened 2018-01-18
[timers] Checks to prevent deleting timers / memory that we shouldn't

brianjjones:
This is a fix for #1670

### DIFF
--- a/src/zjs_timers.c
+++ b/src/zjs_timers.c
@@ -149,8 +149,9 @@ static zjs_timer_t *add_timer(u32_t interval,
 static bool delete_timer(zjs_timer_t *tm)
 {
     if (tm) {
-	    // If the timer isn't in the list, its already been deleted
-        if (!ZJS_LIST_REMOVE(zjs_timer_t, zjs_timers, tm)) {
+        // If the timer isn't in the list, its already been deleted
+        if (zjs_timers == NULL ||
+            !ZJS_LIST_REMOVE(zjs_timer_t, zjs_timers, tm)) {
             return false;
         }
         zjs_port_timer_stop(&tm->timer);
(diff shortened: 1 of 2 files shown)


src/zjs_timers.c
```diff
@@ -149,6 +149,10 @@ static zjs_timer_t *add_timer(u32_t interval,
 static bool delete_timer(zjs_timer_t *tm)
 {
     if (tm) {
+	    // If the timer isn't in the list, its already been deleted
```
grgustaf: indentation is weird here

src/zjs_util.h
```diff
@@ -551,24 +551,26 @@ void zjs_loop_init(void);
...
-                    break;               \
-                }                        \
-                cur = cur->next;         \
-            }                            \
-        }                                \
-        removed;                         \
+#define ZJS_LIST_REMOVE(type, list, p)       \
+    ({                                       \
+        u8_t removed = 0;                    \
+        if (p != NULL && list != NULL) {     \
```
grgustaf: This should never be called with NULL. We could add an assert here to check these so that it would only run in debug mode, but we shouldn't do extra checks, we should document/enforce the calling specification.

Was this actually getting called with NULL? Why? That's the real bug.

src/zjs_timers.c
```diff
@@ -162,7 +166,6 @@ static bool delete_timer(zjs_timer_t *tm)
 #endif
             zjs_remove_callback(tm->callback_id);
         }
-        ZJS_LIST_REMOVE(zjs_timer_t, zjs_timers, tm);
```
grgustaf: Moving this worries me because this code has been modified and debugged a lot. Probably similar to my comment below, I expect the real problem here was that delete_timer was called when it shouldn't be. What was that situation?

We should only call `delete_timer` when we know there is a timer to delete and we are going to delete it. It shouldn't be "delete timer unless it isn't there after all and that's just fine".

src/zjs_util.h
brianjjones: I wasn't actually having things call with NULL, just noticed there's no check and it would fail if passed a NULL list.

jimmy-huang:
Is this reproducible with master?  I reproduced it with 1.9 but not in 1.10, i rather get a weird error where route gets dropped from the iptable and I have to re-add route periodically.

src/zjs_util.h
brianjjones: @grgustaf Quickly grepping for all the code that calls it, none of it checks if the list is NULL before calling it, so we should make sure that its not possible that the list could be NULL at that point. Rather if we don't want a NULL check here, we should review the code that calls this macro to make sure its safe.

grgustaf:
+1

PR #1778, opened 2018-01-22
[tests] Update test case to get server address

cuiyanx:
| New Test | Update Test | Delete Test | Passed | Total |
| :---: | :---: | :---: | :---: | :---: |
 | | 1 | | 1 | 1 |


grgustaf:
+1

I'm not sure what the print at the bottom is trying to say, and it might be more accurate to print '0.0.0.0' since I think we're trying to listen on all interfaces. But this isn't important to hold off the merge.

PR #1784, opened 2018-01-23
[modules] Creates a new build target dynamic

brianjjones:
This new target allows an app to run JS files from the FS, allowing
you to change what JS is running on the fly in regular non-ashell
builds. This mode includes the parser.


jimmy-huang:
I am confused by this dynamic load feature, is this just adding runJS to ashell?  So shouldn't this just be integrated as part of ashell by default?  Because looking at the changes, every check of ZJS_DYNAMIC_LOAD basically requires ZJS_ASHELL.

brianjjones:
@jimmy-huang It doesn't require ZJS_ASHELL, it just loads a lot of the same stuff that ZJS_ASHELL does. Basically its allowing you to run javascript from the file system.

src/zjs_modules.c
```diff
@@ -227,16 +230,23 @@ static ZJS_DECL_FUNC(native_print_handler)
...
-    zjs_ipm_free_callbacks();
-#endif
-#endif
     zjs_modules_cleanup();
+    zjs_remove_all_callbacks();
+    #ifdef CONFIG_BOARD_ARDUINO_101
+    #ifdef CONFIG_IPM
+        zjs_ipm_free_callbacks();
+    #endif
+    #endif
```
grgustaf: Keep the preprocessor commands at the beginning of the line, this is what we do everywhere. The function call shouldn't be indented further, just the same as the other statements. Probably this was just an accident?

src/zjs_modules.c
```diff
@@ -227,16 +230,23 @@ static ZJS_DECL_FUNC(native_print_handler)
...
     jerry_cleanup();
+    jerry_init(JERRY_INIT_EMPTY);
+    zjs_modules_init();
+}
```
grgustaf: I see that `javascript_stop()` in ashell/jerry-code.c duplicates some of this too. Could you refactor things to put that identical code in one place?

src/zjs_modules.c
```diff
@@ -24,7 +24,7 @@
 #include "zjs_timers.h"
 #include "zjs_util.h"
 #include "jerryscript-ext/module.h"
-#ifdef ZJS_ASHELL
+#if defined(ZJS_ASHELL) || defined(ZJS_DYNAMIC_LOAD)
 #include "ashell/file-utils.h"
```
grgustaf: It's not very clean for ashell subdir to depend on things from src/ AND things in src/ to depend on things from ashell like this. I guess we missed it when it went in, but it would probably be better to move these file utils out to src.

And actually we have this exact same function already in zjs_fs.c called `file_exists`.

One problem as I just commented on IRC is how to use just a utility function like that without having to include the FS JS API which may not be needed.

src/zjs_modules.c
```diff
@@ -254,6 +264,53 @@ static ZJS_DECL_FUNC(process_exit)
...
+}
+
+static ZJS_DECL_FUNC(zjs_run_js)
+{
+    ZJS_VALIDATE_ARGS(Z_STRING);
+    size_t len = 15;
+    size_t file_len;
+    load_file = zjs_malloc(len);
+
+    zjs_copy_jstring(argv[0], load_file, &file_len);
```
grgustaf: Why 15? On A101 they're limited to 8.3 I think, you would need 13 bytes including null terminator? But maybe better to not enforce that limit here and just use `zjs_alloc_from_jstring` which is simpler. Then `fs_exist` could change later (or platform-dependently) to allow different sized filenames. You do need to check for NULL on `load_file` here either way though.

src/zjs_modules.c
```diff
@@ -254,6 +264,53 @@ static ZJS_DECL_FUNC(process_exit)
...
+
+    zjs_stop_js();
+    char *buf = NULL;
+    size_t size;
+    jerry_value_t parsed_code = 0;
+    buf = read_file_alloc(load_file, &size);
+    parsed_code = jerry_parse((const jerry_char_t *)buf, size, false);
+
+    if (jerry_value_has_error_flag(parsed_code)) {
+     ERR_PRINT("Error parsing JS\n");
```
grgustaf: bad indent here and similar below

src/zjs_modules.c
```diff
@@ -254,6 +264,53 @@ static ZJS_DECL_FUNC(process_exit)
...
 #endif
+#ifdef ZJS_DYNAMIC_LOAD
+void zjs_modules_check_load_file(char *file)
+{
+    // No file waiting to load, just return
+    if (load_file == NULL) {
+        return;
+    }
+
+    zjs_stop_js();
```
grgustaf: Why do you have to wait for the next event loop to do this stop_js? It would be simpler if you just did it from run_js and you wouldn't need the global `load_file` variable. Since there's a "stopJS()" function which does that immediately that part at least must be okay?

src/ashell/file-utils.c
```diff
@@ -39,6 +39,12 @@ int fs_exist(const char *path)
 
 fs_file_t *fs_open_alloc(const char *filename, const char *mode)
 {
+#ifndef ZJS_ASHELL
+    if (mode[0] == 'w') {
+        DBG_PRINT("Write not enabled on FS for non ASHELL builds\n");
+        return NULL;
+    }
+#endif // !ZJS_ASHELL
```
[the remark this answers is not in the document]
brianjjones: This was to make the FS read only for non ASHELL builds.  The idea being once you built, the device wouldn't be able to save and run malicious JS.

src/zjs_modules.c
brianjjones: Possibly, I'll look.

src/zjs_modules.c
brianjjones: Because load_file is set from the JS API runJS.  If you stop the JS from inside a JS API call the world explodes and it has no way back.  So to get around this I mark the load_file from the JS API, then when the main loop comes back around it sees the JS requested it load a new JS. The main loop can safely stop the JS and start the new JS. Long story short, the JS has to be started and stopped from the main loop for this to work.

src/zjs_modules.c
brianjjones: I was assuming this code would be used on devices other than A101.  15 was arbitrary, I'll either use a globally defined max str length or create one.

src/zjs_modules.c
brianjjones: Yeah was able to combine the two. Now the Ashell one calls the Modules one.  Only reason I still need a separate Ashell one is because it has a couple other things it needs to do that aren't related to the Modules one.

src/ashell/file-utils.c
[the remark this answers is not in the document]
grgustaf: Yeah, but again ashell is is 'insecure mode' so I don't think we need to worry about that. They also wouldn't be able to save and run non-malicious JS, which could be cool. :)

src/zjs_modules.c
brianjjones: @grgustaf Where would you want it to live? I can just move file-utils to src/ or perhaps a new folder src/utils?

grgustaf: I think we should say ashell depends on src/, not the other way around. Probably src/zjs_file_utils.c for now, and make sure that both zjs_fs.c and ashell use this same copy of the function.

src/ashell/file-utils.c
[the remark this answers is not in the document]
brianjjones: Ah ok good point, if the only way to write stuff to the filesystem is by using ASHELL - we're already 'insecure' because who knows what they wrote to the FS.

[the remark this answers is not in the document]
grgustaf: That's true, I think you made my point better than I did. :)

brianjjones:
@grgustaf OK I think I addressed everything you mentioned, let me know if there's anything else that needs changing.

src/main.c
```diff
@@ -356,6 +358,10 @@ int main(int argc, char *argv[])
     }
 #endif
     while (1) {
+#ifdef ZJS_DYNAMIC_LOAD
+	// Check if we should load a new JS file
```
grgustaf: you fixed the next line, but this indent is still off

src/zjs_modules.c
```diff
@@ -24,24 +24,27 @@
 #include "zjs_timers.h"
 #include "zjs_util.h"
 #include "jerryscript-ext/module.h"
-#ifdef ZJS_ASHELL
-#include "ashell/file-utils.h"
+#if defined(ZJS_ASHELL) || defined(ZJS_DYNAMIC_LOAD)
+#include "file-utils.h"
```
grgustaf: Let's rename this to `zjs_file_utils.h` to follow the rest of the files there.

src/file-utils.c
```diff
@@ -210,3 +211,4 @@ char *read_file_alloc(const char *file_name, ssize_t *size)
     fs_close_alloc(fp);
     return file_buf;
 }
+#endif // defined(ZJS_ASHELL) || defined(ZJS_DYNAMIC_LOAD)
```
grgustaf: Search for 'endif //' 8x in this patch and put two spaces before //.

src/zjs_modules.c
```diff
@@ -228,16 +232,23 @@ static ZJS_DECL_FUNC(native_print_handler)
...
 {
+    zjs_modules_cleanup();
+    zjs_remove_all_callbacks();
 #ifdef CONFIG_BOARD_ARDUINO_101
 #ifdef CONFIG_IPM
-    zjs_ipm_free_callbacks();
-#endif
-#endif
-    zjs_modules_cleanup();
+        zjs_ipm_free_callbacks();
```
grgustaf: Indent still off on this function call.

src/zjs_modules.c
```diff
@@ -255,6 +266,53 @@ static ZJS_DECL_FUNC(process_exit)
     exit(status);
 }
 #endif
+#ifdef ZJS_DYNAMIC_LOAD
+void zjs_modules_check_load_file(char *file)
```
grgustaf: It appears this file argument is never used?

It also looks like you might have forgotten to declare this in a header file which might be why that wasn't caught? I would think it would be giving a warning though. I guess that gets through Travis because we're not testing this flag.

src/zjs_modules.c
```diff
@@ -255,6 +266,53 @@ static ZJS_DECL_FUNC(process_exit)
...
+    zjs_stop_js();
+    char *buf = NULL;
+    size_t size;
+    jerry_value_t parsed_code = 0;
+    buf = read_file_alloc(load_file, &size);
+    parsed_code = jerry_parse((const jerry_char_t *)buf, size, false);
+
+    if (jerry_value_has_error_flag(parsed_code)) {
+        ERR_PRINT("Error parsing JS\n");
+    }
```
grgustaf: It looks like you're going to call `jerry_run` on this even though it was an error? Maybe change this to
```c
if (not_an_error) {
    keep_going();
}
else {
    complain();
}
free_stuff_up();
```

src/zjs_modules.c
brianjjones: Indeed, left over code.

src/zjs_modules.c
brianjjones: Guess I was getting it confused with ZJS_ERROR - which would return. I'll add the check thanks.

grgustaf:
+1

PR #1786, opened 2018-01-24
Ws tweaks

grgustaf:
Some random cleanups I found to do while debugging. The message count one helps to see how long websocket sample is running before it hits an error.


jimmy-huang:
+1

PR #1791, opened 2018-01-25
[build] Removed unused NETWORK_BUILD

jimmy-huang:
This setting was used previously to include
$(ZEPHYR_BASE)/samples/net/common/Makefile.ipstack, but switching
to CMake no longer requires this.


grgustaf:
+1

PR #1797, opened 2018-01-26
[ws] Removing unneeded pong

brianjjones:
Node.js automatically returns a pong when it gets a ping,
calling ws.pong causes two to be sent to the webserver.


grgustaf:
+1, nice job!

Issue #1759, opened 2018-01-12
I'd like to fix the WebIDL in the documentation -- thoughts?


t-harvey:
Hi all,
I'm writing a WebIDL-->Jerryscript compiler/code-generator, and I've been using the excellent object-model that zephyr.js has built by copy/pasting the WebIDL out of the documentation and using it as a test bed for my project.

Unfortunately, the WebIDL is not strictly conforming to the spec, and I would like to take on the task of updating the documentation with syntactically correct WebIDL -- before I put in a pull request with my changes, I'd like to see if anyone has any strong feelings that I shouldn't?

Thanks very much,
Tim

jimmy-huang:
Yes, that would be great, anything to help make our spec WebIDL compliant is always welcomed, go ahead and submit the patch and we can review them :+1:

grgustaf:
One caveat is just yesterday I removed the WebIDL from buffer.md, in favor of a style more like what Node.js does for documentation. But you're welcome to reintroduce it if you want to correct it.

Alternately, maybe we could move it into a separate file like buffer.idl and link to that in buffer.md; I'm up for suggestions. As you see we've just been using it in a kind of ad hoc manner.

t-harvey:
Guys,
I just created a pull request (#1800) to let people (you guys? :-) examine the kinds of changes I'd like to make to the documentation.

PR #1572, opened 2017-10-05
[api] Added type definition for the Monaco-base IDE

jimmy-huang:
Creates API type definition that will add to the Intellisense which
allows code completion.

Signed-off-by: Jimmy Huang <jimmy.huang@intel.com>

PR #1800, opened 2018-01-28
Update WebIDL and modify documentation to fit new "node.js-like" style

t-harvey:
Hi -- this is a test edit to spark discussion; please don't accept this pull request!

After a short discussion (#1759), I edited three different .md files in the documentation to get feedback.

The thrust of the changes is that I want to fix the WebIDL to make it not only syntactically correct, but also to make it match the textual description of each api.

However, since Geoff Gustafson updated Buffer.md to be more like node.js documentation, I also changed a couple of the files to match this style.

I propose making similar changes across the documentation, but I would like to limit my commitment to what I've done here -- specifically, I want to make the WebIDL match the description, but my manager would prefer that correctness of the documentation itself be left to other folks :-).  For an example of this, see ble.md, where I left the description of the Characteristic API blank (seach for "TODO") (the 2nd TODO was already in the documentation).

Thanks for considering this!
Tim

### DIFF
--- a/README.md
+++ b/README.md
@@ -817,13 +817,13 @@ reason we have the following possibilities for support:
 | Module    | Linux | [A101](https://store.arduino.cc/usa/arduino-101 "Arduino 101")/[tT](https://software.intel.com/en-us/node/675623 "tinyTILE") | [K64F](https://os.mbed.com/platforms/FRDM-K64F/ "FRDM-K64F")  | [nRF52](http://infocenter.nordicsemi.com/pdf/nRF52_DK_PB_v1.2.pdf "nRF52 DK") | [Due](https://store.arduino.cc/usa/arduino-due "Arduino Due") | [F411RE](http://www.st.com/en/evaluation-tools/nucleo-f411re.html "STM32 Nucleo-F411RE") | [STM32F4](http://www.st.com/en/evaluation-tools/stm32f4discovery.html "STM32F4DISCOVERY") | [OLIMEX E407](https://www.olimex.com/Products/ARM/ST/STM32-E407/open-source-hardware "OLIMEX STM32 E407") | [Carbon](https://www.96boards.org/product/carbon/ "96Boards Carbon") |
 |   :---:   | :---: |  :---:  | :---:| :---: |:---:|  :---: |  :---:  |   :---:   |  :---: |
 |HelloWorld |   X   |    X    |   X  |   X   |  X  |    X   |    X    |     X     |    X   |
-|  ADC      |       |    X    |      |       |     |        |         |           |        |
+|  ADC      |       |    X    |   X  |       |     |        |         |           |        |
 |  PWM      |       |    X    |      |       |     |        |         |           |        |
 |  GPIO     |       |    X    |   X  |       |     |        |         |           |        |
 |  I2C      |       |    X    |   X  |       |     |        |         |           |        |
 |  BLE      |       |    X    |      |       |     |        |         |           |    X   |
 |  UART     |       |    X    |      |   NT  |     |        |         |           |        |
-| Sensor    |       |    X    |      |       |     |        |         |           |        |
+| Sensor    |       |    X    |   X  |       |     |        |         |           |        |
 | Buffer    |   X   |    X    |   X  |   X   |  X  |    X   |    X    |     X     |    X   |
 | Console   |   X   |    X    |   X  |   X   |  X  |    X   |    X    |     X     |    X   |
 | Event     |   X   |    X    |   X  |   X   |  X  |    X   |    X    |     X     |    X   |
--- a/src/zjs_callbacks.c
+++ b/src/zjs_callbacks.c
@@ -400,7 +400,7 @@ void signal_callback_priv(zjs_callback_id id,
     int key = 0;
     if (in_thread) CB_LOCK();
     if (id < 0 || id >= cb_size || !cb_map[id]) {
-        DBG_PRINT("callback ID %u does not exist\n", id);
+        DBG_PRINT("callback ID %d does not exist\n", id);
         if (in_thread) CB_UNLOCK();
         return;
     }
@@ -665,10 +665,10 @@ void zjs_defer_work(zjs_deferred_work callback, const void *buffer, u32_t bytes)
         memcpy(defer->data, buffer, bytes);
     }
 #ifdef DEBUG_CALLBACKS
-    DBG_PRINT("deferring work: %d bytes\ncontents: ", len);
+    DBG_PRINT("full packet: %d bytes\ncontents: ", len);
     int count32 = (len + 3) / 4;
     for (int i = 0; i < count32; ++i) {
-        ZJS_PRINT("0x%x ", ((u32_t *)buf)[i]);
+        ZJS_PRINT("0x%08x ", ((u32_t *)buf)[i]);
     }
     ZJS_PRINT("\nbuffer: ");
     count32 = (bytes + 3) / 4;
--- a/src/zjs_gfx.c
+++ b/src/zjs_gfx.c
@@ -31,7 +31,7 @@ typedef struct gfx_handle {
 } gfx_handle_t;
 
 typedef struct gfx_data {
-    u32_t coords[4];
+    int32_t coords[4];
     u8_t color[COLORBYTES];
     char *text;
     jerry_size_t textSize;
@@ -45,7 +45,7 @@ static jerry_value_t zjs_gfx_prototype;
 bool drawImmediate = true;
 // Draw function depends on whether we are in immediate draw mode or not.
 // It gets set by init
-void (*zjs_gfx_draw_pixels)(u32_t, u32_t, u32_t, u32_t, u8_t[], gfx_handle_t*);
+void (*zjs_gfx_draw_pixels)(int32_t, int32_t, u32_t, u32_t, u8_t[], gfx_handle_t*);
 
 static void zjs_gfx_callback_free(void *native)
 {
@@ -75,7 +75,7 @@ static void args_to_data(gfx_data_t *data, u32_t argc, const jerry_value_t argv[
         for (u8_t i = 0; i < argc; i++) {
             if (jerry_value_is_number(argv[i])) {
                 if (i < 4)
-                    data->coords[i] = (u32_t)jerry_get_number_value(argv[i]);
+                    data->coords[i] = (int32_t)jerry_get_number_value(argv[i]);
                 else
                     data->size = (u32_t)jerry_get_number_value(argv[i]);
             }
@@ -96,44 +96,58 @@ static void args_to_data(gfx_data_t *data, u32_t argc, const jerry_value_t argv[
 
 static void zjs_gfx_reset_touched_pixels(gfx_handle_t *gfxHandle)
 {
-    gfxHandle->tpX0 = gfxHandle->screenW;
+    gfxHandle->tpX0 = 0;
     gfxHandle->tpX1 = 0;
-    gfxHandle->tpY0 = gfxHandle->screenH;
+    gfxHandle->tpY0 = 0;
     gfxHandle->tpY1 = 0;
     gfxHandle->touched = false;
 }
 
-static void zjs_gfx_touch_pixels(u32_t x, u32_t y, u32_t w, u32_t h, u8_t color[], gfx_handle_t *gfxHandle)
+static void zjs_gfx_touch_pixels(int32_t x, int32_t y, u32_t w, u32_t h, u8_t color[], gfx_handle_t *gfxHandle)
 {
     // Check that x and y aren't past the screen
-    if (x >= gfxHandle->screenW || y >= gfxHandle->screenH)
+    if (x >= gfxHandle->screenW || y >= gfxHandle->screenH || (int32_t)(x + w) <= 0 || (int32_t)(y + h) <= 0)
         return;
 
-    if (gfxHandle->tpX0 > x ) {
-        gfxHandle->tpX0 = x;
-    }
-
-    if (gfxHandle->tpX1 < x + w - 1) {
-        if (x + w - 1 < gfxHandle->screenW) {
+    if (x < 0) {
+        if (x + w <= gfxHandle->screenW) {
+            gfxHandle->tpX0 = 0;
             gfxHandle->tpX1 = x + w - 1;
         }
         else {
-            gfxHandle->tpX1 = gfxHandle->screenW - 1;
-            w = gfxHandle->tpX1 - x;
+            gfxHandle->tpX0 = 0;
+            gfxHandle->tpX1 = gfxHandle->screenW;
+        }
+    }
+    else if (x < gfxHandle->screenW) {
+        if (x + w <= gfxHandle->screenW) {
+            gfxHandle->tpX0 = x;
+            gfxHandle->tpX1 = x + w - 1;
+        }
+        else {
+            gfxHandle->tpX0 = x;
+            gfxHandle->tpX1 = gfxHandle->screenW;
         }
     }
 
-    if (gfxHandle->tpY0 > y) {
-        gfxHandle->tpY0 = y;
-    }
-
-    if (gfxHandle->tpY1 < y + h - 1) {
-        if (y + h - 1 < gfxHandle->screenH) {
+    if (y < 0) {
+        if (y + h <= gfxHandle->screenH) {
+            gfxHandle->tpY0 = 0;
             gfxHandle->tpY1 = y + h - 1;
         }
         else {
-            gfxHandle->tpY1 = gfxHandle->screenH - 1;
-            h = gfxHandle->tpY1 - y;
+            gfxHandle->tpY0 = 0;
+            gfxHandle->tpY1 = gfxHandle->screenH;
+        }
+    }
+    else if (y < gfxHandle->screenH) {
+        if (y + h <= gfxHandle->screenH) {
+            gfxHandle->tpY0 = y;
+            gfxHandle->tpY1 = y + h - 1;
+        }
+        else {
+            gfxHandle->tpY0 = y;
+            gfxHandle->tpY1 = gfxHandle->screenH;
         }
     }
 
@@ -215,7 +229,7 @@ static jerry_value_t zjs_gfx_flush(gfx_handle_t *gfxHandle)
         u32_t currH = 0;
         u16_t currPass = 0;
 
-            for (u16_t j = gfxHandle->tpY0; currPass < passes ; j++) {
+        for (u16_t j = gfxHandle->tpY0; currPass < passes ; j++) {
             currH++;
             currY++;
             xStart = gfxHandle->tpX0;
@@ -269,7 +283,7 @@ static jerry_value_t zjs_gfx_flush(gfx_handle_t *gfxHandle)
     return ZJS_UNDEFINED;
 }
 
-static void zjs_gfx_fill_rect_priv (u32_t x, u32_t y, u32_t w, u32_t h, u8_t color[], gfx_handle_t *gfxHandle)
+static void zjs_gfx_fill_rect_priv (int32_t x, int32_t y, u32_t w, u32_t h, u8_t color[], gfx_handle_t *gfxHandle)
 {
     zjs_gfx_touch_pixels(x, y, w, h, color, gfxHandle);
     for (u8_t cbyte = 0; cbyte < COLORBYTES; cbyte++) {
@@ -384,7 +398,7 @@ static ZJS_DECL_FUNC(zjs_gfx_draw_line)
     if (xLen <= yLen) {
         // We always draw left to right, swap if argData.coords[0] is larger
         if (argData.coords[0] > argData.coords[2]) {
-            u32_t tmp = argData.coords[0];
+            int32_t tmp = argData.coords[0];
             argData.coords[0] = argData.coords[2];
             argData.coords[2] = tmp;
             tmp = argData.coords[1];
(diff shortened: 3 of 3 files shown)


docs/aio.md
```diff
@@ -50,57 +58,57 @@ interface AIOPin {
...
+AIOPin API
+----------
+### pin.read()
 
-Returns the latest reading from the pin. Blocks until it gets the result.
+Returns the latest reading from the pin (an unsigned integer). Blocks until it gets the result.
 
-### AIOPin.readAsync
+### pin.readAsync(ReadCallback)
+* 'ReadCallback' *callback* User-provided callback function that takes
```
grgustaf: If callback is like a type and the variable name is `ReadCallback`, let's call it `readCallback` instead to match case in other arguments. (I'm a little confused. :D)

docs/aio.md
```diff
@@ -50,57 +58,57 @@ interface AIOPin {
...
 passed for the change event, the previously registered callback will be
 discarded and no longer called.
 
-### AIOPin.close
-
-`void close();`
+### pin.close()
 
-Closes the AIOPin. Once it is closed, all event handlers registered will no
+Closes the AIOPin. Once it is closed, all registered event handlers will no
```
grgustaf: Nice English / voice edits - thanks.

docs/ble.md
```diff
@@ -58,36 +73,36 @@ dictionary CharacteristicInit {
...
+    attribute ReadCallback onReadRequest;
+    attribute WriteCallback onWriteRequest;
+    attribute SubscribeCallback onSubscribe;
+    attribute UnsubscribeCallback onUnsubscribe;
+    attribute NotifyCallback onNotify;
+	attribute CharacteristicResult response;
+};
+
+callback ReadCallback = void (unsigned long offset,
+                              FulfillReadCallback fulfillreadcallback);
```
grgustaf: This is why I'm getting confused - here `FulfillReadCallback` is treated like a type. So maybe in my earlier comment `ReadCallback` was the "type" and the variable should just be `callback`?

docs/ble.md
```diff
@@ -118,97 +133,114 @@ one previously sent with the 'accept' event.
...
+  UUIDs (four hex digits each).  These UUIDs identify available
+  services to nearby BLE devices.
+* url *string* The `url` is optional and limited to around 24
+  characters (slightly more if part of the URL is able to be
+  [encoded](https://github.com/google/eddystone/tree/master/eddystone-url). If
+  provided, this will be used to create a physical web advertisement
+  that will direct users to the given URL. At that URL they might be
+  able to interact with the advertising device somehow.
+  
+Advertises the name of the device.
```
grgustaf: name and **url**

url is really the interesting bit

grgustaf:
Also please fix the trailing whitespace issues here: https://travis-ci.org/intel/zephyr.js/jobs/334508966

docs/Notes_on_WebIDL.md
```diff
@@ -0,0 +1,33 @@
+Notes on WebIDL for zephyr.js documentation:
+
+The WebIDL fragments are for reference only; the description of the
+API should be considered correct -- please report discrepancies as
+soon as possible.
+
+Although both WebIDL and Javascript are aimed at web applications,
```
kenchris: JavaScript

docs/aio.md
```diff
@@ -23,23 +29,25 @@ not have support for this.
...
 
-[NoInterfaceObject]
+[ReturnFromRequire]
 interface AIO {
     AIOPin open(AIOInit init);
 };
 
 dictionary AIOInit {
-    unsigned long pin;
+    (unsigned long or string) pin;
```
kenchris: Many many web devs are using TypeScript today, so I see more and more web frameworks actually documenting APIs using typescript. This is also useful because you can distribute the documentation as *.ts.d files which can be used in IDE's like VSCode to do autocompletion etc.

The above would be

```
interface AIOInit {
  pin: number|string;
}

File: docs/ble.md [hunk 7]
```diff
@@ -25,21 +38,23 @@ treat them like decimals.*
...
+explanation of zephyr.js' WebIDL conventions.
+
 
 ```javascript
 // require returns a BLE object
 // var ble = require('ble');
 
-[NoInterfaceObject]
+[ReturnFromRequire]
 interface BLE: EventEmitter {
```
kenchris: And this would be

```
class BLE extends EventEmitter {
  disconnect(address: string): void;
  startAdvertising(name: string, uuids: string[], url: string): void;
  stopAdvertising(): void;
  setServices(services: PrimaryService[]) void;
  newPrimaryService(init: PrimarySeviceInit): PrimaryService;
...
```

kenchris: It is very JS like but has a bit of extra info, like types, readonly etc

docs/ble.md
```diff
@@ -58,36 +73,36 @@ dictionary CharacteristicInit {
...
+    attribute NotifyCallback onNotify;
+	attribute CharacteristicResult response;
+};
+
+callback ReadCallback = void (unsigned long offset,
+                              FulfillReadCallback fulfillreadcallback);
 callback WriteCallback = void (Buffer data, unsigned long offset,
-                               boolean withoutResponse, FulfillWriteCallback);
+                               boolean withoutResponse,
+							   FulfillWriteCallback fulfillwritecallback);
```
kenchris: really weird indentation

docs/buffer.md
```diff
@@ -20,29 +21,64 @@ Buffer is a [Node.js API](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.ht
...
+    unsigned long copy(Buffer target, optional unsigned long targetStart = 0,
+                                      optional unsigned long sourceStart = 0,
+                                      optional unsigned long sourceEnd = this.length);
+    this fill((string or Buffer or long) value, optional long offset = 0,
+                                                optional long end = this.length,
+                                                optional string encoding = "utf8");
+    octet readUInt8(optional unsigned long offset = 0);
+    short readUInt16BE(optional unsigned long offset = 0);
+    short readUInt16LE(optional unsigned long offset = 0);
+    long readUInt32BE(optional unsigned long offset = 0);
```
kenchris: in TS: `readUInt32BE(offset: number = 0): number`

docs/buffer.md
```diff
@@ -20,29 +21,64 @@ Buffer is a [Node.js API](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.ht
...
+```javascript
+[ Constructor(Uint8Array initial_values),
+  Constructor(unsigned long size),
+  Constructor(ByteString initial_string) ]
+interface Buffer {
+    readonly attribute unsigned long length;
+    unsigned long copy(Buffer target, optional unsigned long targetStart = 0,
+                                      optional unsigned long sourceStart = 0,
+                                      optional unsigned long sourceEnd = this.length);
+    this fill((string or Buffer or long) value, optional long offset = 0,
```
kenchris: You can not add "this" in an API. and same with this.length.

Could be `Buffer fill(value, offset, end = -1)` then you enterpret -1 to measure yourself or so

docs/aio.md
grgustaf: @t-harvey is working on a tool to generate code based on WebIDL. So for his purposes I think he needs WebIDL specifically, or is that itself up to suggestion? Earlier I suggested maybe we should move the WebIDL itself out to a .idl file. We could also have TypeScript in a .ts.d file or both... the only question would be how well would we actually maintain them. :)

If we could turn this into some automated testing that might alert us of mismatch (because of failures) and we'd be more likely to keep it synced correctly.

docs/ble.md
```diff
@@ -58,36 +73,36 @@ dictionary CharacteristicInit {
...
-    unsigned long RESULT_SUCCESS;
-    unsigned long RESULT_INVALID_OFFSET;
-    unsigned long RESULT_INVALID_ATTRIBUTE_LENGTH;
-    unsigned long RESULT_UNLIKELY_ERROR;
-};
 ```
 
-API Documentation
------------------
+BLE-supported Events
```
grgustaf: The term "BLE-supported" doesn't make sense to me. See how they list events in the [Node dgram API](https://nodejs.org/dist/latest-v8.x/docs/api/dgram.html) - they go in the same "Class: xyz" section as the APIs below.

docs/aio.md
jimmy-huang: We did have a existing PR that had typescript before #1572, but that was providing auto-completetions, and I haven't figure out how to import it into the ide without conflicting with other .ts.d files, so i had to close the PR, but we do have similar apis.

docs/buffer.md
```diff
@@ -20,29 +21,64 @@ Buffer is a [Node.js API](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.ht
...
+and initialized with the array's contents. If there is not enough
+available memory, an error will be thrown.
 
 ### new Buffer(size)
 * `size` *integer* Length in bytes of the new buffer.
 
 The `size` argument specifies the length in bytes of the array that the Buffer
 represents. If a negative length is passed, a 0-length Buffer will be returned.
-If the size is too long and there is not enough available memory, an error will
+If there is not enough available memory to allocate the Buffer, an error will
```
grgustaf: Again, thanks for these English improvements - nice to have more eyes on it!

docs/ble.md
```diff
@@ -25,21 +38,23 @@ treat them like decimals.*
...
 interface BLE: EventEmitter {
     void disconnect(string address);
     void startAdvertising(string name, string[] uuids, string url);
     void stopAdvertising();
-    void setServices(PrimaryService services[]);
-    PrimaryService PrimaryService(PrimaryServiceInit init);
-    Characteristic Characteristic(CharacteristicInit init);
-    Descriptor Descriptor(DescriptorInit init);
+    void setServices(PrimaryService[] services);
+    PrimaryService newPrimaryService(PrimaryServiceInit init);
```
jimmy-huang: is newXXX the standard in webidl to declare a constructor?

t-harvey:
Guys,
Thanks to all three of you -- the detail-oriented suggestions will serve as a list of things I should check before checking in any more changes (I'll answer each of them inline, but the theme is: you're right, thanks for catching that, and I've checked in a better copy with your improvements :-)...My takeaways in terms of the big-picture answers to this pull request are:

1) Editing the documentation is okay, and the style that Geoff started with Buffer seems to be acceptable.  I'll probably be able to edit 3-6 files per week (there are on the order of 25 or so to do), and we can decide to have one big ol' pull request, or periodic small ones.  As I say, I'm getting ready to provide detailed comments, above, and I've checked in a new version so you can see the deltas -- as I said at the beginning, my lawyers haven't given me the final okay (only b/c they're slow, not b/c it's anything but a done deal), so this pull request can't go in for a while (anywhere from 1 day to 2-3 weeks from now would mesh with earlier experiences)...  Remember that my commitment includes minimal changes to the actual API descriptions (e.g., ble has a number of correctness issues) -- someone else will have to do those changes. :-)

2) WebIDL vs. Typescript -- I agree that Typescript is cool, but my experience over the last year is that there are a ton(!) of cool ways of doing things, and that the most important idea to get things done is to pick one and then go get things done.  (You should have seen my dismal year-end review for 2016; the preceding sentence pretty much summed it up... :-)  The advantages of staying the course with WebIDL are: 1) it's an industry standard for web developers/iot folks (which is kind of ZJS's focus, after all), 2) it's (mostly) done, 3) it now has a person (me, obviously :-) who has a vested interest in maintaining it over time, 4) using WebIDL doesn't preclude anyone else from _also_ adding Typescript et al. to the docs if they want.

docs/aio.md
```diff
@@ -23,23 +29,25 @@ not have support for this.
 
 Web IDL
 -------
-This IDL provides an overview of the interface; see below for documentation of
-specific API functions.
+
+This IDL provides an overview of the interface; see below for
+documentation of specific API functions.  Click
+[here](Notes_on_WebIDL.md) for an explanation of zephyr.js' WebIDL
+conventions.
```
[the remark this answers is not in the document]
t-harvey: Done.

docs/aio.md
```diff
@@ -50,57 +58,57 @@ interface AIOPin {
...
-The `init` object lets you set the pin number. You can either use a raw
+When setting the pin number, you can either use a raw
 number for your device or use the board support module such as
 [Arduino 101](./boards/arduino_101.md) or [K64F](./boards/frdm_k64f.md) to
 specify a named pin.
 
-Use the AIOPin object returned to read values from the pin.
-
-### AIOPin.read
+Returns an AIOPin object that be used to read values from the pin.
```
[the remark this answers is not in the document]
t-harvey: Done.

docs/aio.md
t-harvey: So I'll respond here; push the explanation to the other places it applies:

I inconsistently read the Buffer documentation and just got this wrong.

I've now edited the documents to fit the following rules (which match the examples in the WebIDL spec):

Variable names are camelCase, starting with a lower-case letter.
Type names are UpperCase.

docs/aio.md
t-harvey: I'm as human as the next guy, so I make these errors, too.
I'm also as OCD as the next guy, so expect more of these... :-)

docs/ble.md
t-harvey: As many ex-girlfriends have said to me: it's not you, it's me. :-)

The confounding inconsistency has been fixed (see above comment for more).

docs/ble.md
t-harvey: Done.

docs/Notes_on_WebIDL.md
t-harvey: Done.

docs/aio.md
t-harvey: Geoff, I'd love to move the WebIDL to its own file and then just include it in the .md file -- but markdown doesn't seem to allow such a thing (correct me if I'm wrong).  It's true that I could just have a link to a .idl file in the main page, but I tried that, and it doesn't look nearly as nice as having the WebIDL inline -- and I really like the streamlined API that WebIDL gives you in addition to the textual description; it's a lot like having a picture to reference.

Currently, I cut/paste the WebIDL out of the docs into their own .idl file and test them (which is why I missed the problem that @kenchris caught).

docs/ble.md
t-harvey: Yeah, sorry; if you ever see this again, it's b/c Emacs tabs have to be converted to spaces for the Markdown to line up correctly.

docs/ble.md
t-harvey: @grgustaf -- my tool could easily(!) be adapted to check for WebIDL syntax.  In my copious spare time, I'll gen up a simple app -- but I'll need someone who knows (useful) things to put it into the automatic testing...

...If that's what we want; I had sort of just planned on periodically running my actual tool over the code as a natural result of my testing/development and keeping the WebIDL up to date that way...

t-harvey: In re: events (i.e., "ble-supported" events)

The problem here is that each event isn't really part of the BLE object; in some ways, it's bifurcating the documentation: on one hand, most of the docs describe the interface the user sees/manipulates, etc. -- but with events (and maybe other things?), it's really addressing concerns of the guys implementing the underlying behavior.  I don't know that the text describing events even belongs in this documentation...(?)

docs/ble.md
t-harvey: ...no, I don't think there _is_ a standard WebIDL way to declare a constructor of one object as a method of some other object.

...but given that that's what this was doing, I made up a consistent naming scheme and hoped it would suffice (it gets the idea across; I didn't explain it and you intuited its function!).  As with almost anything I write, I'm not married to it and would welcome(!) better ideas.

kenchris:
Just to give some comments.

To generate bindings etc Web IDL is the way to go, that is what it is designed for and what browsers use.

TypeScript definition files are super useful, even when not using TypeScript. Like in VSCode I can just add a comment like `// @ts-check` at the top of a JavaScript file and the TypeScript compiled will be used to check and identify a lot of issues in the JavaScript code - I use it all the time :-)

For external libraries you can include these .ts.d files like I have the following in package.json for projects using Web Bluetooth:

```
   "devDependencies": {
    "@types/web-bluetooth": "0.0.4",
    ...
   }
``` 

I think that offering the same would be quite nice. I guess it is probably doable to generate these ts.d files from Web IDL as well. 

I found a few related issues here: https://github.com/Microsoft/TypeScript/issues/3027 and https://github.com/Microsoft/TSJS-lib-generator/issues/101 https://github.com/Microsoft/TSJS-lib-generator/pull/227 https://github.com/SaschaNaz/webidl-serializer

docs/ble.md
kenchris: For Web IDL please always refer to the editors draft - the stable one is very old: https://heycam.github.io/webidl/

`PrimaryService newPrimaryService(PrimaryServiceInit init);` is pretty weird

In old fashioned JS (no class) the constructor is a function which has the same name as the object, ie.

```
function PrimaryService(name) {
  this.name = name;
}
```

So in Web IDL you would just write it as

```
[Constructor(USVString name), Exposed=Window]
interface PrimaryService  {
};
```

docs/ble.md
```diff
@@ -25,21 +38,22 @@ treat them like decimals.*
...
+[ReturnFromRequire]
 interface BLE: EventEmitter {
     void disconnect(string address);
     void startAdvertising(string name, string[] uuids, string url);
     void stopAdvertising();
-    void setServices(PrimaryService services[]);
-    PrimaryService PrimaryService(PrimaryServiceInit init);
-    Characteristic Characteristic(CharacteristicInit init);
-    Descriptor Descriptor(DescriptorInit init);
+    void setServices(PrimaryService[] services);
```
kenchris: in Web IDL we don't use [], you will normally use `sequence<PrimaryService>` or `FrozenArray<PrimaryService>` depending on the exact use-case.

docs/ble.md
```diff
@@ -25,21 +38,22 @@ treat them like decimals.*
...
 
 ```javascript
 // require returns a BLE object
 // var ble = require('ble');
 
-[NoInterfaceObject]
+[ReturnFromRequire]
 interface BLE: EventEmitter {
     void disconnect(string address);
     void startAdvertising(string name, string[] uuids, string url);
```
kenchris: We also don't have strings in Web IDL, we have DOMString, ByteString and USVString. Thus here you probably want

`startAdvertising(USVString name, sequence<DOMString>, USVString url)`

docs/ble.md
```diff
@@ -58,157 +72,171 @@ dictionary CharacteristicInit {
     NotifyCallback onNotify;            // optional
 };
 
-callback ReadCallback = void (unsigned long offset, FulfillReadCallback);
+interface Characteristic {
+    attribute ReadCallback onReadRequest;
+    attribute WriteCallback onWriteRequest;
+    attribute SubscribeCallback onSubscribe;
```
kenchris: Very un-web like... should be `onsubscribe` all lowercase plus these should be nullable.

`attribute SubscribeCallback? onsubscribe`

docs/ble.md
```diff
@@ -58,157 +72,171 @@ dictionary CharacteristicInit {
...
+                               boolean withoutResponse,
+                               FulfillWriteCallback fulfillWriteCallback);
 callback SubscribeCallback = void (unsigned long maxValueSize,
-                                   FulfillSubscribeCallback);
+                                   FulfillSubscribeCallback fullfillSubscribeCallback);
 callback FulfillReadCallback = void (CharacteristicResult result, Buffer data);
 callback FulfillWriteCallback = void (CharacteristicResult result);
 callback FulfillSubscribeCallback = void (Buffer data);
 
+enum CharacteristicResult { "RESULT_SUCCESS", "RESULT_INVALID_OFFSET", "RESULT_INVALID_ATTRIBUTE_LENGTH", "RESULT_UNLIKELY_ERROR" } ;
```
kenchris: When enumeration types are for strings (preferred today) then they should all be LOWER CASE and with HYPHEN between.. not underscore.

`enum CharacteristicResult { "success", "invalid-offset"... }`

docs/buffer.md
```diff
@@ -20,29 +21,63 @@ Buffer is a [Node.js API](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.ht
...
+  Constructor(ByteString initialString) ]
+interface Buffer {
+    readonly attribute unsigned long length;
+    unsigned long copy(Buffer target, optional unsigned long targetStart = 0,
+                                      optional unsigned long sourceStart = 0,
+                                      optional unsigned long sourceEnd);
+    this fill((string or Buffer or long) value, optional long offset = 0,
+                                                optional long end,
+                                                optional string encoding = "utf8");
+    octet readUInt8(optional unsigned long offset = 0);
```
kenchris: you cannot make anything optional and also give it a default value. It is one of the other

docs/buffer.md
```diff
@@ -20,29 +21,63 @@ Buffer is a [Node.js API](https://nodejs.org/dist/latest-v8.x/docs/api/buffer.ht
...
+    short readUInt16BE(optional unsigned long offset = 0);
+    short readUInt16LE(optional unsigned long offset = 0);
+    long readUInt32BE(optional unsigned long offset = 0);
+    long readUInt32LE(optional unsigned long offset = 0);
+    string toString(string encoding);
+    long write(string value, optional long offset = 0,
+                             optional long length = this.length-offset,
+                             optional string encoding = "utf8");
+    long writeUInt8(octet value, unsigned long offset);
+    long writeUInt16BE(unsigned short value, unsigned long offset);
```
kenchris: why not optional unsigned long offset = 0 here?

docs/aio.md
grgustaf: Yeah, that cut and pasting was part of my motivation. But yes I was thinking of just having a link to it. Hopefully by standardizing the way the WebIDL appears you can write a script to cut and paste automatically. I don't have any reason to force them out into `.idl` files right now, but eventually that might be best. One thing about the "Node.js" style is the functions are all listed again in the table of contents so it's a bit repetitive.

docs/ble.md
grgustaf: This might be a misunderstanding? Those events are indeed part of the BLE object and the ZJS user can (and needs to) register functions to listen for them with the .on() method. For example, see this [BLE sample](https://github.com/t-harvey/zephyr.js/blob/master/samples/WebBluetoothGroveLcdDemo.js#L161).

grgustaf: The `on()` method comes in because BLE is an EventEmitter; perhaps this needs to be made more clear, but it does basically behave like Node now and this was an attempt to document things in the same way they do. Maybe we missed some connecting glue though. :)

grgustaf:
Part of the detail-oriented nature of my comments was that I didn't notice your comments about requesting just general discussion the first time I read it. Details are what we normally do around here. :)

But yes, absolutely editing the documentation is okay. We try to keep them up to date but that's not always in the forefront of our minds. And yes, if you want correct WebIDL then by all means we're willing to accept improvements. We'll try to understand the changes you're making so that we can make new stuff better out of the gate and maintain things properly. We haven't cared to date because basically it just served as a documentation format; specific correctness wasn't important and the couple times I looked into the spec didn't overwhelm me with a desire to read the whole thing. :)

PR #1816, opened 2018-02-01
[events] Fix mem leak of overwriting callback id

### LINKED ISSUE #1737: [tcp] `General Protection Fault` occurs when running TCPEchoServ6.js on arduino101 over ble for long time
### Description

`General Protection Fault` occurs after running TCPEchoServ6.js for 1 hour on arduino101 over ble. 
 
### Test Code

[TCPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ6.js);

### Steps to Reproduction

1. Flash Arduino101 with test sample.
```
$ make JS=samples/TCPEchoServ6.js ROM=256
$ make dfu
```
2. connect to PC via 6lowpan, add router and address.
```
modprobe bluetooth_6lowpan
echo 1 > /sys/kernel/debug/bluetooth/6lowpan_enable
echo "connect FF:EE:DD:CC:BB:AA 2" > /sys/kernel/debug/bluetooth/6lowpan_control
ip -6 route add 2001:db8::/64 dev bt0
ip -6 addr add 2001:db8::2 dev bt0
```
3. Connect TCPEchoServ6.js  using python client: 
`python test-tcp6-client.py`
 
### Actual Result

`General Protection Fault` occurs after about 1hour.
selection_001

### Expected Result

Server should always works fine.

### Test Builds

---|---|---|---|---
master |  b32969f | Arduino 101 | Dec 28, 2017 | Fail
 
### Additional Information

The matched python client is here: https://github.com/01org/web-test-suite/tree/master/zephyr.js/tools/tests

grgustaf:
Only one event callback was supposed to be created, but the line was
moved to zjs_make_emitter so that it was created for each new emitter
object. For tests that close and open sockets repeatedly, for example,
this eventually caused the system to run out of memory.

Fixes #1737

### DIFF
--- a/src/zjs_event.c
+++ b/src/zjs_event.c
@@ -12,6 +12,7 @@
 #define DEFAULT_MAX_LISTENERS   10
 
 static jerry_value_t zjs_event_emitter_prototype = 0;
+zjs_callback_id emit_id = -1;
 
 typedef struct listener {
     jerry_value_t func;
@@ -43,6 +44,8 @@ static void free_listener(void *ptr)
 static void zjs_event_proto_free_cb(void *native)
 {
     zjs_event_emitter_prototype = 0;
+    zjs_remove_callback(emit_id);
+    emit_id = -1;
 }
 
 static const jerry_object_native_info_t event_proto_type_info = {
@@ -308,8 +311,6 @@ static ZJS_DECL_FUNC(get_listeners)
     return rval;
 }
 
-zjs_callback_id emit_id = -1;
-
 typedef struct emit_event {
     jerry_value_t obj;
     zjs_pre_emit pre;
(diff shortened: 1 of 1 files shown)


src/zjs_event.c
```diff
@@ -12,6 +12,7 @@
 #define DEFAULT_MAX_LISTENERS   10
 
 static jerry_value_t zjs_event_emitter_prototype = 0;
+zjs_callback_id emit_id = -1;
```
jimmy-huang: making this global doesn't it mean that it can be overwritten with multiple events from different modules?

grgustaf: It was already global... I just moved it up from down below because it's lifetime is now tied to that to that of the prototype defined just above.

However, it could probably be static...

jimmy-huang:
+1

cuiyanx:
Verified with commit d09e4b3 on master, the issue #1737 will be fixed.

brianjjones:
LGTM +1.

PR #1818, opened 2018-02-01
[jerryscript] Rebase to latest JerryScript

jimmy-huang:
JerryScript updated cmake on how to build with external toolchains,
this patch will fix the breakage.


grgustaf:
+1

PR #1820, opened 2018-02-02
Update ocf server in 6lowpan-ble.md

xiaoyu3x:
Signed-off-by: Zhang xiaoyu<xiaoyux.zhang@intel.com>
Ocf server retrieve resource method has been altered, so update relative README


jimmy-huang:
@xiaoyu3x I thought this is only a work-around for the A101, if you call oic-get /res, the device should still respond (if they can).  So for example, if you run this on a k64f that has more ram, and if you use /res, do you not see the whole list?  Maybe @nagineni can chime in here, i recall the patch was only changes made to the iotivity-rest-api-server, but iotivity-constrained would still respond to /res queries

xiaoyu3x:
Thanks, @jimmy-huang 
According to your proposal, I just updated this PR.
Please help me to check it again.

jimmy-huang:
+1

PR #1824, opened 2018-02-04
Fix error: "[: -v: unary operator expected" in macOS build

JohnVillalovos:
The macOS uses bash 3.x and the '-v' test command is unknown and we
see the error "[: -v: unary operator expected".

Change test to use '-z' to resolve this issue.


jimmy-huang:
+1

PR #1825, opened 2018-02-04
Use "#!/usr/bin/env python3" for zjsbanner

JohnVillalovos:
On macOS builds python3 is installed in /usr/local/bin/python3 instead
of /usr/bin/python3 (which is where it is on Linux systems).

Use /usr/bin/env as it will find python3 if it is in the path.


grgustaf:
+1

You find my proudest work in ZJS. :)

PR #1830, opened 2018-02-06
[build] Simplify zjsbanner code

JohnVillalovos:
Use a dict to make the zjsbanner code simpler and hopefully easier to
understand.


grgustaf:
This code was intentionally written to be somewhat difficult to understand. :)
But I guess that has run its course, and progress must occur.

PR #1837, opened 2018-02-08
[sensor] Implements new W3C Generic Sensor API

### LINKED ISSUE #1829: Follow new Generic Sensor API
The current implementation differs from that of the spec, like onchange is now onreading.

It would be nice to follow the spec now, especially now that FitBit also follows it: https://dev.fitbit.com/build/reference/device-api/sensors/

jimmy-huang:
In the latest W3C spec, onchange event handler has been renamed to
onreading, and the state property is no longer exposed in the sensor
object.

Fixes #1829

### DIFF
--- a/docs/sensors.md
+++ b/docs/sensors.md
@@ -108,7 +108,7 @@ API Documentation
 ### onreading
 `Sensor.onreading`
 
-The onreading attribute is an EventHandler which is called whenever a new reading is available.
+The onreading attribute is an EventHandler, which is called whenever a new reading is available.
 
 ### onactivate
 `Sensor.onactivate`
--- a/docs/sensors.md
+++ b/docs/sensors.md
@@ -28,26 +28,19 @@ specific API functions.
 ####Sensor Interface
 ```javascript
 interface Sensor {
-    readonly attribute SensorState state;   // current state of Sensor object
+    readonly attribute boolean activated;   // whether the sensor is activated or not
+    readonly attribute boolean hasReading;  // whether the sensor has readings available
     readonly attribute double timestamp;    // timestamp of the latest reading in milliseconds
     attribute double frequency;             // sampling frequency in hertz
     void start();                           // starts the sensor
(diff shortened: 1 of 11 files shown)


jimmy-huang:
@kenchris I've updated the api a bit, I also saw that it now has a hasReading attribute, but I am confused on how is that value being set, maybe I don't quite understand this psudo-algorithm:

The getter of the hasReading attribute must run these steps:
 - Let timestamp be the result of invoking get value from latest reading with this and "timestamp" as arguments.

 - If timestamp is not null, return true.

 - Otherwise, return false.

So is it basically once the sensor is started, and after it gets its first reading, then the timestamp value is always non-null, and therefore hasReading will always be true then?

cuiyanx:
+1

jimmy-huang:
@kenchris any chance you can look at this patch?

kenchris:
Well when you have values available for the first time, then it is hasReading == true

docs/sensors.md
```diff
@@ -112,15 +104,15 @@ dictionary TemperatureSensorOptions : SensorOptions  {
...
+`Sensor.onreading`
 
-The onchange attribute is an EventHandler which is called whenever a new reading is available.
+The onreading attribute is an EventHandler which is called whenever a new reading is available.
 
 ### onactivate
 `Sensor.onactivate`
 
-The onactivate attribute is an EventHandler which is called when this.[[state]] transitions from "activating" to "activated".
+The onactivate attribute is an EventHandler which is called when the sensor is activated after calling start().
```
kenchris: ,before which

kenchris:
I think the patch looks fine, from skimming through it, but I think we should add the hasReading as well

jimmy-huang:
@kenchris ok, thanks, that's easy to add then.

jimmy-huang:
updated patch and added hasReading property

grgustaf:
+1

PR #1840, opened 2018-02-09
[general] Replace PRINT_FLOAT with new way of printing float

jimmy-huang:
Instead of using %f formating to print a floating point value,
use JerryScript to convert the float to a string.


scripts/trlite
```diff
@@ -334,17 +333,17 @@ if [ "$RUN" == "all" -o "$RUN" == "2" ]; then
...
 
     MODULES=(net ws)
     write_modules_test $TMPFILE $MODULES $SENSORS
 
-    # also, try printing a float w/ PRINT_FLOAT
+    # also, try printing a float
     echo "console.log(3.14159);" >> $TMPFILE
 
-    try_command "k64f net" make $VERBOSE JS=$TMPFILE PRINT_FLOAT=on ROM=256 BOARD=frdm_k64f
+    try_command "k64f net" make $VERBOSE JS=$TMPFILE ROM=256 BOARD=frdm_k64f
```
grgustaf: The point of these two sets of alternatives was to test both w/ and without. There's no point in that anymore so we could at least remove two of these... and maybe all of them. (We don't really have a ZJS feature to test anymore per se.)

jimmy-huang:
@grgustaf ok, updated patch

grgustaf:
+1

PR #1843, opened 2018-02-09
[build] analyze: Change args.verbose comparison to integer

JohnVillalovos:
In commit 490ff661a54944456145b060d179c34a8f19101e the args.verbose
parameter was changed to an integer value but the comparison was still
using a string comparison.

Change it to compare with an integer value so that --verbose=1 works
and will print extra output.


jimmy-huang:
+1

grgustaf:
+1

Issue #1833, opened 2018-02-07
[ashell] 'grovelcd.js' module can't be found when running I2CBMP280.js


xiaoyu3x:
## Issue description
'grovelcd.js' module can't be found  when running [I2CBMP280.js](https://github.com/intel/zephyr.js/blob/master/samples/I2CBMP280.js) in ashell. 
In fact, [GroveLCD.js](https://github.com/intel/zephyr.js/blob/master/modules/GroveLCD.js)  has been imported A101.

## Test Code
[I2CBMP280.js](https://github.com/intel/zephyr.js/blob/master/samples/I2CBMP280.js)
[GroveLCD.js](https://github.com/intel/zephyr.js/blob/master/modules/GroveLCD.js)
[BMP280.js](https://github.com/intel/zephyr.js/blob/master/modules/BMP280.js)

## Steps to reproduce
1.Flash ashell into 101 board, open minicom port `minicom -D /dev/ttyACM0`. 
2.Copy [GroveLCD.js](https://github.com/intel/zephyr.js/blob/master/modules/GroveLCD.js) and [BMP280.js](https://github.com/intel/zephyr.js/blob/master/modules/BMP280.js) to ashell.
3.Copy [I2CBMP280.js](https://github.com/intel/zephyr.js/blob/master/samples/I2CBMP280.js) to ashell and run it.
4.When running [I2CBMP280.js](https://github.com/intel/zephyr.js/blob/master/samples/I2CBMP280.js)  on Web IDE, get the same result as ashell

## Actual Result

(ERROR) failed to evaluate JS, NotSupporttedError: Module not found

![selection_072](https://user-images.githubusercontent.com/17694094/35896101-8a4e4f9c-0bf5-11e8-829d-f2d2f89db8fe.png)

## Expected Result
I2CBMP280.js works normally and no error.

## Test Build
Branch| Commit id | Target Device |Test Date | Result 
---| ---| ---| ---|--- 
master | [ede4509](https://github.com/intel/zephyr.js/commit/ede4509d98da5f0187b82a7867fc0277641c4ba7) | Arduino 101 | Feb 7, 2018 | Fail

jimmy-huang:
Is this a regression?  Was this broken for 0.5 release or after?

xiaoyu3x:
No, it is not a regression.
I just reproduced this issue on 0.5 branch

jimmy-huang:
[...]

PR #1846, opened 2018-02-12
[modules] Removing long lines from module examples

brianjjones:
Fix for #1833.  This will ultimately be fixed by the upcoming ASHELL / IDE refactor, but for now this is necessary to keep the IDE from chopping off lines longer than 81 chars.


jimmy-huang:
I don't think this is the right fix, more like a hack, which hides the real problem, given that this would be fixed by the upcoming IDE changes,  we should just wait for that patch instead.

Issue #1844, opened 2018-02-09
[ashell] Should stm32f4_disco and olimex_stm32_e407 support ashell like ide?


qiaojingx:
### Description

Should stm32f4_disco and olimex_stm32_e407 support ashell since they support ide now?
Build ashell target for these two boards failed.
 
### Steps to Reproduction

1.  `make BOARD=stm32f4_disco ashell`
2.  `make BOARD=olimex_stm32_e407 ashell`

### Actual Result

![selection_072](https://user-images.githubusercontent.com/19813546/36020036-307f8468-0dbc-11e8-8430-ae92d3ef1b84.png)
 
### Expected Result

Build ashell target successfully.

### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 0ae1d62| stm32f4_disco | Feb 09, 2019 | Fail
master | 0ae1d62| olimex_stm32_e407 | Feb 09, 2019 | Fail

brianjjones:
Was just missing a change to the CLI json file, I have a fix open to take care of this.

qiaojingx:
Verified with commit 9db8f2b. This issue is fixed.
Close.

PR #1849, opened 2018-02-28
Fix to allow new boards build CLI ashell

brianjjones:
Fix for #1844

### DIFF
--- a/src/zjs_ashell_cli.json
+++ b/src/zjs_ashell_cli.json
@@ -2,6 +2,6 @@
     "module": "ashell_cli",
     "depends": ["ashell"],
     "zephyr_conf": {
-        "arduino_101": ["CONFIG_USB_CDC_ACM=y"]
+        "all": ["CONFIG_USB_CDC_ACM=y"]
     }
 }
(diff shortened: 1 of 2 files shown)


brianjjones:
@jimmy-huang I thought the only reason to have different boards listed was because they required different settings? That is how its been done for all the other files.  The JSON file isn't responsible for preventing the user from building ashell with an unsupported board.  I can break it out into several boards, but they will all have the same setting, ["CONFIG_USB_CDC_ACM=y"] and that seems to add unnecessary clutter.  If we want to give a warning to the user that their board isn't supported for ASHELL we'll need to add that to the Makefile.

jimmy-huang:
I think the issue is that by putting in "all", it will just append this config to the prj.conf, which we don't want, because if later we support Ashell over some other board that has different transport or configurations, this could interfere or stop the board from working complete, so instead, it is better to just only insert this for the boards we have tested and guaranteed that it will work.  For example, currently only 3 boards  that if you are building ashell later for arduino_101", "stm32f4_disco", "olimex_stm32_e407".   We have ran into this kind of problem before where configs are added without knowing why they were added to the prj.conf, and it makes it really hard to developers later to figure out why they were needed at the first place.

brianjjones:
@jimmy-huang ok updated

jimmy-huang:
+1

PR #1850, opened 2018-03-01
Update git modules recursively if needed

rzr:
If project's sources are not cloned recursively,
or updated manualy using:

  git submodule update --init --recursive

Dependencies modules will be fetched by GNU make.

Signed-off-by: Philippe Coval <philippe.coval@osg.samsung.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -319,13 +319,14 @@ analyze: $(JS)
 
 # Update dependency repos
 .PHONY: update
-update:
-	@git submodule update --init
-	@cd $(OCF_ROOT); git submodule update --init;
+update: .gitmodules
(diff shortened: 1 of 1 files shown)


jimmy-huang:
+1

grgustaf:
+1

grgustaf:
Note #1855 will change this behavior a bit because of some perhaps unintended consequences.

jimmy-huang:
@rzr I found an issue with this patch where if you already cloned the submodules, then it will checkout the commit and so if you are working on the submodule repo, it will switch branches and goes into a detached head state, I submitted a patch here #1855 to hopefully address this issue, comment if you need.

PR #1854, opened 2018-03-07
Minor change on README file.

ziransun:
Flash images to Arduino 101 should happen pretty much immediately
after the Master Reset button is pressed. Replace "After a few seconds"
with "Within a few seconds".

Signed-off-by: Ziran Sun <ziran.sun@samsung.com>


jimmy-huang:
Yeah good catch, i think the official documentation is within 5 seconds, +1

PR #1855, opened 2018-03-08
[build] Only checkout submodule if not found

jimmy-huang:
This fixes a non-desired behavior that git submodule update
will always be called when building, and it will immediately
checkout jerryscript at the desired commit, causing that repo
to switch from the working branch into a detached head state.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -320,7 +320,9 @@ analyze: $(JS)
 # Update dependency repos
 .PHONY: update
 update: .gitmodules
-	@git submodule update --init --recursive
+	@if ! [ -f ${JERRY_BASE}/CMakeLists.txt ]; then \
+		git submodule update --init --recursive; \
+	fi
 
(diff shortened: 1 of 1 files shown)


grgustaf:
LGTM +1, I'll make up for not adequately considering the last patch by rushing this one through too. :)

Makefile
```diff
@@ -320,7 +320,9 @@ analyze: $(JS)
 # Update dependency repos
 .PHONY: update
 update: .gitmodules
-	@git submodule update --init --recursive
+	@if ! [ -f ${JERRY_BASE}/CMakeLists.txt ]; then \
+		git submodule update --init --recursive; \
+	fi
 
 ${JERRY_BASE}/CMakeLists.txt: update
```
rzr: maybe you can test is the other line

${JERRY_BASE}/CMakeList:
     [ -f $@ ] || ${MAKE} update

I should have done it in earlier patch

PR #1863, opened 2018-03-22
[build] Fix regression the make update no longer works

jimmy-huang:


grgustaf:
+1

PR #1875, opened 2018-04-19
[main] Simplified jslinux command line options

jprestwo:
jslinux was manually parsing arguments which is ugly.
The optarg functionality does argument parsing automatically.

Also, a usage print was added in the case of invalid arguments
or if --help was specified.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -17,6 +17,8 @@
 #endif
 #else
 #include "zjs_linux_port.h"
+#include <unistd.h>
+#include <getopt.h>
 #endif  // ZJS_LINUX_BUILD
 #include "zjs_script.h"
 #include "zjs_util.h"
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -81,40 +83,107 @@ static u8_t no_exit = 0;
...
+    if (!args)
+        return;
+
+    char *ptr = args;
+    char *last = args;
+    jerry_value_t array = ZJS_UNDEFINED;
+    ZVAL global_obj = jerry_get_global_object();
+    ZVAL process = zjs_create_object();
+
+    while ((ptr = strchr(last, ' '))) {
```
grgustaf: I'd prefer you do the assignment outside of the condition. So probably make it a forever loop that you break out of.

src/main.c
```diff
@@ -81,40 +83,107 @@ static u8_t no_exit = 0;
...
         }
     }
+
     return 1;
 }
+
+void init_process(char *args)
+{
+    if (!args)
+        return;
```
grgustaf: Braces, or maybe change this to an assert?

src/main.c
```diff
@@ -299,6 +368,7 @@ int main(int argc, char *argv[])
 
 #ifdef ZJS_LINUX_BUILD
     zjs_free(script);
+    init_process(js_args);
```
grgustaf: We should probably make this it's own module, conditional based on usage w/ analyze, like other things. Otherwise it's just going to be overhead. Particularly when we don't really have a use case. :)

So perhaps split that out here for a separate PR?

PR #1881, opened 2018-05-17
[net] Increase multicast addresses so neighbor solicitation will work

grgustaf:
Thanks to Paul Sokolovsky for finding this solution.


pfalcon:
Just a note about figure of 4 in "CONFIG_NET_IF_MCAST_IPV6_ADDR_COUNT=4". For a bearer like Ethernet, IPv6 requires 3 multicast addresses just to satisfy needs of the network control protocols (address resolution, etc.). And 1 extra is left for higher-level applications, and I believe that would be already used by OCF (assuming OCF requires just one multicast address, if not, you may need to bump the count higher).

jimmy-huang:
+1

Issue #1842, opened 2018-02-09
[k64f] IPv6 address '2001:db8::1' is unreachable


qiaojingx:
### Description

This is a regression issue which is caused by zephyr update. When running IPv6 Server(UDP, TCP, WS and OCF), IPv6 address '2001:db8::1' is unreachable and the corresponding client can't connect to its server.
 
### Test Code

[TCPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/TCPEchoServ6.js)
[UDPEchoServ6.js](https://github.com/01org/zephyr.js/blob/master/samples/UDPEchoServ6.js) 
[OcfServer.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfServer.js) 
[OcfSensorServer.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfSensorServer.js)
[OcfMultiSensor.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfMultiSensor.js)

### Steps to Reproduction

1. `make BOARD=frdm_k64f JS=samples/TCPEchoServ6.js `
2. `cp outdir/frdm_k64f/zephyr/zephyr.bin /media/<username>/MBED/`
3. add route and address on PC
```
ip -6 route add 2001:db8::/64 dev eno1
ip addr add 2001:db8::2 dev eno1
```
4. `echo "hello" | nc -6 2001:db8::1 4242`

### Actual Result

Connection cannot be established.
![screenshot from 2018-05-18 16-59-25](https://user-images.githubusercontent.com/19813546/40226396-292b090e-5abe-11e8-927f-a55f5ec90e5e.png)

`ping6 2001:db8::1` is unreachable.
![screenshot from 2018-05-18 16-58-27](https://user-images.githubusercontent.com/19813546/40226447-4489db12-5abe-11e8-8869-d2c9553a6946.png)
 
### Expected Result

Connection can be set up, and server side can receive data "hello".
 
### Test Builds

Branch | Commit Id | Target Device | Test Date | Result
---|---|---|---|---
master | 4324cf21 - ede4509 | FRDM-K64F | Feb 09, 2018 | Fail
master | ebd7364 | FRDM-K64F | Feb 09, 2018 | Pass
 
### Additional Information

This issue is caused by commit 4324cf2.

jimmy-huang:
this might be a zephyr bug, see https://github.com/zephyrproject-rtos/zephyr/issues/6583

[...]

PR #1886, opened 2018-05-30
[ocf] Fix OCF resources not discoverable on FRDM-K64F

### LINKED ISSUE #1885: [ocf] Ocf Client cannot connect to OcfServer on K64F
### Description

Ocf connection can't be established on K64F, while ocf works fine on Arduino101.

### Test Code

[OcfServer.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfServer.js)
[OcfClient.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfClient.js) 
[OcfSensorServer.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfServer.js)
[OcfSensorClient.js](https://github.com/01org/zephyr.js/blob/master/samples/OcfClient.js) 
### Steps to Reproduction

1. `make BOARD=frdm_k64f JS=samples/OcfServer.js `
2. `cp outdir/frdm_k64f/zephyr/zephyr.bin /media/<username>/MBED/`
3. Connect PC via Ethernet cable and add router on PC:
```
ip -6 route add 2001:db8::/64 dev eno1
ip -6 addr add 2001:db8::2 dev eno1
```
4. `./outdir/linux/release/jslinux samples/OcfClient.js --noexit`

### Actual Result

Ocf connection can't set up while ping6 2001:db8::1 is ok.
screenshot from 2018-05-24 08-33-40

### Expected Result

Ocf Client can connect to Ocf Server.

### Test Builds

---|---|---|---|---
master | f90d9d1 | frdm_k64f | May 24, 2018 | Fail
master | f90d9d1 | arduino 101 | May 24, 2018 | Pass

jimmy-huang:
This patch fixes the issue that multicast discovery messages cannot be
routed to the server after Zephyr update 1.11 with IPV6, but turning on
promiscuous mode resolves this issue.

Fixes #1885


grgustaf:
See if the solution from #1881 would work for this too... (see discussion in #1842 where @pfalcon showed the right way to solve it).

jimmy-huang:
@grgustaf Yeah, that solution didn't work, in fact, ocf already sets CONFIG_NET_IF_MCAST_IPV6_ADDR_COUNT=4, and I tried increasing it with no success either, until I turned on promiscuous mode, then it worked, that's why I am hoping @pfalcon might know the real answer here.

pfalcon:
> @pfalcon might know the real answer here.

I don't know the exact answer.

> CONFIG_NET_IF_MCAST_IPV6_ADDR_COUNT=4, and I tried increasing it with no success either

To what value did you increase it? How many multicast addresses does OCF require?

jimmy-huang:
@pfalcon I've tried ranges from 4 up to 8, I believe ocf only requires 1 multicast addr for the discovery.

PR #1893, opened 2018-07-03
Update WebIDL and modify documentation to fit new "node.js-like" style, Part 2

t-harvey:
Folks,
This is the actual pull request for updating the documentation to: 1) follow the style of node.js and 2) correct the WebIDL.

I tried to incorporate the advice/corrections from the original test pull request (#1800).

As with the test pull request, I've limited myself to fixing grammatical errors and the actual WebIDL, but the documentation could definitely do with a thorough going over for technical detail as well.

That said, I've run all of the updated WebIDL through my own parser, and it all builds correctly.  I wrote an automated tester for this, but I'm largely a newbie, so I don't know where I can put it and how to get it to run as part of the automated testing; any advice, pointers, etc. would be very useful.

Thanks,
Tim


t-harvey:
Jimmy,
In the "test" pull request for updating the docs, one of the reviewers wanted the WebIDL put into a separate file and just linked to -- but I didn't think this worked very well, so I decided I'd try using the markdown feature called "details", which, as you can see, lets you hide away *details* until you need them.
Unfortunately, the text inside "details" is interpreted as html rather than as markdown, so we have to use html formatting there -- and I soon learned that without using the <p>'s, all of the indentation goes away.
Thus, we're stuck with some ugly text that displays nicely.
If you've got a better method (b/c I'm not an html expert!), I'm all ears...but as I said, I've got a script that rips out the WebIDL for testing, and it was trivial to get it to ignore "<p>", so it's easily worked around, and you and I are the only ones who'll ever see the awkwardness of the text... :-)

grgustaf:
Yeah, I looked through it too and looks good. Since it's all in documentation there should be no problem with just merging it. :)

@t-harvey are you sure you still needed the `<p>` tags after you started using `<pre>`? I think that takes all your newlines literally. So I suspect you could remove all those. Also don't need the double blank lines in some places.

docs/ble.md
```diff
@@ -24,191 +37,215 @@ treat them like decimals.*
...
+This IDL provides an overview of the interface; see below for documentation of
+specific API functions.  We also have a short document explaining [ZJS WebIDL conventions](Notes_on_WebIDL.md).
+<details>
+<summary> Click to show/hide WebIDL</summary>
+<pre>
 // require returns a BLE object
 // var ble = require('ble');
-
-[NoInterfaceObject]
+<p><p>
```
grgustaf: Here's where there are a bunch of double blanks; there are others around though.

t-harvey: Geoff,
Since both you and Jimmy noticed, this, I'll go back and rip out as many as possible... :-)

t-harvey: ...also, I, too, thought the "<pre>" would suffice for formatting, but I just went back and checked, and it doesn't do the trick; the only thing that worked was to wrap <pre> around <details> (instead of the other way around), but then the "Details" pulldown menu is in the wrong font, which looks ugly...

I can make this switch (i.e., precede <details> with <pre>), if you guys want, but as the <p>'s are invisible to everyone except us, I'm more inclined to leave it pretty but (mildly) cumbersome. :-)

I've uploaded one file to show what moving <pre> looks like -- once Travis has its way with the docs, compare/contrast the docs for i2c.

t-harvey: ...also, I, too, thought the pre-construct inside of the details-construct would suffice for formatting, but I just went back and checked, and it doesn't do the trick; the only thing that worked was to wrap the pre-construct around the details-construct (instead of the other way around), but then the pulldown menu is in the wrong font, which looks ugly...

I can make this switch (i.e., put the pre-construct first), if you guys want, but as the  html-newlines are invisible to everyone except us, I'm more inclined to leave it pretty but (mildly) cumbersome. :-)

I've uploaded one file to show what swapping the two constructs looks like -- once Travis has its way with the docs, compare/contrast the docs for i2c.

grgustaf: Yeah, I think the old way was better.

t-harvey: Great -- I'll clean up the git record (so expect a --force'd commit :-), remove as many newlines as I can aesthetically :-), and put in a final upload.

grgustaf:
LGTM +1

PR #1895, opened 2018-07-24
Fixed small things on grove and web-socket and redid webusb, which I somehow missed on the first go around.

t-harvey:
For grove_lcd, the single blank line caused all of the links to be double-spaced.

For web-socket, I noticed that the actual C code was passing around the OptionsObject, which existed in some previous version of the docs, so I reinstated it.

For webusb, well, that was just one I missed; it isn't any different than all of the rest.


jimmy-huang:
+1

PR #1900, opened 2019-04-26
Upgrade to Zephyr 1.14 LTS, update codebase for it

pfalcon:
This updates deps/zephyr to 1.14 tag, and then makes minimal codebase (mostly build files) changes to make it build. Note that I tested only buildability with a few boards, and run-tested only HelloWorld.js on qemu_x86.

This partially addresses https://github.com/intel/zephyr.js/issues/1899


pfalcon:
With these changes changes, there're still build failures for some ARM targets (96b_carbon, frdm_k64f). A more controversial change to address that is split into https://github.com/intel/zephyr.js/pull/1901.

jimmy-huang:
I was testing out this patch, it seems that Zephyr 1.14 has changes so many of its APIs that almost every module will fail to build and will need to be ported to the new APIs.  A lot of the configs has been removed and/or replaced with new ones, although I tried to fix some of it, it seems like there are way more than I expected and I don't think we have to resource at this moment to maintain the code base.  In my best effort, I tried at least removing all those components that fails to build so trlite won't complain.  We can probably merge this patch along with #1901, and that will leaves a big task for updating the code to work with Zephyr 1.14.0 LTS again.  Here's the list of components I found broken with new API changes:

aio, bluetooth, dgram, filesystem, sensors, all net stuff, ws, ashell, uart, and OCF, as a result, ashell is also not going to work.

pfalcon:
@jimmy-huang: Thanks for looking into this and for the reply! And sorry for late reply from my side. The main motivator for these changes was to make CI build again on our side: https://ci.linaro.org/view/lite-iot-ci/job/lite-aeolus-zephyr.js/ . Literally, I asked my boss whether we should try to fix Z.js for 1.14 and make builds green, or drop it from our CI. He said fixing is good, so that's what I tried to do. But as I mentioned, we do only the basic build/boot test for HelloWorld.js, and I didn't look beyond that.

You're right that the API churn in Zephyr is (still) quite big. I also saw that Travis CI was broken with 1.14 upgrade, but didn't have a chance to look into, thanks for doing that.

PR #1901, opened 2019-04-26
jerry.cmake: Disable LTO for JerryScript build

pfalcon:
With LTO compiling enabled, and Zephyr SDk 0.10.0, during final Zephyr
link (which isn't done with LTO option), for armv7m targets, weird
errors are thrown:

zephyr-sdk-0.10.0/arm-zephyr-eabi/.../arm-zephyr-eabi/bin/ld: error:
zephyr_prebuilt.elf uses VFP register arguments, /tmp/ccTrD4ap.ltrans1.ltrans.o does not

Note that both JerryScript and Zephyr seem to be built with correct and
consistent target options, which enable VFP usage.

Disabling LTO option for JerryScript build fixes this issue.


pfalcon:
This is follow up to https://github.com/intel/zephyr.js/pull/1900 . I split this into separate PR because I only can explain the need for this patch with LTO bugs as presented in Zephyr SDK 0.10.0 toolchain. (I'd say, there's industrial folklore that LTO can break with non-trivial projects in rather unobvious, hard to debug ways. I personally avoid using, despite being an optimization geek.)

Again, split from the main patch to see if other folks may have better ideas.

jimmy-huang:
+1, I ran into the same problem, so yeah, thanks for fixing this.

PR #1903, opened 2019-05-31
Update cmake and fix remaining build issues with Zephyr 1.14.0

jimmy-huang:
This patch will disable all the trlite tests on components that will fail after updating the Zephyr 1.14.0 LTS.  The list include aio, bluetooth, dgram, filesystem, sensors, all net stuff, ws, ashell, uart, and OCF.


pfalcon:
@jimmy-huang: Thanks for looking into this! Well, I'd say it makes sense to move the codebase to 1.14, even if some things are disabled, rather than let it stay unupdated on older version. That way, interested parties may be at least motivated to look into updating the problematic parts further, rather then say to themselves: "Oh, this requires too old Zephyr, probably won't work for us at all".

